PR #14, opened 2026-02-21
fix: prevent duplicate message delivery (issue #12)

### LINKED ISSUE #12: Bug: duplicate message delivery to agents (race condition in processedIds)
## Bug Report

A single Discord message (ID `1474889469185491025`) was delivered to an agent across **3 separate conversation turns**, causing the agent to respond "already done" twice.

## Root Cause

Race condition between three message processing paths. The in-memory `processedIds` set only covers one of them.

### Three paths that can trigger processing of the same message:

1. **Message loop** (`src/index.ts:798-869`) — polls DB, checks `processedIds` ✅
2. **Recovery** (`src/index.ts:882-894`) — `recoverPendingMessages()` does NOT check `processedIds` ❌
3. **Queue fallback** (`src/index.ts:853-863`) — `sendMessage` fails → re-enqueues without dedup check ❌

### Why `processedIds` fails:

- **Cleared on timestamp advancement** (line 795) — wiped before all consumers finish
- **Only checked in path 1** — paths 2 and 3 bypass it entirely
- **Non-persistent** — lost on process restart
- **Race-prone** — cleared while the message loop is still iterating

### Reproduction timeline:

```
T=0ms    Discord message arrives → stored to DB
T=50ms   Message loop polls → processedIds.add(id) → enqueueMessageCheck → Turn 1
T=5000ms Container finishes Turn 1 → lastAgentTimestamp advances
T=5200ms processedIds.clear() (timestamp advanced, line 795)
T=???    recoverPendingMessages() → finds message again → Turn 2 (DUPLICATE)
T=???    Another trigger → Turn 3 (DUPLICATE)
```

## Files

- `src/index.ts` — lines 124 (`processedIds`), 776-894 (message loop + recovery)
- `src/channels/discord.ts` — lines 339-694 (message ingestion)
- `src/group-queue.ts` — lines 122-156 (queue management)
- `src/db.ts` — lines 298-394 (queries use `INSERT OR REPLACE` so no uniqueness enforcement)

## Suggested Fix

Option A (persistent dedup):
- Add a `processed` boolean column to the `messages` table
- Mark messages as processed when `lastAgentTimestamp` advances
- Filter by `processed = 0` in **all** query paths
- Remove the in-memory `processedIds` set

Option B (cursor-only):
- Use `lastAgentTimestamp` as the single source of truth across **all** code paths
- Ensure `recoverPendingMessages()` and queue fallback both respect this cursor
- Never re-query messages with `timestamp <= lastAgentTimestamp`

Peyton-Spencer:
## Summary

• Changed `getNewMessages` SQL query from `>=` to `>` (strict greater-than), matching how `getMessagesSince` already works
• Removed the fragile in-memory `processedIds` Set that was supposed to deduplicate boundary messages but was cleared prematurely and bypassed by recovery/queue paths
• Updated test to reflect strict-after semantics

## Root Cause

The message loop polled with `timestamp >= lastTimestamp`, re-fetching messages at the cursor boundary every cycle. `processedIds` was meant to handle this but:
1. Cleared when timestamp advanced — before all consumers finished
2. Only checked in the main message loop — `recoverPendingMessages()` and queue fallback bypassed it entirely
3. Non-persistent — lost on process restart

This caused the same message to be delivered 2-3x to agents.

## Fix

Use `>` instead of `>=` in `getNewMessages`, eliminating boundary re-fetches at the source. Since `lastTimestamp` is set to the max timestamp of each batch, `>` ensures only truly new messages are returned. This makes `processedIds` unnecessary — removed entirely.

## Test plan

- [x] `bun test src/db.test.ts` — all 20 tests pass
- [x] `bun test` — all relevant tests pass (2 unrelated failures in container/agent-runner due to missing SDK dep)

Closes #12

### DIFF
--- a/src/db.test.ts
+++ b/src/db.test.ts
@@ -287,15 +287,14 @@ describe('getNewMessages', () => {
     expect(newTimestamp).toBe('2024-01-01T00:00:04.000Z');
   });
 
-  it('filters by timestamp', () => {
+  it('filters by timestamp (strictly after)', () => {
     const { messages } = getNewMessages(
       ['group1@g.us', 'group2@g.us'],
       '2024-01-01T00:00:02.000Z',
     );
-    // Messages at or after timestamp (g2 msg1 at 00:00:02, g1 msg2 at 00:00:04)
-    expect(messages).toHaveLength(2);
-    expect(messages[0].content).toBe('g2 msg1');
-    expect(messages[1].content).toBe('g1 msg2');
+    // Only messages strictly after the cursor (g1 msg2 at 00:00:04)
+    expect(messages).toHaveLength(1);
+    expect(messages[0].content).toBe('g1 msg2');
   });
 
   it('returns empty for no registered groups', () => {
--- a/src/db.ts
+++ b/src/db.ts
@@ -351,7 +351,7 @@ export function getNewMessages(
   const sql = `
     SELECT id, chat_jid, sender, sender_name, content, timestamp, sender_user_id, mentions
     FROM messages
-    WHERE timestamp >= ? AND chat_jid IN (${placeholders}) AND (is_from_me IS NULL OR is_from_me = 0)
+    WHERE timestamp > ? AND chat_jid IN (${placeholders}) AND (is_from_me IS NULL OR is_from_me = 0)
     ORDER BY timestamp
   `;
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -121,7 +121,6 @@ let agents: Record<string, Agent> = {};
 let channelRoutes: Record<string, ChannelRoute> = {};
 let lastAgentTimestamp: Record<string, string> = {};
 let messageLoopRunning = false;
-let processedIds = new Set<string>(); // Deduplication for timestamp boundary messages
 
 // Track consecutive errors per group to prevent infinite error loops.
 // After MAX_CONSECUTIVE_ERRORS, the cursor advances past the failing batch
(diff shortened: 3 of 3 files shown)


Peyton-Spencer (commented):
## Review ✅

Clean, well-reasoned fix. Reviewed the full change and it's correct.

**Analysis:**

1. **`>=` → `>` in `getNewMessages`** — This is the right fix. Since `lastTimestamp` is set to `max(batch timestamps)` after each poll cycle, using `>=` was guaranteed to re-fetch the boundary message every poll. The `>` strict comparison eliminates re-fetches at the source.

2. **`processedIds` removal** — Good cleanup. The PR description accurately identifies three flaws in the old dedup:
   - Cleared on timestamp advance *before* consumers finished
   - Bypassed by `recoverPendingMessages()` (which uses `getMessagesSince`, already `>`)
   - Bypassed by `queue.enqueueMessageCheck()` → `processGroupMessages()` (also uses `getMessagesSince` with `>`)
   - Non-persistent across restarts

3. **Consistency with `getMessagesSince`** — Confirmed: `getMessagesSince` (line 383) already uses `>`. This PR makes `getNewMessages` match, so both functions now have identical semantics.

4. **Test update** — Correctly reflects strict-after semantics. The old test expected 2 messages (at-or-after `00:00:02`), new test expects 1 (strictly after `00:00:02`).

**Edge case consideration:** If two messages arrive with *identical* timestamps, the second could be missed if `lastTimestamp` advances to that timestamp after the first is processed. However, this is the same behavior `getMessagesSince` has always had, and in practice WhatsApp/Discord timestamps have sub-second precision making true collisions extremely rare. The previous `processedIds` approach didn't actually protect against this either (it tracked IDs, not timestamps).

Net -14 LOC, CI green. LGTM ✅

PR #20, opened 2026-02-22
fix: cross-platform homedir + recursive skill copy

Peyton-Spencer:
## Summary

Two upstream-inspired fixes that improve cross-platform compatibility and prevent crashes:

• Replace hardcoded `/Users/user` fallback with `os.homedir()` in 3 files — the macOS-specific path fails on Linux containers
• Fix skill copy to handle subdirectories (scripts/, templates/) — `copyFileSync` crashes with EISDIR on directories

## Changes

*1. Cross-platform home directory* (inspired by upstream 6e22abb)

| File | Change |
|------|--------|
| `src/config.ts` | `os.homedir()` instead of `'/Users/user'` |
| `src/mount-security.ts` | Same fix in `expandPath()` |
| `src/s3/file-sync.ts` | Same fix for SSH key path resolution |

*2. Recursive skill copy* (inspired by upstream d336b32)

| File | Change |
|------|--------|
| `src/backends/local-backend.ts` | Replace `copyFileSync` loop with `fs.cpSync(srcDir, dstDir, { recursive: true })` |
| `src/s3/file-sync.ts` | Recursive directory walk for S3 upload (skills with subdirectories) |

## Why not cherry-pick directly?

Our fork has different file structure from upstream:
- Skill copy code lives in `src/backends/local-backend.ts` (not `src/container-runner.ts`)
- S3 file sync has its own skill upload path that also needed the recursive fix
- The `os.homedir()` fix extends to a 3rd file (`file-sync.ts`) that upstream doesn't have

## Test plan

- [x] All 118 tests pass, 0 failures
- [x] Type-check unchanged (same pre-existing errors as main)
- [x] No new `'/Users/user'` references remain in codebase

### DIFF
--- a/src/backends/local-backend.ts
+++ b/src/backends/local-backend.ts
@@ -146,12 +146,7 @@ function buildVolumeMounts(
       const srcDir = path.join(skillsSrc, skillDir);
       if (!fs.statSync(srcDir).isDirectory()) continue;
       const dstDir = path.join(skillsDst, skillDir);
-      fs.mkdirSync(dstDir, { recursive: true });
-      for (const file of fs.readdirSync(srcDir)) {
-        const srcFile = path.join(srcDir, file);
-        const dstFile = path.join(dstDir, file);
-        fs.copyFileSync(srcFile, dstFile);
-      }
+      fs.cpSync(srcDir, dstDir, { recursive: true });
     }
   }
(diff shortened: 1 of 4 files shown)


Peyton-Spencer:
https://github.com/greptileai review

PR #111, opened 2026-02-26
fix: adopt upstream security + stability improvements

Peyton-Spencer:
## Summary

Adopts three targeted fixes from upstream PRs, adapted for our multi-backend architecture:

• *[Upstream PR #507]* Shell injection defense-in-depth in `cleanupOrphanedContainers()`. Validates container names against `SAFE_CONTAINER_NAME_RE` before passing to `Bun.spawn`. Names from container runtime output are re-validated even though the generation path already sanitizes them.

• *[Upstream PR #503]* Pass `ANTHROPIC_BASE_URL` to containers for custom API endpoint support (corporate proxies, cloud gateways). Added to env var allowlist in `local-backend.ts` and `SECRET_ENV_VARS` in agent-runner to prevent leaking via Bash subprocesses.

• *[Upstream PR #503]* Session resume recovery: detect `error_during_execution` before `system/init` (indicates stale session transcript, e.g. after container recreation) and automatically retry as a fresh session instead of silently failing.

## Fork Adaptations

- Shell injection: Adapted for `LOCAL_RUNTIME` abstraction (docker/container dual support) instead of upstream's single `container-runner.ts`. Uses `Bun.spawn` array form (already shell-safe) + name validation for defense-in-depth.
- ANTHROPIC_BASE_URL: Added alongside existing `ANTHROPIC_MODEL` in env allowlist (upstream only has the former).
- Session recovery: Integrated into existing query loop with `sessionStale` flag pattern, preserving the idle-wait and close-sentinel logic.

## Test plan

- [ ] Verify container name validation rejects names with shell metacharacters
- [ ] Verify ANTHROPIC_BASE_URL is forwarded to containers when set in .env
- [ ] Verify stale session detection triggers fresh session retry (test by corrupting sessionId)
- [ ] All existing tests pass (pre-existing failures unchanged)

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects and recovers from stale sessions during queries, automatically retrying so interrupted agent runs resume without manual intervention.

* **Chores**
  * Expanded secret-environment sanitization for container tools.
  * Hardened orphan-container cleanup with safer name validation and more accurate cleanup logging.

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -297,7 +297,7 @@ function createPreCompactHook(): HookCallback {
 // Secrets to strip from Bash tool subprocess environments.
 // These are needed by claude-code for API auth but should never
 // be visible to commands Kit runs.
-const SECRET_ENV_VARS = ['ANTHROPIC_API_KEY', 'CLAUDE_CODE_OAUTH_TOKEN'];
+const SECRET_ENV_VARS = ['ANTHROPIC_API_KEY', 'ANTHROPIC_BASE_URL', 'CLAUDE_CODE_OAUTH_TOKEN'];
 
 /**
  * createSanitizeBashHook
@@ -672,7 +672,7 @@ async function runQuery(
   containerInput: ContainerInput,
   sdkEnv: Record<string, string | undefined>,
   resumeAt?: string,
-): Promise<{ newSessionId?: string; lastAssistantUuid?: string; closedDuringQuery: boolean }> {
+): Promise<{ newSessionId?: string; lastAssistantUuid?: string; closedDuringQuery: boolean; sessionStale?: boolean }> {
   const stream = new MessageStream();
   stream.push(prompt);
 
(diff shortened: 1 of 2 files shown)


Peyton-Spencer:
**Review:** Good adoption of three upstream fixes. Two minor CodeRabbit findings to address:

### 1. IPC Poller Leak on Early Return (Medium)
The stale session early return exits `runQuery()` without setting `ipcPolling = false`. Orphaned timer callbacks could theoretically race with the fresh retry's close sentinel detection.

**Fix:** Add `ipcPolling = false; stream.end();` before the early return.

### 2. Stale `queryResult` Reference After Retry (Low)
Downstream `closedDuringQuery` check reads from stale `queryResult` after retry block. Practically a no-op but ambiguous.

**Fix:** Reassign `queryResult = freshResult` after retry.

Both are 1-2 line changes. The rest is solid — container name regex, ANTHROPIC_BASE_URL dual treatment, and sessionStale flag all look good.

PR #159, opened 2026-03-01
fix: clean up stale dispatch IPC folders instead of quarantining

Peyton-Spencer:
## Summary

- Fixes a race condition where dispatch runtime IPC folders were falsely quarantined after a container exited
- Stale dispatch folders (valid owner, container just finished) now have their remaining files processed before being deleted — nothing is lost
- Truly unknown folders (unrecognized owner or malformed digest) are still quarantined as before

## Root Cause

When a container exits, `activeRuntimeFolders.delete(runtimeFolder)` runs in the `finally` block of `runAgent`. On the next IPC poll (1s later), the watcher found the orphaned `{agent}__dispatch__{digest}` directory still on disk but no longer in the allowlist → spurious `warn` log and quarantine.

## Fix

Separates two concerns that were conflated in `resolveOwnerGroupFolder`:

1. **Ownership validation** (stays in `resolveOwnerGroupFolder`): known owner + valid 16-char hex digest → return owner. Unknown/malformed → return null → quarantine.

2. **Liveness / cleanup** (moves to `startIpcWatcher`): after resolving ownership, detect stale dispatch folders (`DISPATCH_RUNTIME_SEP` present, valid owner, not in `activeRuntimeFolders`). Process any remaining IPC files, then `rmSync` the directory. Debug log only.

## Test plan

- [ ] All 807 existing tests pass (`bun test`)
- [ ] No more `Skipping IPC from unrecognized source folder — quarantining` warnings in logs after containers exit normally
- [ ] Truly rogue/unknown folders (e.g. `omniaura-discord` from old migration) still get quarantined

## Summary by CodeRabbit

* **Refactor**
  * Simplified IPC identity validation to validate owners and digests only, decoupling it from runtime liveness checks.

* **New Features**
  * Added detection of stale dispatch IPC folders and automatic cleanup after processing, removing orphaned directories and retrying cleanup on subsequent polls if removal fails.

### DIFF
--- a/src/ipc.ts
+++ b/src/ipc.ts
@@ -477,8 +477,8 @@ export function startIpcWatcher(deps: IpcDeps): void {
         try {
           fs.rmSync(staleDir, { recursive: true, force: true });
           logger.debug({ sourceGroup }, 'Cleaned up stale dispatch IPC folder');
-        } catch {
-          /* ignore — will retry next poll */
+        } catch (err) {
+          logger.debug({ sourceGroup, err }, 'Failed to clean up stale dispatch IPC folder — will retry next poll');
         }
       }
     }


nicktallen (approved):
## Review: ✅ LGTM

Clean fix for a real race condition. The separation of concerns is well thought out.

### What I like

- **Root cause is clearly identified**: the `finally` block in `runAgent` removes the folder from `activeRuntimeFolders`, but the directory lingers on disk for up to 1s until the next IPC poll — classic TOCTOU race.
- **No data loss**: stale dispatch folders get their remaining IPC files fully processed *before* deletion. This is the key correctness property.
- **Minimal blast radius**: only `src/ipc.ts` is touched, and the structural ownership validation in `resolveOwnerGroupFolder` is preserved. Truly unknown/malformed folders still get quarantined as before.
- **Good use of `force: true`** on `rmSync` so it's idempotent if the dir is already gone.

### Minor observations (non-blocking)

1. **Silent `catch` on cleanup** — the empty catch block at the stale-dir cleanup is fine for retry-on-next-poll semantics, but a `logger.debug` there could help during future debugging if cleanup repeatedly fails (e.g. permission issue). Very minor.

2. **Security note** — removing the `activeRuntimeFolders` check from `resolveOwnerGroupFolder` means a folder with a valid owner + valid digest format will now pass ownership validation even if it was never launched by the orchestrator. The stale-dispatch cleanup downstream mitigates this for dispatch folders, but it's worth noting the shift: ownership validation is now purely structural, not liveness-based. The PR description is transparent about this, so 👍.

Overall this is a well-scoped fix with a clear commit message and good PR description. Nice work.

PR #164, opened 2026-03-01
fix: route scheduled task messages through the task's assigned Discord bot

Peyton-Spencer:
## Summary

- `SchedulerDependencies.sendMessage` and `findChannel` now accept an optional `discordBotId` parameter
- `runTask` passes `group.discordBotId` at all three call sites (channel lookup, thread announcement, result delivery)
- `index.ts` scheduler callbacks forward the bot ID to `findChannelForJid`

## Root cause

When a scheduled task fired, `sendMessage` called `findChannelForJid(jid)` with no preferred bot ID. `getDiscordBotIdForJid` resolved the channel's **primary** subscriber's bot ID (`'PRIMARY'` — a placeholder string, not a real snowflake), found no channel match, and fell back to `DISCORD_DEFAULT_BOT_ID` (PeytonOmni). Tasks assigned to OCPeyton were running in the right container but replying through PeytonOmni.

## DB fix (applied directly to live DB)

`channel_subscriptions.discord_bot_id` had placeholder strings for all 11 shared channels:
- `ditto-discord`: `'PRIMARY'` → `1471198154434154508`
- `ocpeyton-discord`: `'OCPEYTON'` → `1476396931709276191`

Also patched the 3 affected `channel_routes` rows.

## Test plan

- [ ] Deploy and trigger a scheduled task assigned to OCPeyton — confirm OCPeyton bot replies
- [ ] Confirm PeytonOmni-assigned tasks still reply through PeytonOmni
- [ ] `bun run typecheck` passes (verified locally)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1987,8 +1987,8 @@ async function main(): Promise<void> {
         undefined,
         lane,
       ),
-    sendMessage: async (jid, rawText) => {
-      const ch = findChannelForJid(jid);
+    sendMessage: async (jid, rawText, discordBotId?) => {
+      const ch = findChannelForJid(jid, discordBotId);
       if (!ch) {
         logger.warn({ jid }, 'No channel found for scheduled message');
         return;
@@ -2004,7 +2004,7 @@ async function main(): Promise<void> {
         return msgId ? String(msgId) : undefined;
       }
     },
-    findChannel: (jid) => findChannelForJid(jid),
+    findChannel: (jid, discordBotId?) => findChannelForJid(jid, discordBotId),
   });
   startIpcWatcher({
     sendMessage: async (jid, rawText) => {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1987,7 +1987,7 @@ async function main(): Promise<void> {
         undefined,
         lane,
       ),
-    sendMessage: async (jid, rawText, discordBotId?) => {
+    sendMessage: async (jid, rawText, discordBotId) => {
       const ch = findChannelForJid(jid, discordBotId);
       if (!ch) {
         logger.warn({ jid }, 'No channel found for scheduled message');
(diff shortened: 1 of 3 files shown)


nicktallen (commented):
## Review — PR #164

**Overall: Looks good ✅ — clean, minimal fix with one CI nit to address.**

### What it does

Threads `group.discordBotId` through `SchedulerDependencies.sendMessage` and `findChannel` so scheduled task messages are routed through the correct Discord bot instead of always falling back to the default (PeytonOmni).

### What I like

- **Minimal and well-scoped** — only touches the two files that matter, no unnecessary changes
- **Backward compatible** — `discordBotId` is optional in the interface, so existing callers are unaffected
- **All three call sites covered** — channel lookup, thread announcement, and result delivery all pass `group.discordBotId`
- **Excellent PR description** — root cause analysis, DB fix documentation, and test plan are all clear

### Nit: Format CI failure

The `format` check is failing — Prettier is flagging `src/task-scheduler.ts` (and a pre-existing issue in `src/ipc.ts`). The likely cause is the `discordBotId?` syntax in the arrow function parameters in `src/index.ts`:

```ts
sendMessage: async (jid, rawText, discordBotId?) => {
findChannel: (jid, discordBotId?) => findChannelForJid(jid, discordBotId),
```

Since the arrow functions already infer their parameter types from the `SchedulerDependencies` interface, you can just drop the `?` — the optionality is already conveyed by the interface:

```ts
sendMessage: async (jid, rawText, discordBotId) => {
findChannel: (jid, discordBotId) => findChannelForJid(jid, discordBotId),
```

A quick `bun run format:check` / `bunx prettier --write src/task-scheduler.ts src/index.ts` should sort it out.

### Verdict

Logic is correct and the approach is sound. Just needs the formatter fix to go green. 👍

nicktallen (approved):
Clean fix — threads `discordBotId` through the scheduler deps at all three call sites. Backward-compatible optional param, good root cause analysis.

Minor suggestion for follow-up: consider a DB migration script for the placeholder → snowflake ID fix so it survives DB recreation. Not a blocker.

PR #172, opened 2026-03-02
fix(container): preinstall agent-browser Chromium deps

Peyton-Spencer:
## Summary
- install `agent-browser` during container build instead of relying on first-use setup
- run `playwright install --with-deps chromium` at build time so required Linux system libraries are present
- set a shared Playwright browser path and permissions so the non-root `bun` runtime user can use the preinstalled Chromium binary

## Why
The runtime container user cannot escalate to root, so `playwright install --with-deps chromium` fails when executed on demand. Preinstalling Chromium and its OS dependencies during image build removes that failure mode and makes browser automation work out of the box.

## Summary by CodeRabbit

* **Chores**
  * Browser tooling (agent-browser and Playwright) is pre-installed during image build; browser binaries are placed in a shared, permissioned location and required system libraries are ensured. Retains existing CLI installation path and version pinning for tools.

* **Documentation**
  * Removed first-time setup instructions for agent-browser to simplify onboarding.

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -59,8 +59,15 @@ ARG GRAPHITE_VERSION=1.7.16
 RUN npm install -g @withgraphite/graphite-cli@${GRAPHITE_VERSION}
 
 # Install claude-code globally (default agent runtime)
-# Note: agent-browser is available as an on-demand skill (installed on first use)
 RUN bun install -g @anthropic-ai/claude-code
+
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Triage pass (2026-03-02): looks ready to merge.

- Container-only scope and low blast radius
- CI is fully green
- Diff removes obsolete setup docs in sync with image behavior

No blockers from triage.

PR #231, opened 2026-03-05
Add DB-backed agent discovery health snapshots

Peyton-Spencer:
## Summary
- Adds an additive discovery layer to `agent_registry.json` by introducing a `discoveryVersion` field while preserving all existing identity fields (`id`, `jid`, `jids`, `name`, `trigger`, `sendTo`).
- Introduces a DB-backed `agent_health` table and writes a separate sibling IPC snapshot, `agent_discovery.json`, for volatile runtime state (`isOnline`, `lastHeartbeatAt`, sanitized `capabilities`).
- Tracks online/offline transitions during `runAgent()` and keeps health data read-only for visibility use cases (no cross-instance control surface in this pass).

## Security
- Keeps volatile health data isolated from identity data to avoid leaking container configs or mount details.
- Capabilities are intentionally sanitized to high-level, non-sensitive labels only.

## Validation
- `bun test src/db.test.ts src/db.migrations.test.ts`
- `bun run build`

## Follow-up
- This is PR 1 of 2 (schema + discovery protocol).
- PR 2 will wire channel roster consumers/UI on top of these snapshots.

## Release Notes

* **New Features**
  * Agent health tracking now persists online/offline status, heartbeat timestamps, and capabilities for each agent.
  * Agent discovery data is generated and exported, including version information, generation timestamp, and agent health records.
  * Agent capabilities are automatically computed and tracked during agent lifecycle (startup to completion/error).

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -190,11 +190,12 @@ function mapRowToAgentHealth(row: AgentHealthRow): AgentHealth {
     isOnline: row.is_online === 1,
     lastHeartbeatAt: row.last_heartbeat_at || row.updated_at,
     updatedAt: row.updated_at,
-    capabilities: safeJsonParse<string[]>(
-      row.capabilities,
-      { agentId: row.agent_id },
-      'agent health capabilities',
-    ) || [],
+    capabilities:
+      safeJsonParse<string[]>(
+        row.capabilities,
+        { agentId: row.agent_id },
+        'agent health capabilities',
+      ) || [],
   };
 }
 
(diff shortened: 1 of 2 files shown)


Peyton-Spencer:
Triage review: agent-health persistence/discovery additions look coherent and tests cover schema + DB API, but CI format job is failing; once formatting is fixed this looks mergeable.

Peyton-Spencer:
Current triage pass: useful change set, but still blocked.

- CI remains non-green because `format` is failing.
- I also still see two correctness concerns in the branch: `agent_discovery.json` is written directly instead of atomically, and agent online/offline state is still toggled with plain `setAgentHealth(...)` calls rather than a concurrency-safe active-run counter.

Please fix the formatting failure and address those discovery-state race edges before merge.

PR #230, opened 2026-03-05
feat: Discord guild roster fetching and caching (Phase 1)

Peyton-Spencer:
## Summary

Phase 1 of cross-instance agent discovery via Discord guild rosters. Resolves the problem where agents on different machines (e.g., MacBook Pro vs Orange Pi) can't see each other because each OmniClaw instance has an isolated SQLite database with no cross-machine sync.

This PR enables each instance to fetch the full guild member list from Discord's API, cache it locally, and sync it to agent containers via IPC snapshots.

### Changes

- **Discord adapter** (`discord.ts`): Add `GuildMembers` intent + `fetchGuildRoster()` method that returns humans and bots with roles
- **Database** (`db.ts`): Add `guild_rosters` and `guilds` tables with CRUD helpers (`storeGuildRoster`, `getGuildRoster`, `getAllGuildRosters`, `getGuildInfo`)
- **IPC snapshots** (`ipc-snapshots.ts`): Add `writeRostersSnapshot()` to write `guild_rosters.json` to each agent's IPC directory
- **Orchestrator** (`index.ts`): Wire up roster refresh on startup + 15-minute periodic refresh interval

### How It Works

1. On startup (after Discord connects), `refreshGuildRosters()` collects all known guild IDs from registered groups and channel subscriptions
2. Fetches full member list from Discord API via `guild.members.fetch()`
3. Stores members in `guild_rosters` SQLite table (upsert with full replace per guild)
4. Writes `guild_rosters.json` snapshot to every agent IPC directory
5. Repeats every 15 minutes via `setInterval`

### Requirements

- **Discord Bot Permission**: "Server Members" privileged intent must be enabled in the Discord Developer Portal for each bot token
- No breaking changes to existing functionality

### Phase 2 (follow-up PRs)

- [ ] Inject roster into `<messages participants="...">` formatting (OCPeyton's PR)
- [ ] Bot metadata discovery (identify which bots are OmniClaw-managed)
- [ ] Optional heartbeat-based instance health tracking

## Test Plan

- [x] `tsc --noEmit` passes (0 errors)
- [x] `bun test` passes (931 tests, 0 failures)
- [x] `prettier --check` passes on all modified files
- [ ] Manual: verify `guild_rosters.json` appears in agent IPC dirs after startup
- [ ] Manual: verify "Server Members" intent is enabled in Discord Developer Portal

## Release Notes

* **New Features**
  * Added automatic guild roster synchronization for Discord servers
[...]

PR #232, opened 2026-03-05
feat(router): inject channel roster context into messages

Peyton-Spencer:
## Summary
- Inject `channel_roster` into formatted `<messages ...>` envelopes so agents receive channel-level humans+bots context, not only excerpt-local senders.
- Clarify participant semantics by adding `excerpt_participants` while keeping backward-compatible `participants` for existing prompts/parsers.
- Wire roster lookup at all message formatting callsites so both queued and fresh dispatch paths include roster context when Discord guild roster cache is available.

## Why
Cross-instance visibility needs two layers: shared roster data (Phase 1 in #230) and prompt-level context usage. This PR is the Phase 1B consumer layer so agents can actually use the roster for reasoning and participant attribution.

## Changes
- `src/router.ts`
  - Add `FormatMessagesOptions` with `channelRosterNames`.
  - Add `excerpt_participants` attribute.
  - Keep legacy `participants` attribute for compatibility.
  - Add `channel_roster` attribute when roster names are available.
- `src/index.ts`
  - Add `getChannelRosterNames(chatJid, explicitGuildId?)` helper using cached guild roster data.
  - Pass roster names into `formatMessages(...)` in all dispatch paths.
- `src/formatting.test.ts`
  - Update exact-header assertion for `excerpt_participants` + compatibility `participants`.
  - Add coverage for `channel_roster` formatting and dedup behavior.

## Validation
- `bun run typecheck`
- `bun test src/formatting.test.ts`
- `bunx prettier --check src/router.ts src/index.ts src/formatting.test.ts`

## Notes
- This branch is intended as a follow-up stacked on top of #230.
- Full `bun test` has an existing unrelated failure in `src/schedule-utils-dst.test.ts` in this environment.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -123,6 +123,10 @@ export const SESSION_MAX_AGE = parseInt(
   process.env.SESSION_MAX_AGE || '14400000',
   10,
 ); // 4 hours — rotate sessions to prevent unbounded context growth
+export const ROSTER_REFRESH_INTERVAL = parseInt(
+  process.env.ROSTER_REFRESH_INTERVAL || '900000',
+  10,
+); // 15min default — how often to refresh Discord guild rosters
 /** Max containers actively processing messages or tasks. */
 export const MAX_ACTIVE_CONTAINERS = Math.max(
   1,
--- a/src/formatting.test.ts
+++ b/src/formatting.test.ts
@@ -58,13 +58,21 @@ describe('escapeXml', () => {
 // --- formatMessages ---
 
 describe('formatMessages', () => {
-  it('formats a single message as XML with participants attribute', () => {
+  it('formats a single message as XML with participant attributes', () => {
     const result = formatMessages([makeMsg()]);
     expect(result).toBe(
-      '<messages participants="Alice">\n' +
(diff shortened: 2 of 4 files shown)


Peyton-Spencer (commented):
**Clayton's Review — LGTM (>95% confidence)**

Clean and well-structured Phase 1B.

**What I like:**
- `excerpt_participants` + `participants` backward-compat alias is a smart split — existing prompts keep working
- `channel_roster` dedup via Set handles the common case of overlapping message senders + guild members
- `getChannelRosterNames()` properly falls back through `explicitGuildId` → `getChatGuildId(chatJid)`
- All three `formatMessages` callsites wired up (processGroupMessages + both queued dispatch paths)
- Test coverage for the new header format and dedup behavior

**One note:** My env var fix (config.ts `ROSTER_REFRESH_INTERVAL`) is also in this diff since it's stacked on the original #230 commit. I've pushed that fix to #230 already, so the diff will shrink once GH rebases. No merge conflict expected since the changes are identical.

>95% confidence — ready to merge after #230 lands.

PR #312, opened 2026-03-14
fix(web): align page morphing with datastar

Peyton-Spencer:
## Summary
- switch web UI page navigation from JSON/manual DOM swaps to Datastar HTML morph patches via `/api/page/*`
- wire nav links to Datastar `@get(...)` actions so patched pages re-run `data-init` behavior through the framework instead of bypassing it
- fix the live SSE status badge class mismatch and add regression coverage for the new patch-based navigation flow

## Testing
- `bun test src/web/server.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Stream and view real-time logs from trusted peers
  * Added "Logs" action button and a remote logs panel for live display

* **Behavior Changes / Refactor**
  * Navigation now uses full-page reloads with partial page patching for updates
  * Server enforces a limit on concurrent raw log streams (excess clients receive an error)

* **Tests**
  * Added tests for log streaming endpoints, proxying trusted peer logs, and concurrency limits

Issue #301, opened 2026-03-13
bug: globalThis.fetch mock leak in discovery/routes.test.ts fails 100+ tests in CI


Peyton-Spencer:
## Problem

CI is failing on every PR branch due to a test isolation bug. When the full test suite runs, `src/discovery/routes.test.ts` overwrites `globalThis.fetch` with a mock that throws "should not be called" (line 74), but never restores the original `fetch` reference. The `afterEach(() => mock.restore())` on line 65-67 only restores mocks created through Bun's `mock()` system — it does NOT restore the direct `globalThis.fetch = fetchSpy` assignment.

This causes 100+ cascading failures in `src/web/server.test.ts` and other files that depend on `fetch` working correctly.

### Reproduction

```bash
# These pass individually:
bun test src/discovery/routes.test.ts      # 9 pass, 0 fail
bun test src/web/server.test.ts            # 61 pass, 0 fail

# But fail together:
bun test src/discovery/routes.test.ts src/web/server.test.ts
# 9 pass, 61 fail — all server tests fail with "should not be called"

# Full suite:
bun test
# ~100 failures, all downstream of the fetch mock leak
```

### Root Cause

`src/discovery/routes.test.ts:74`:
```ts
globalThis.fetch = fetchSpy as unknown as typeof fetch;
```

This direct assignment is not tracked by Bun's `mock.restore()`, so it persists across test files.

### Fix

Save and restore the original `fetch` reference:

```ts
let originalFetch: typeof fetch;

beforeEach(() => {
  originalFetch = globalThis.fetch;
});

afterEach(() => {
  globalThis.fetch = originalFetch;
  mock.restore();
});
```

Or use `spyOn(globalThis, 'fetch')` which integrates with `mock.restore()`.

### Impact

- All open PRs (#298, #299, #300) are failing CI
- No PR can merge with green CI until this is fixed
- The test failures mask real issues in PR test runs

Peyton-Spencer:
**Priority escalation**: This is effectively blocking all CI on every PR branch. The two open PRs (#312, #313) cannot get clean CI runs until this is fixed.

The fix is straightforward — save and restore `globalThis.fetch` in the test's setup/teardown:

```ts
// In src/discovery/routes.test.ts
let originalFetch: typeof globalThis.fetch;
beforeEach(() => {
  originalFetch = globalThis.fetch;
  // ... existing mock setup
});
afterEach(() => {
  globalThis.fetch = originalFetch;
  mock.restore();
[...]

PR #313, opened 2026-03-14
feat(web): proxy remote peer logs into the dashboard

Peyton-Spencer:
## Summary
- add an authenticated raw log stream endpoint and trusted peer proxy route so one OmniClaw instance can relay another instance's logs without sharing dashboard credentials
- surface remote peer log streaming in the Network page with a dedicated remote logs panel and per-peer `Logs` action
- cover the new log stream/proxy flow with tests and typecheck the updated web/discovery code paths

## Testing
- `bun test src/web/server.test.ts src/web/log-stream.test.ts src/discovery/routes.test.ts`
- `bun run tsc --noEmit`

### DIFF
--- a/src/discovery/routes.test.ts
+++ b/src/discovery/routes.test.ts
@@ -5,6 +5,8 @@ import {
   type DiscoveryRouteContext,
 } from './routes.js';
 
+const originalFetch = globalThis.fetch;
+
 const defaultState = {
   getAgents: () => ({}),
   getChannelSubscriptions: () => ({}),
@@ -62,6 +64,7 @@ function makeContext(
 
 afterEach(() => {
   mock.restore();
+  globalThis.fetch = originalFetch;
 });
 
 describe('handleDiscoveryRequest', () => {
--- a/src/web/server.test.ts
+++ b/src/web/server.test.ts
@@ -850,6 +850,27 @@ describe('SSE', () => {
     reader.releaseLock();
   });
 
+  it('enforces the MAX_SSE_CLIENTS limit for raw log streams', async () => {
+    handle = startWebServer(testConfig(), makeState());
+
+    const readers: CancellableReader[] = [];
+    for (let i = 0; i < 100; i++) {
+      const res = await authedFetch('/api/logs/stream', {
+        headers: { Accept: 'text/event-stream' },
+      });
+      expect(res.status).toBe(200);
+      expect(res.body).toBeTruthy();
+      readers.push(res.body!.getReader() as unknown as CancellableReader);
+    }
+
+    const overflow = await authedFetch('/api/logs/stream', {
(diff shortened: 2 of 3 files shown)


src/discovery/routes.test.ts
```diff
@@ -97,4 +97,53 @@ describe('handleDiscoveryRequest', () => {
       status: 'already_trusted',
     });
   });
+
```
Peyton-Spencer: **`globalThis.fetch` mock leak** — this sets the global mock but never restores it, which is the same pattern issue #301 calls out as breaking 100+ tests in CI.

Needs `beforeEach`/`afterEach` save+restore:
```typescript
let originalFetch: typeof globalThis.fetch;
beforeEach(() => { originalFetch = globalThis.fetch; });
afterEach(() => { globalThis.fetch = originalFetch; });
```
Or better: inject fetch via `PeerClient` constructor so tests don't touch the global.

Peyton-Spencer (commented):
Both findings addressed cleanly. LGTM — ready to merge.

- `globalThis.fetch` saved at module scope and restored in `afterEach` — mock leak fixed
- `rawLogStreamClients` counter with `MAX_SSE_CLIENTS` check + 429 response — matches `/api/events` behavior
- `cleanup()` with `closed` guard prevents double-decrement — nice defensive coding
- New test opens 100 streams then verifies 101st gets 429

The rest was already solid: timeout null handling, log serialization extraction, client-side rendering with 300-line cap, dual abort cleanup path.

_Review by @Clayton_

Peyton-Spencer (commented):
Triage review complete.

- The follow-up fixes address the main review concerns: the `globalThis.fetch` mock is restored in `afterEach`, and `/api/logs/stream` now enforces the same SSE client cap pattern as `/api/events`.
- The remote log proxy flow is coherent end to end: peer route allowance, raw log stream endpoint, browser-side stream handling, and tests for both happy path and connection limit.
- I do not see a new blocker in the feature diff.

This is a stacked PR on top of #312, so merge readiness depends on that base landing cleanly first. Also, the only visible status here is CodeRabbit, which is not a full CI substitute, so I would still want the normal test signal before merge.

Peyton-Spencer (commented):
I reviewed the incremental diff here and validated the core paths locally.

What I checked:
- PR description vs diff
- local `bun test src/web/server.test.ts src/web/log-stream.test.ts src/discovery/routes.test.ts`
- local `bunx tsc --noEmit`

The raw log stream endpoint, trusted peer proxy route, and client-side Network page wiring all line up with the test coverage in the PR.

I am not merging this yet because it targets `fix/datastar-web-audit` rather than `main`, so it is still blocked on the base branch in this stack. After `#312` lands (or this branch is rebased onto `main`), this should be ready for a final merge check.

PR #328, opened 2026-03-16
fix(routing): preserve hot-mode follow-ups

Peyton-Spencer:
## Summary
- preserve one-message hot mode when internal IPC or system messages land between a mention and the user's follow-up
- gate attentive-state arming and consumption on user-authored messages only
- add focused tests for the new attentive routing helper

## Testing
- `bun test src/attentive-routing.test.ts`
- `bun run typecheck`
- `bun run format:check`

### DIFF
--- a/src/attentive-routing.test.ts
+++ b/src/attentive-routing.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it } from 'bun:test';
+
+import {
+  hasAttentiveEligibleMessage,
+  isAttentiveEligibleMessage,
+} from './attentive-routing.js';
+
+describe('isAttentiveEligibleMessage', () => {
+  it('accepts user-authored chat messages', () => {
+    expect(
+      isAttentiveEligibleMessage({
+        sender: 'discord:123',
+        sender_name: 'Alice',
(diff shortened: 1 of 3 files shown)


Peyton-Spencer (commented):
Good root cause identification — IPC/system messages consuming the attentive window is a real vector. The extraction into `attentive-routing.ts` is clean and testable.

However, I think this fixes one of *two* issues causing the "all agents wake up" behavior.

**TL;DR:** This PR fixes attentive state being prematurely consumed. But even when attentive state survives correctly, the follow-up still wakes fallback agents *alongside* the attentive agent due to the merge behavior at the existing `selectedSubs = [...selectedSubs, s]` block (unchanged in this PR). When attentive agents exist, they should *replace* the fallback selection, not merge with it.

Suggested follow-up — change the attentive merge to a replace:
```typescript
if (attentiveSubs.length > 0) {
  // Replace fallback selection — only attentive agents should respond
  selectedSubs = attentiveSubs;
  // Consume attentive state...
}
```

This way, the follow-up routes exclusively to the agent that was mentioned.

Inline notes below.

PR #345, opened 2026-03-18
fix(security): redact Telegram avatar tokens from web APIs

### LINKED ISSUE #341: security: Telegram bot token leaks via avatar URLs in Web UI and discovery APIs
## Severity

High (credential disclosure)

## Summary

Telegram avatar URLs are constructed with the raw bot token embedded in the path (`https://api.telegram.org/file/bot<TOKEN>/...`). OmniClaw then stores and re-serializes those URLs through Web UI and discovery APIs, so any authenticated dashboard user or trusted peer can recover the bot token.

## Affected flow

- `src/channels/telegram.ts` returns avatar/chat image URLs that include `this.botToken`
- `src/avatar-sync.ts` persists those URLs into agent records via `updateAgentAvatar(...)`
- `src/web/agent-channels.ts` includes `avatarUrl` in `/api/agents` responses and dashboard topology JSON
- `src/web/routes.ts` returns raw `avatarUrl` from `GET /api/agents/:id/avatar`
- trusted peer sync also consumes `/api/agents`, so paired instances can receive the token-bearing URL too

## Why this matters

Telegram bot tokens are bearer credentials. Anyone who obtains the token can act as that bot via the Telegram Bot API.

Even if the Web UI is authenticated, this is still a secret-handling bug because:

- the dashboard should not expose upstream service credentials to viewers
- trusted peers receive the same serialized data during discovery sync
- the token is stored in SQLite as part of `agents.avatar_url`, increasing accidental disclosure risk in logs/backups/debug dumps

## Reproduction

1. Configure a Telegram bot and start OmniClaw with the Web UI enabled.
2. Let avatar sync run, or call any code path that stores Telegram avatar URLs.
3. Request `GET /api/agents` or `GET /api/agents/{agentId}/avatar` with valid Web UI credentials.
4. Observe that the returned `avatarUrl` contains a URL like:

```text
https://api.telegram.org/file/bot123456:ABCDEF.../photos/file_42.jpg
```

5. The embedded token is usable as the bot's API credential.

## Recommended fix

- never persist Telegram token-bearing URLs in DB/API payloads
- store a token-free descriptor instead (for example provider + file path, or an internal proxy URL)
- ensure `/api/agents` and related dashboard payloads expose only server-local proxy URLs
- keep token-bearing fetches server-side only
- consider rotating any Telegram bot tokens that may already have been exposed through the dashboard or peer sync

## Notes

Discord and Slack avatar URLs do not have this problem because their returned URLs are not bearer credentials. This issue is specific to the Telegram `file/bot<TOKEN>/...` URL format.

Peyton-Spencer:
## Summary
- replace token-bearing Telegram agent avatar URLs with token-free `tg-file:` descriptors before persisting or serializing them
- resolve Telegram avatar descriptors back to real Bot API file URLs only on the server when proxying avatar images
- add regression coverage for descriptor parsing, Web UI redaction, and server-side avatar proxying

## Testing
- `bun test src/channels/telegram.test.ts src/web/routes.test.ts src/web/agent-channels.test.ts src/web/agent-detail.test.ts src/web/dashboard.test.ts`
- `bun run typecheck`

Closes #341.

### DIFF
--- a/src/avatar-sync.ts
+++ b/src/avatar-sync.ts
@@ -174,7 +174,10 @@ export async function syncAvatars(
 
     try {
       const fetchedUrl = await candidate.channel.getAvatarUrl();
-      const safeUrl = sanitizeTelegramAvatarUrl(fetchedUrl || undefined, candidate.platform);
+      const safeUrl = sanitizeTelegramAvatarUrl(
+        fetchedUrl || undefined,
+        candidate.platform,
+      );
       if (safeUrl && safeUrl !== agent.avatarUrl) {
         updateAgentAvatar(agent.id, safeUrl, candidate.platform);
         agent.avatarUrl = safeUrl;
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -455,7 +455,9 @@ export class TelegramChannel implements Channel {
     }
   }
 
-  async resolveStoredAvatarUrl(storedAvatarUrl: string): Promise<string | null> {
+  async resolveStoredAvatarUrl(
+    storedAvatarUrl: string,
+  ): Promise<string | null> {
     const descriptor = parseTelegramFileDescriptor(storedAvatarUrl);
     if (descriptor) {
       if (descriptor.botId !== this.botId) return null;
--- a/src/index.ts
+++ b/src/index.ts
@@ -2339,7 +2339,10 @@ async function main(): Promise<void> {
       fs.writeFileSync(resolved, content, 'utf-8');
     },
     updateAgentAvatar: (agentId, url, source) => {
-      const safeUrl = sanitizeTelegramAvatarUrl(url || undefined, source || undefined);
+      const safeUrl = sanitizeTelegramAvatarUrl(
+        url || undefined,
+        source || undefined,
+      );
       updateAgentAvatar(agentId, safeUrl || null, source);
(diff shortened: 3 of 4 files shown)


src/telegram-avatar.ts
```diff
@@ -0,0 +1,69 @@
...
+): string {
+  return `https://api.telegram.org/file/bot${botToken}/${filePath}`;
+}
+
+export function sanitizeTelegramAvatarUrl(
+  avatarUrl: string | undefined,
+  _avatarSource?: string,
+): string | undefined {
+  if (!avatarUrl) return avatarUrl;
+  const parsed = parseTelegramApiFileUrl(avatarUrl);
```
Peyton-Spencer: This currently sanitizes any Telegram Bot API file URL, even when the avatar source is `custom`. That means a user who intentionally sets a custom avatar to a Telegram-hosted image will get `tg-file:...` persisted, but `resolveAgentAvatarUrl()` only resolves descriptors for `avatarSource === "telegram"`, so `/api/agents/:id/avatar/image` later tries to fetch the descriptor string as a URL and the avatar breaks. I think this helper should either respect `avatarSource` (only rewrite when it is `telegram`, or possibly when the source is absent and we are redacting outbound data) or the callers that persist values should gate the rewrite themselves.

PR #347, opened 2026-03-18
docs: refresh README for current OmniClaw architecture

Peyton-Spencer:
## Summary
- rewrite the top-level README to match the current OmniClaw product instead of the older WhatsApp-only, skill-only positioning
- document the built-in web UI, multi-channel support, multiple runtimes, scheduling, trusted peer discovery, and Bun-based development workflow
- clarify which parts of the old philosophy still hold and which are now outdated given the current architecture

## Testing
- `bunx prettier --check README.md`


README.md
```diff
@@ -3,177 +3,303 @@
 </p>
 
 <p align="center">
-  Personal Claude assistant from the Omni ecosystem. Runs securely in containers with filesystem isolation.
+  Agent infrastructure for the software factory era: multi-channel, multi-agent, container-isolated, and built to operate itself.
```
Peyton-Spencer: "built to operate itself" — this is aspirational. OmniClaw does not currently operate itself autonomously. Agents run tasks on schedules and respond to messages, but there's no self-healing, self-deploying, or self-modifying loop in the codebase. Consider something more grounded like "built for daily operations" or just drop the clause.

Issue #335, opened 2026-03-16
bug: WiFi discovery / trusted WiFi gating not working — silent detection failures


Peyton-Spencer:
## Problem

WiFi discovery / trusted WiFi gating is not working (reported by Peyton).

## Analysis

The discovery feature depends on `detectCurrentNetwork()` in `src/discovery/network-identity.ts` to identify the current WiFi SSID. The `DiscoveryRuntimeController.recomputeActive()` logic is:

```typescript
const currentActive =
  this.enabled &&
  (this.trustedNetworks.length === 0 ||
    (!!this.currentNetwork?.id &&
      this.trustedNetworks.some(item => item.id === this.currentNetwork?.id)));
```

If `detectCurrentNetwork()` returns `null` and trusted networks are configured, `active` stays `false` and discovery never starts.

### Potential failure points

1. **macOS command deprecation**: `networksetup -getairportnetwork` was deprecated in macOS 14+ Sequoia. If the command fails, `runCommand` returns `''` silently (errors only logged at `debug` level in `src/discovery/network-identity.ts:109`).

2. **WiFi device name mismatch**: `getMacWifiDevice()` matches `Hardware Port: (Wi-Fi|AirPort)`. Apple Silicon Macs or newer macOS versions may use different port names.

3. **Silent failure masking**: The `runCommand` helper catches all errors and returns `''`, and `detectCurrentNetwork` catches further with `logger.warn`. The debug-level logging means these failures are invisible in normal log output.

4. **Polling timing**: `refresh()` is called on a 15s interval. If the first `refresh()` at startup fails, discovery won't retry until the next poll tick.

### Suggested fixes

- Bump the `runCommand` failure log from `debug` to `warn` so detection failures are visible
- Add fallback detection for macOS 15+: try `system_profiler SPAirPortDataType` or `wdutil info` if `networksetup` fails
- Add a `/api/discovery/debug` endpoint that returns the raw `detectCurrentNetwork()` result for troubleshooting
- Consider logging the detection result in `refresh()` at info level on state change

## Related files

- `src/discovery/network-identity.ts` — WiFi SSID detection (`detectCurrentNetwork`)
- `src/discovery/runtime.ts` — `DiscoveryRuntimeController`, `recomputeActive()`
- `src/index.ts:2496` — Controller initialization and startup
- `src/web/network.ts` — Network page UI

PR #352, opened 2026-03-18
fix(discovery): add modern macOS Wi-Fi fallbacks

Peyton-Spencer:
## Summary
- add macOS Wi-Fi SSID fallbacks so discovery can still identify the current network when `networksetup -getairportnetwork` no longer works reliably
- broaden Wi-Fi hardware port matching and surface command failures at `warn` level so trusted-network gating failures are debuggable
- add regression coverage for `networksetup`, `system_profiler`, and `wdutil` detection paths plus failure logging

## Testing
- `bun test src/discovery`
- `bun run typecheck`

Refs #335.

### DIFF
--- a/src/discovery/network-identity.ts
+++ b/src/discovery/network-identity.ts
@@ -76,7 +76,9 @@ async function getMacWifiDevice(): Promise<string | null> {
   const blocks = output.split(/\n\n+/);
 
   for (const block of blocks) {
-    if (!/Hardware Port: (Wi-?Fi|AirPort|Wireless LAN|WLAN|802\.11)/i.test(block)) {
+    if (
+      !/Hardware Port: (Wi-?Fi|AirPort|Wireless LAN|WLAN|802\.11)/i.test(block)
+    ) {
       continue;
     }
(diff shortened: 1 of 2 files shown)


src/discovery/network-identity.ts
```diff
@@ -81,12 +111,78 @@ export function getMacNetworksetupCommand(): string[] {
   return [MAC_NETWORKSETUP_PATH];
 }
 
+function toWifiIdentity(ssid: string): DiscoveryNetworkIdentity | null {
+  const normalized = normalizeSsid(ssid);
+  if (!normalized) return null;
```
Peyton-Spencer: Nit: `toWifiIdentity` calls `normalizeSsid(ssid)` again, but all three parse functions (`parseMacNetworksetupSsid`, `parseSystemProfilerSsid`, `parseWdutilSsid`) already normalize before returning. The double call is harmless (idempotent) but could be simplified — either have the parse functions return raw strings and normalize only here, or drop the normalize call in `toWifiIdentity`. Not blocking.

PR #362, opened 2026-03-18
fix(security): block private avatar image fetch targets

Peyton-Spencer:
## Summary
- block loopback, link-local, and other private-network avatar URLs before they are persisted or fetched by the web image proxy
- add DNS-aware remote image validation plus a fetch timeout in the shared remote image cache path
- add regression tests covering blocked avatar submissions and blocked remote image fetches

## Security
Authenticated web users could previously save a custom avatar URL like `http://127.0.0.1:8080/private.png` and then trigger `/api/agents/:id/avatar/image` to make OmniClaw fetch internal resources. This change closes that SSRF path at both write time and fetch time.

## Testing
- `bun test src/web/image-cache.test.ts src/web/routes.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **New Features**
  * Validates remote image URLs, rejecting localhost and private IP addresses for avatars and remote images.
  * Adds timeout protection for remote image fetches to reduce hangs.

* **Bug Fixes**
  * Server now rejects disallowed avatar URLs with a clear 400 error instead of accepting or proxying them.

* **Tests**
  * Added comprehensive tests covering loopback, private ranges, embedded credentials, DNS failures, and successful public HTTPS cases.

### DIFF
--- a/src/web/routes.ts
+++ b/src/web/routes.ts
@@ -5,7 +5,10 @@ import { GROUPS_DIR } from '../config.js';
 import { assertPathWithin } from '../path-security.js';
 import type { ScheduledTask } from '../types.js';
 import type { WebStateProvider } from './types.js';
-import { serveCachedRemoteImage, validateRemoteImageUrl } from './image-cache.js';
+import {
+  serveCachedRemoteImage,
+  validateRemoteImageUrl,
+} from './image-cache.js';
 import { renderAgentDetail, buildAgentDetailData } from './agent-detail.js';
 import { renderConversations } from './conversations.js';
 import {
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
Triage pass: this looks merge-ready from my side.

- Scope is tight and the defense-in-depth approach makes sense: URL validation on write plus fetch-time validation for stored avatar URLs.
- Current CI is green (`format`, `test`, `typecheck`, `validate-skill-only`).
- Regression coverage looks good in `src/web/image-cache.test.ts` and `src/web/routes.test.ts`.

Only remaining note is the documented DNS rebinding / TOCTOU gap in the fetch path, but that is called out explicitly and does not block this PR for me.

PR #386, opened 2026-03-23
fix(security): validate external MCP server config

### LINKED ISSUE #378: security: external MCP server config can overwrite built-in omniclaw server
## Summary

PR #361 shipped external MCP server support, enabling agents to use custom MCP servers via container config. The CodeRabbit review flagged a security concern that hasn't been addressed: user-configured `mcpServers` entries can overwrite the built-in `omniclaw` MCP server because the spread order places user config after the hardcoded definition.

## Problem

In `container/agent-runner/src/index.ts`, the MCP server config merges as:

```ts
mcpServers: {
  omniclaw: { /* built-in */ },
  ...(containerInput.mcpServers || {}),  // can overwrite omniclaw
}
```

A config-provided key named `omniclaw` replaces the internal server entirely, breaking core IPC/routing at runtime.

Additionally:
- External MCP servers are injected but `allowedTools` is not extended with `mcp__<servername>__*` patterns, so the tools may be blocked at the SDK level
- No validation that MCP server `command` paths are safe (path traversal, arbitrary binary execution)

## Impact

- A typo or intentional `omniclaw` key in external MCP config silently breaks agent internals
- External MCP tools may not actually work if allowedTools isn't extended
- No input validation on MCP server command/args could be a container escape vector

## Suggested Fix

1. Spread user MCP servers first, then define `omniclaw` to always win
2. Dynamically extend `allowedTools` for each external MCP server key
3. Validate MCP server command paths (reject absolute paths outside container, reject traversal)

## References

- PR #361 (merged): feat: support external MCP servers via agent container config
- CodeRabbit review comment on PR #368 flagged the spread-order issue

Peyton-Spencer:
## Summary
- validate external MCP server names and config shapes before passing them into the agent SDK
- reject reserved `omniclaw` collisions and block stdio command paths that escape mounted workspaces
- extend `allowedTools` for accepted external MCP servers and add focused regression tests

## Testing
- `bun test ./src/__tests__/mcp-config.test.ts` (in `container/agent-runner/`)
- `bun test ./src/__tests__/security-hooks.test.ts` (in `container/agent-runner/`)
- `bunx tsc --noEmit -p container/agent-runner/tsconfig.json`

Closes #378.

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -156,7 +156,9 @@ export function validateExternalMcpCommand(
 ): string {
   const trimmed = command.trim();
   if (!trimmed) {
-    throw new Error(`External MCP server '${serverName}' must define a command`);
+    throw new Error(
+      `External MCP server '${serverName}' must define a command`,
+    );
   }
   if (trimmed.includes('\0')) {
     throw new Error(
@@ -184,7 +186,10 @@ export function validateExternalMcpCommand(
     : path.resolve('/workspace/group', trimmed);
   const isAllowed = ALLOWED_EXTERNAL_MCP_COMMAND_ROOTS.some((root) => {
     const resolvedRoot = path.resolve(root);
-    return resolved === resolvedRoot || resolved.startsWith(`${resolvedRoot}${path.sep}`);
+    return (
+      resolved === resolvedRoot ||
+      resolved.startsWith(`${resolvedRoot}${path.sep}`)
+    );
   });
   if (!isAllowed) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


container/agent-runner/src/index.ts
```diff
@@ -80,6 +80,214 @@ interface ContainerOutput {
...
+  serverName: string,
+  command: string,
+): string {
+  const trimmed = command.trim();
+  if (!trimmed) {
+    throw new Error(`External MCP server '${serverName}' must define a command`);
+  }
+  if (trimmed.includes('\0')) {
+    throw new Error(
+      `External MCP server '${serverName}' command contains invalid null bytes`,
```
Peyton-Spencer: The `env` field is passed through to the MCP server process without screening. A config like:
```json
{ "env": { "LD_PRELOAD": "/tmp/evil.so", "PATH": "/tmp:$PATH" } }
```
could hijack the spawned process. Inside a container the blast radius is limited, but it's still worth considering:
1. An explicit blocklist for dangerous env vars (`LD_PRELOAD`, `LD_LIBRARY_PATH`, `DYLD_*`, `PATH`), or
2. A documentation comment noting that container isolation is the mitigation here

Not a blocker if the container boundary is trusted, but worth documenting the threat model.

PR #436, opened 2026-03-27
feat(scheduler): inject recent handoffs into chat prompts

Peyton-Spencer:
## Summary
- read the most recent scheduled-run handoff JSON files for a group and inject a compact summary block into fresh chat-run prompts
- cap the prompt context to the last three handoffs and truncate long result or error text to 200 characters
- skip unreadable or invalid handoff files with warnings instead of blocking chat runs

## Testing
- `bun test src/task-handoffs.test.ts`
- `bunx tsc --noEmit`
- `bunx prettier --check src/task-handoffs.ts src/task-handoffs.test.ts src/index.ts`
- `bun run build`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -126,6 +126,10 @@ import {
   hasPriorRuntimeState,
   STARTUP_CONFIRMATION_PROMPT,
 } from './startup-notifications.js';
+import {
+  formatScheduledRunHandoffsForPrompt,
+  readScheduledRunHandoffs,
+} from './task-handoffs.js';
 import { startSchedulerLoop } from './task-scheduler.js';
 import {
   parseTelegramApiFileUrl,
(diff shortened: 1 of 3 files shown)


Peyton-Spencer (commented):
Overall this is clean and well-scoped. Tests pass, types check, security is handled correctly with `assertPathWithin`. A few observations:

**Correctness** — Looks solid. The sort-by-filename-descending approach works because filenames are ISO-timestamp-prefixed, so lexicographic order = chronological order. The `limit: 3` + `slice(0, limit)` is correct.

**Security** — `getHandoffDir` validates against traversal via `assertPathWithin`, and each file path is also validated. Good. Since we're reading from a directory the orchestrator itself wrote (not user-supplied filenames from chat), the attack surface is minimal.

**Error handling** — `onError` callback + `flatMap` returning `[]` on failure is a good pattern. Unreadable/corrupt files are skipped without blocking chat runs.

**Performance** — `readdirSync` + `readFileSync` is synchronous I/O on the hot path (every chat run). With the 50-file cap from pruning and the 3-file read limit, this is fine in practice. If handoff volume grows significantly, consider async reads, but not worth optimizing now.

**Minor items** noted in inline comments below.

PR #451, opened 2026-03-28
test: cover IPC snapshot writers

Peyton-Spencer:
## Summary
- add deterministic filesystem-backed tests for `src/ipc-snapshots.ts` instead of reimplementing its filtering logic inside tests
- verify `writeTasksSnapshot`, `writeGroupsSnapshot`, and `writeRostersSnapshot` serialize the expected payloads and respect main-vs-agent visibility rules
- freeze time in snapshot tests so `lastSync` assertions stay stable across runs

## Coverage
- tests: 1545 -> 1546
- overall line coverage: 82.24% -> 83.26%
- overall function coverage: 85.54% -> 86.32%
- `src/ipc-snapshots.ts`: 19.57% -> 100.00% line coverage, 40.00% -> 100.00% function coverage

## Verification
- `bun test src/ipc-snapshots.test.ts`
- `bun test`
- `bun test --coverage --coverage-reporter=text`

## Remaining gaps
- `src/channels/discord.ts`
- `src/discovery/mdns.ts`
- `src/channels/slack.ts`
- `src/channels/telegram.ts`

## Summary by CodeRabbit

* **Tests**
  * Enhanced test infrastructure with snapshot-based verification for improved reliability and maintainability.
  * Implemented deterministic test timing controls to ensure consistent test execution.
  * Expanded test isolation and cleanup procedures.
  * Increased test coverage for roster data persistence and validation.

### DIFF
--- a/src/ipc-snapshots.test.ts
+++ b/src/ipc-snapshots.test.ts
@@ -125,7 +125,12 @@ describe('ipc-snapshots', () => {
   describe('task visibility filtering', () => {
     it('writes only matching tasks for non-main groups', () => {
       const groupFolder = uniqueFolder('ipc-snapshot-tasks');
-      const tasksFile = path.join(DATA_DIR, 'ipc', groupFolder, 'current_tasks.json');
+      const tasksFile = path.join(
+        DATA_DIR,
+        'ipc',
+        groupFolder,
+        'current_tasks.json',
+      );
       const allTasks = [
         {
           id: 't1',
@@ -189,7 +194,12 @@ describe('ipc-snapshots', () => {
 
     it('writes all tasks for the main group', () => {
       const groupFolder = uniqueFolder('ipc-snapshot-tasks-main');
-      const tasksFile = path.join(DATA_DIR, 'ipc', groupFolder, 'current_tasks.json');
+      const tasksFile = path.join(
+        DATA_DIR,
+        'ipc',
+        groupFolder,
+        'current_tasks.json',
+      );
       const allTasks = [
         {
           id: 't1',
@@ -321,7 +331,13 @@ describe('ipc-snapshots', () => {
       const subscribedJids = new Set(['j1', 'j3']);
 
       try {
-        writeGroupsSnapshot(groupFolder, false, groups, new Set(), subscribedJids);
+        writeGroupsSnapshot(
+          groupFolder,
+          false,
+          groups,
+          new Set(),
+          subscribedJids,
+        );
         expect(readJson(groupsFile)).toEqual({
           groups: [groups[0], groups[2]],
           lastSync: '2026-03-28T12:34:56.000Z',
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
Nice improvement — testing the actual production functions against the filesystem instead of reimplementing filtering logic is much more trustworthy. Coverage jump from 19.57% to 100% on this module is well-earned. A few observations below.

src/ipc-snapshots.test.ts
```diff
@@ -185,12 +274,30 @@ describe('ipc-snapshots', () => {
           isRegistered: false,
         },
       ];
-      const isMain = true;
-      const visibleGroups = isMain ? groups : [];
-      expect(visibleGroups).toHaveLength(2);
+
+      try {
+        writeGroupsSnapshot(groupFolder, true, groups, new Set(['j1', 'j2']));
```
Peyton-Spencer: Production observation surfaced by this test: `registeredJids` is accepted as a required parameter by `writeGroupsSnapshot` but is **never referenced** in the function body — the filtering only uses `subscribedJids`. This `new Set(['j1', 'j2'])` is silently ignored.

Not a test issue, but worth a follow-up to either use or remove that parameter in `ipc-snapshots.ts`.

src/ipc-snapshots.test.ts
```diff
@@ -211,15 +318,68 @@ describe('ipc-snapshots', () => {
...
+      } finally {
+        fs.rmSync(path.join(DATA_DIR, 'ipc', groupFolder), {
+          recursive: true,
+          force: true,
+        });
+      }
+    });
+  });
+
+  describe('writeRostersSnapshot', () => {
```
Peyton-Spencer: Consider two lightweight additions:

1. **Empty rosters** — `writeRostersSnapshot(folder, [])` to verify the degenerate case serializes correctly.
2. **`assertPathWithin` rejection** — all three writers call `assertPathWithin` as a security invariant. A test like `expect(() => writeRostersSnapshot('../evil', [])).toThrow()` would be high-value for ~2 lines of code and confirms the security boundary holds through these entry points.

Fine to leave for a follow-up if you'd rather keep this PR scoped.

src/ipc-snapshots.test.ts
```diff
@@ -1,7 +1,46 @@
...
+
+function uniqueFolder(prefix: string): string {
+  return `${prefix}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+}
+
+function readJson(filePath: string): unknown {
+  return JSON.parse(fs.readFileSync(filePath, 'utf8'));
+}
+
+afterEach(() => {
```
Peyton-Spencer: Good — top-level `afterEach` ensures the global `Date` is always restored even if a test throws. One edge case: if a test fails *inside* `installFixedDate` before the override takes effect, `globalThis.Date` is still `RealDate`, so the restore is a no-op. Correct behavior — just noting it's sound.

PR #452, opened 2026-03-28
fix(security): cap GitHub webhook request bodies

### LINKED ISSUE #447: security: GitHub webhook server lacks pre-auth request body limits
## Summary

The standalone GitHub webhook server reads the entire request body with `await req.text()` before it verifies `X-Hub-Signature-256`, and it does not enforce any request-size cap first.

## Severity

- P1 / High

## Affected areas

- `src/github-webhooks.ts:261`
- `src/github-webhooks.ts:273`
- `src/github-webhooks.ts:275`
- `src/index.ts:2654`
- `src/config.ts:258`

## Root cause

`startGitHubWebhookServer()` accepts POSTs on the configured webhook port and path, then immediately does:

```ts
const rawBody = await req.text();
```

Only after fully buffering that body does it verify the HMAC signature:

```ts
if (!verifyGitHubWebhookSignature(rawBody, signature, options.secret)) {
  return new Response('Invalid signature', { status: 401 });
}
```

There is no `Content-Length` check, streamed byte cap, or early `413` rejection before body buffering.

## Impact

- Any attacker who can reach the configured webhook port/path can send oversized POST bodies without knowing the webhook secret.
- OmniClaw will buffer the full payload into memory and then compute an HMAC before rejecting it, creating avoidable memory and CPU pressure on a public-facing endpoint.
- Repeated requests can degrade or crash a deployment that enables GitHub webhooks without an external proxy limit.

## Reproduction

1. Configure OmniClaw with `GITHUB_WEBHOOK_PORT` and `GITHUB_WEBHOOK_SECRET` so the webhook listener starts.
2. Expose that port directly or through a proxy that does not already enforce a small body-size limit.
3. Send a large POST request to `GITHUB_WEBHOOK_PATH` (default `/webhooks/github`) with any invalid or empty `X-Hub-Signature-256` value.
4. Observe that the process still reads the entire body before returning `401 Invalid signature`.
5. Repeat with multiple large requests and watch process memory and CPU usage rise before rejection.

## Suggested fix

- Enforce a hard maximum webhook body size before calling `req.text()`.
- Prefer streaming the request body into a bounded buffer or hashing stream so oversized payloads fail with `413 Payload Too Large`.
- Add regression coverage for:
  - a normal signed webhook payload succeeding
  - an oversized payload being rejected before signature verification
  - a missing/invalid `Content-Length` path still honoring the streamed byte cap

## Notes

- This is distinct from `#317` and `#365`, which cover replay handling rather than request-size enforcement.
- Reverse-proxy body limits help, but the server should still defend itself in-process because the webhook listener is an independently exposed network surface.

Peyton-Spencer:
## Summary

- cap GitHub webhook request bodies before HMAC verification so unauthenticated clients cannot force unbounded buffering
- enforce both a `Content-Length` precheck and a streamed byte cap for missing or misleading length headers
- add regression coverage for valid signed payloads, oversize declared payloads, and oversize streamed payloads

## Testing

- `bun test src/github-webhooks.test.ts`
- `bun run tsc --noEmit`

## Security Review Log

- reviewed `src/github-webhooks.ts`, `src/config.ts`, and `src/index.ts`
- confirmed issue `#447` as P1 / High
- remaining trust-boundary review targets: `src/web/routes.ts`, `src/discovery/routes.ts`, `src/simtest/admin-api.ts`

Fixes #447

## Summary by CodeRabbit

* **New Features**
  * GitHub webhook server now enforces a configurable maximum request body size limit (default: 256KB).
  * Oversized requests are rejected with HTTP 413 (Payload Too Large).

* **Tests**
  * Added comprehensive test coverage for webhook request body size validation and error handling.


Peyton-Spencer:
Triage check: the fix addresses the right trust boundary. The two-stage defense (header precheck plus streamed byte cap) is the correct shape for issue #447, and the regression coverage hits the key signed / declared-oversize / streamed-oversize paths.

Not merge-ready yet for two reasons:
1. `format` is currently failing in CI.
2. `src/github-webhooks.test.ts` now unconditionally removes `github-watches.json` in `afterAll`, which looks like a plausible test-isolation footgun if another suite is saving/restoring that file concurrently.

After a scoped cleanup fix and a formatting pass, this should be ready.

Peyton-Spencer (commented):
I reviewed the webhook body-cap fix and the approach looks sound. The implementation now rejects oversized GitHub webhook payloads both from a declared `Content-Length` and while streaming the body, which closes the unauthenticated buffering risk called out in `#447`. The added regression coverage also exercises the right trust-boundary cases.

Local validation from `/workspace/group/omniclaw-merger-20260328`:
- `bun test src/github-webhooks.test.ts`
- `bunx tsc --noEmit`
- `bunx prettier --check src/config.ts src/github-webhooks.ts src/github-webhooks.test.ts src/index.ts`

The targeted tests and typecheck passed locally. I am not merging yet because the PR is still red on formatting, and local Prettier also flags `src/config.ts` and `src/github-webhooks.ts`. After those files are formatted and CI is green, I would be comfortable merging this security fix.

PR #485, opened 2026-04-01
test: expand logger and backend factory coverage

Peyton-Spencer:
## Summary
- Expand deterministic coverage for `src/logger.ts` around pretty output formatting, subscriber fan-out, unsubscribe behavior, and debug stack capture.
- Add backend factory tests for default initialization, backend deduplication, clean shutdown, and shutdown warning handling in `src/backends/index.ts`.
- Increase suite coverage without touching production code paths.

## Verification
- `bun test`
- `bun test --coverage`

## Coverage Notes
- Test count before: 1623 tests across 90 files.
- Test count after: 1633 tests across 90 files.
- Coverage before: 86.83% funcs / 83.58% lines.
- Coverage after: 87.79% funcs / 84.76% lines.
- Improved files: `src/logger.ts` to 81.25% funcs / 73.12% lines, `src/backends/index.ts` to 83.33% funcs / 95.24% lines.

## Remaining Gaps
- Lowest remaining coverage reported in the post-change summary: `src/web/solid-handler.ts` and `src/web/server.ts`.

## Summary by CodeRabbit

* **Tests**
  * Strengthened backend lifecycle tests to verify proper initialization of default backends, deduplication of multiple mappings, and graceful shutdown handling with error resilience.
  * Expanded logger test coverage for JSON formatting, terminal output rendering with metric suffixes and truncation, and subscription notification behavior.

### DIFF
--- a/src/logger.test.ts
+++ b/src/logger.test.ts
@@ -390,10 +390,7 @@ describe('logger', () => {
     });
 
     it('truncates long tags to 16 characters in pretty mode', () => {
-      const logger = createLogger(
-        { group: '1234567890abcdefghijk' },
-        'info',
-      );
+      const logger = createLogger({ group: '1234567890abcdefghijk' }, 'info');
       logger.info('hello');
 
       expect(stderrOutput).toHaveLength(1);
--- a/src/logger.test.ts
+++ b/src/logger.test.ts
@@ -1,14 +1,6 @@
-import {
-  describe,
-  it,
-  expect,
-  beforeEach,
-  afterEach,
-  mock,
-  spyOn,
-} from 'bun:test';
+import { describe, it, expect, beforeEach, afterEach, mock } from 'bun:test';
 
-import { createLogger, type Logger } from './logger.js';
+import { createLogger } from './logger.js';
 
 describe('logger', () => {
   let stderrOutput: string[];
@@ -177,6 +169,25 @@ describe('logger', () => {
       const parsed = JSON.parse(stderrOutput[0]);
       expect(parsed.container).toBe('test-container');
     });
+
+    it('includes error stacks when LOG_LEVEL is debug', () => {
+      const originalLevel = process.env.LOG_LEVEL;
+      process.env.LOG_LEVEL = 'debug';
+
+      const logger = createLogger({}, 'error');
+      logger.error({ err: new Error('with stack') }, 'boom');
+
+      const parsed = JSON.parse(stderrOutput[0]);
+      expect(parsed.err).toBe('with stack');
+      expect(typeof parsed.errStack).toBe('string');
+      expect(parsed.errStack).toContain('with stack');
+
+      if (originalLevel === undefined) {
+        delete process.env.LOG_LEVEL;
+      } else {
+        process.env.LOG_LEVEL = originalLevel;
+      }
+    });
   });
 
   describe('child loggers', () => {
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
## Review: PR #485 — test: expand logger and backend factory coverage

Good test expansion. The new tests are well-structured and cover meaningful production paths — pretty formatting, subscriber lifecycle, debug stack capture, backend init/shutdown. A few findings below, mostly nits and one potential issue.

### CI Note

The `format` check is currently failing. Likely needs a `bun run format` pass before merge.

### Findings Summary

**src/logger.test.ts**

1. **[suggestion] Lines 172-188 — "includes error stacks when LOG_LEVEL is debug"**: The `LOG_LEVEL` env var cleanup uses a manual `if/else/delete` pattern. Consider using a try/finally to ensure cleanup even if an assertion throws mid-test. The current pattern would leak `LOG_LEVEL=debug` into subsequent tests if `expect(parsed.errStack).toContain('with stack')` throws. (Same risk exists in Bun's test runner depending on isolation — worth being defensive.)

2. **[suggestion] Lines 172-188 — Same test**: This test creates the logger *after* setting `LOG_LEVEL=debug`, but `createLogger()` reads `process.env.LOG_LEVEL` at call time for `effectiveLevel`, while the stack-capture check happens inside `flattenError()` which reads `process.env.LOG_LEVEL` at log-write time. The test works correctly since both reads happen after the env var is set, but it would be worth adding a brief comment noting that `flattenError` checks `LOG_LEVEL` at write time (not logger creation time), since that's a subtle but important behavior distinction.

3. **[nit] Lines 349-403 — pretty output describe block**: The `isTTY` save/restore in `beforeEach`/`afterEach` is clean. However, the pretty-mode tests only check `toContain` on the string fragments. There's no assertion that the output is a single line (no embedded newlines before the trailing `\n`). Not critical, but since `writePretty` builds a single-line format, asserting `stderrOutput[0].split('\n').length === 2` (content + trailing newline) would catch any accidental multi-line regression.

4. **[suggestion] Lines 358-387 — "renders inline metric suffixes"**: Great coverage of all 6 metric suffixes. Missing one production path: the `err` field in pretty mode only renders `ERR: ...` when `record.err` is a `string`. This test passes `err: 'network'` which is already a string. Consider adding a companion case where `err` is an `Error` object to verify that `flattenError` converts it to a string *before* `writePretty` checks `typeof record.err === 'string'`.

5. **[praise] Lines 405-466 — subscriber tests**: The subscriber test suite is excellent — covers the subscribe/notify, unsubscribe, and error-swallowing behaviors cleanly. The fan-out error isolation test (`'swallows subscriber errors and continues notifying others'`) is particularly valuable since silent subscriber failures could easily hide bugs in production.

6. **[question] Lines 417-437 — "notifies subscribers with the emitted record"**: The subscriber mock returns `record` — was that intentional to test that the return value is ignored, or just leftover? The production code doesn't use the return value from `fn(record)`, so a void mock would be slightly more precise. Minor.

7. **[suggestion] Missing test — child logger subscriber inheritance**: The production code shares a `Set<LogSubscriber>` via `parentSubscribers` so child loggers inherit subscribers from parents. There's no test covering this: subscribing on a parent and logging from a child should trigger the subscriber. This is a meaningful behavior worth verifying.

**src/backends/index.test.ts**

8. **[nit] Line 20 — `afterEach(() => { mock.restore() })`**: This `mock.restore()` in `afterEach` will restore all mocks including `spyOn` calls. Since `getBackend` returns singletons from a module-level `Map`, spying on a singleton's methods in one test and then restoring in afterEach is correct. But note that the singletons themselves persist across tests (the `backends` Map is never cleared). This means the `initializeBackends({})` test (line 64) depends on both backends already being in the Map from prior `getBackend` calls. This coupling is fragile — if test ordering changes or tests run in isolation, behavior could differ. Consider documenting this dependency with a comment.

9. **[bug] Lines 64-73 — "initializes only the default backend when no entities are provided"**: This test calls `getBackend('docker')` and spies on its `initialize`, then calls `initializeBackends({})`. The test asserts `dockerInitSpy` is not called, which is correct. However, calling `getBackend('docker')` has a side effect: it inserts the docker backend into the module-level `backends` Map. This means `shutdownBackends()` (tested later) will iterate over both backends — which is fine for *that* test, but it means this test has invisible side effects on later tests. Not a bug in the assertions themselves, but the test isolation is weaker than it appears. If a future test checks "only apple-container is initialized after a clean start," it would fail because docker is already in the Map.

10. **[suggestion] Lines 78-112 — "deduplicates backend initialization across entities"**: Good test. The entity objects use `new Date().toISOString()` for `added_at`. Consider using a fixed timestamp string (e.g., `'2026-01-01T00:00:00.000Z'`) for determinism — `new Date()` in a test creates a subtle time dependency. Not a correctness issue here since `added_at` isn't used by `initializeBackends`, but it's a hygiene pattern.

11. **[suggestion] Lines 78-112 — Same test**: Alpha has no `backend` field, so `getBackendType` defaults it to `'apple-container'`. It would strengthen the test to add an explicit comment like `// alpha: no backend field → defaults to apple-container` to make the intent clear to future readers.

12. **[praise] Lines 120-141 — shutdown failure handling test**: Excellent test. Verifying that `shutdownBackends` continues to the next backend after a failure and logs the correct structured warning is exactly the kind of resilience test that catches real production issues. The assertion on the exact `logger.warn` call shape (`{ backend: 'apple-container', error: failure }`) is appropriately strict.

13. **[suggestion] Missing test — `initializeBackends` with entities that all specify the same backend**: The dedup test covers apple-container + docker. A case where all entities specify `backend: 'docker'` (no default needed) would verify that the default backend is NOT initialized when entities explicitly specify backends. Looking at the production code: the default is only added when `neededTypes.size === 0`, so if 3 entities all say docker, apple-container should not be initialized. Worth a test since the "default backend" logic is a distinct code path.

### Overall Assessment

Solid test expansion. The logger pretty-output and subscriber tests fill real gaps, and the backend factory tests cover the important init/shutdown lifecycle. The main area for improvement is test isolation in the backend tests (singleton side effects leaking between tests) and one missing edge case (child logger subscriber inheritance). The format CI failure needs a fix before merge.

Peyton-Spencer (commented):
Reviewed the full diff against the source implementations (`src/logger.ts` and `src/backends/index.ts`). This is a solid test coverage PR -- assertions are well-aligned with the source, cleanup/restore patterns are correct throughout, and the edge cases (shutdown failure resilience, subscriber error swallowing, unsubscribe) are valuable additions.

I would APPROVE if I could (token is the PR author). A few minor observations inline but nothing blocking -- this is good to merge.

-- Clayton

src/logger.test.ts
```diff
@@ -335,4 +346,120 @@ describe('logger', () => {
...
+      const logger = createLogger({}, 'info');
+      const subscriber = mock(() => {});
+      const unsubscribe = logger.subscribe(subscriber);
+
+      unsubscribe();
+      logger.info('after unsubscribe');
+
+      expect(subscriber).not.toHaveBeenCalled();
+    });
+
```
Peyton-Spencer: Good coverage of the error swallowing behavior. This matches the `try { fn(record); } catch { }` pattern in the source. The assertion that `goodSubscriber` is still called after `badSubscriber` throws confirms fan-out resilience.

Peyton-Spencer:
Triage pass:

- Scope is clean and the added coverage looks worthwhile: logger formatting/subscriber behavior plus backend init/shutdown lifecycle are both good targets.
- Current blocker is CI: `format` is failing while `test` and `typecheck` are green.
- Existing review comments also call out a couple of test-isolation concerns in `src/backends/index.test.ts`, but those look fixable without changing product code.

Status: not merge-ready yet. Once formatting is fixed and the outstanding test hygiene comments are addressed or explicitly waived, this should be ready.

Peyton-Spencer (commented):
Triage: the coverage work is focused on logger and backend factory seams and looks worth merging once green. Test and typecheck are already passing; the only blocker I see is the failing format check.

Peyton-Spencer (commented):
I reviewed this pass and I am not merging it yet.

- The branch still has a failing `format` check, so it does not meet the green-CI merge bar.
- Once the formatting issue is fixed and the PR is back to an all-green state, I can take another merge pass quickly.

Peyton-Spencer:
Triage: coverage-focused change with otherwise green test and typecheck. Main blocker is the failing format check.

PR #491, opened 2026-04-03
fix(security): bind pair requests to source host to prevent hijack

### LINKED ISSUE #489: security: discovery pairing request overwrite lets LAN attacker hijack trusted instance IDs
## Summary

`/api/discovery/pair` trusts caller-supplied `instanceId` values and `createPairRequest()` deduplicates pending requests by `from_instance_id` alone. A LAN attacker can submit or overwrite a pending pairing request for another instance ID, replace the callback token/public key/port with attacker-controlled values, and receive the approved pairing secret when an admin approves the request.

Severity: High (P1)

## Affected code

- `src/discovery/routes.ts:397`
- `src/discovery/trust-store.ts:213`
- `src/discovery/trust-store.ts:304`
- `src/discovery/routes.ts:669`
- `src/discovery/routes.ts:1343`

## Impact

The pairing flow treats `from_instance_id` as the peer identity before any proof that the requester actually owns that identity. Because pending requests are updated in place, an attacker on the same LAN can:

1. Pick a victim `instanceId` or wait for a legitimate request to appear.
2. POST a new `/api/discovery/pair` request with that same `instanceId` but attacker-controlled `callbackToken`, `keyAgreementPublicKey`, `name`, and `port`.
3. Cause `createPairRequest()` to overwrite the existing pending row for that instance ID.
4. Wait for an admin to approve the visible request in the Web UI.
5. Receive the encrypted shared secret at the attacker-controlled callback endpoint via `sendPairingSecretToPeer()`.

After approval, `approvePairRequest()` upserts `discovery_peers` by the spoofed `instanceId`, and `checkPeerAuth()` authenticates future peer requests using that instance ID plus the attacker-received shared secret. This lets the attacker become the trusted peer for the spoofed identity.

## Reproduction

1. Start two OmniClaw instances on the same LAN: `target` (admin approval UI) and a legitimate peer `peer-a`.
2. Have `peer-a` submit a normal pairing request to `target`, or simply learn/guess the `instanceId` that `peer-a` will use.
3. Before approval, send a forged request from an attacker-controlled machine:

```bash
curl -X POST http://TARGET_HOST:PORT/api/discovery/pair \
  -H 'Content-Type: application/json' \
  -d '{
    "instanceId": "peer-a-instance-id",
    "name": "Peer A",
    "port": 8080,
    "callbackToken": "attacker-token",
    "keyAgreementPublicKey": "ATTACKER_PUBLIC_KEY"
  }'
```

4. Observe that `pair_requests` now contains the attacker-supplied callback/public key/port for the victim `instanceId` instead of preserving the original request.
5. Approve the request from the Web UI.
6. Observe that the approval callback is sent to the attacker-controlled endpoint, and the trusted peer row is written under `peer-a-instance-id`.

## Why this happens

- `handlePairRequest()` accepts unauthenticated caller-provided `instanceId` values.
- `TrustStore.createPairRequest()` treats `from_instance_id` as unique for pending requests and overwrites the row in place.
- `handleApproveRequest()` and `approvePairRequest()` trust the stored request contents without binding them to the original requester.
- `checkPeerAuth()` later uses only the stored `instanceId` and shared secret, so the attacker now passes as the spoofed peer.

## Suggested fixes

- Do not overwrite pending requests solely by `from_instance_id`; require a new request row or bind updates to the original requester identity.
- Bind the pairing request to stronger provenance, such as a request ID + approval token returned to the original requester, or a verified possession proof tied to the key agreement public key.
- Surface request provenance clearly in the admin UI so host/IP changes cannot silently replace an existing request.
- Consider rejecting a new pending request when the same `instanceId` arrives from a different host/port unless the existing request is explicitly cancelled.

## Duplicate check

I checked the current discovery/security issue set (`#284`, `#293`, `#311`, `#341`) and did not find an existing issue covering pending request overwrite or instance ID hijack in the discovery pairing flow.

Peyton-Spencer:
## Summary

Fixes #489 — a P1 security vulnerability where a LAN attacker could hijack discovery pairing requests.

**The vulnerability**: `createPairRequest()` deduplicates pending requests by `from_instance_id` alone and blindly overwrites all fields (callback token, public key, host, port). An attacker on the same LAN could submit a forged request with a victim's `instanceId` but attacker-controlled callback details, overwriting the legitimate pending request. When an admin approved the request via the Web UI, the shared secret was sent to the attacker's endpoint.

**The fix**: Pair requests are now bound to their originating source IP. When an existing pending request exists for the same `instanceId`:
- If the new request comes from the **same host** → allowed (supports legitimate key rotation after restart)
- If from a **different host** → rejected with `PairRequestHostMismatchError` (HTTP 409)
- Once a request is **resolved** (approved/rejected) → new requests from any host are accepted normally

### Changes

- `trust-store.ts`: Added `PairRequestHostMismatchError` class; `createPairRequest()` now validates host match before updating existing pending requests
- `routes.ts`: `handlePairRequest()` catches `PairRequestHostMismatchError` and returns 409 Conflict
- `index.ts`: Export the new error class

### Test coverage

- 3 new trust-store tests: host-binding enforcement, rejection from different host, post-resolution new request
- 1 new route handler test: verifies 409 response on host mismatch
- All 1629 existing tests pass, `tsc --noEmit` clean

## Test plan

- [x] `bun test` — 1629 pass, 0 fail
- [x] `tsc --noEmit` — clean
- [ ] Verify legitimate pair requests from the same IP can still refresh
- [ ] Verify attacker from different IP gets 409 when overwriting pending request
- [ ] Verify new requests work after previous request is rejected/approved

## Summary by CodeRabbit

* **Bug Fixes**
  * Pairing requests are now strictly bound to their originating host: attempts to update an existing pending pairing from a different host are rejected with an HTTP 409 and an error indicating the request is already bound to a different host. Resolved/rejected requests can be recreated from another host. Tests added to verify these behaviors.

### DIFF
--- a/src/discovery/index.ts
+++ b/src/discovery/index.ts
@@ -1,7 +1,7 @@
 export { startDiscovery } from './mdns.js';
 export { DiscoveryRuntimeController } from './runtime.js';
 export { detectCurrentNetwork } from './network-identity.js';
-export { TrustStore } from './trust-store.js';
+export { TrustStore, PairRequestHostMismatchError } from './trust-store.js';
 export { PeerClient } from './peer-client.js';
 export {
   handleDiscoveryRequest,
--- a/src/discovery/routes.test.ts
+++ b/src/discovery/routes.test.ts
@@ -6,6 +6,7 @@ import {
   handleDiscoveryRequest,
   type DiscoveryRouteContext,
 } from './routes.js';
+import { PairRequestHostMismatchError } from './trust-store.js';
 
 const realFetch = globalThis.fetch;
 
@@ -274,6 +275,43 @@ describe('handleDiscoveryRequest', () => {
       });
       expect((res as Response).status).toBe(400);
     }
+  });
+
+  it('returns 409 when pair request is bound to a different host (#489)', async () => {
+    const req = withSocketAddress(
+      new Request('http://localhost/api/discovery/pair', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({
+          instanceId: 'victim-instance',
+          name: 'Attacker',
+          host: '10.0.0.99',
+          port: 9999,
+          callbackToken: 'attacker-callback',
+          keyAgreementPublicKey: 'attacker-pub',
+        }),
+      }),
+      '10.0.0.99',
+    );
+
+    const ctx = makeContext({
+      trustStore: {
+        isPeerTrusted: () => false,
+        createPairRequest: () => {
+          throw new PairRequestHostMismatchError(
+            'victim-instance',
+            '10.0.0.30',
+            '10.0.0.99',
+          );
+        },
+      } as any,
+    });
+
+    const res = await handleDiscoveryRequest(req, new URL(req.url), ctx);
+    expect(res).not.toBeNull();
+    expect((res as Response).status).toBe(409);
+    const body = (await (res as Response).json()) as { error: string };
+    expect(body.error).toContain('bound to a different host');
   });
 
   it('rejects oversized pair requests before creating trust-store state', async () => {
--- a/src/discovery/routes.ts
+++ b/src/discovery/routes.ts
@@ -19,7 +19,10 @@ import {
   encryptPairingSecret,
   generatePairingKeyPair,
 } from './pairing-crypto.js';
-import type { TrustStore } from './trust-store.js';
+import {
+  type TrustStore,
+  PairRequestHostMismatchError,
+} from './trust-store.js';
 import type { DiscoveryRuntimeController } from './runtime.js';
 import type {
   ContextFileEntry,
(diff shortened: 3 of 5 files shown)


Peyton-Spencer:
Triage pass: the host-binding fix looks correct and the new `routes`/`trust-store` tests cover the hijack path well. Current blocker is still CI: `test` and `typecheck` are green, but `format` is failing.

Peyton-Spencer (commented):
Triage: the host-binding fix is focused and comes with the right routes and trust-store regression coverage. Test and typecheck are green; the only blocker I see is the failing format check.

Peyton-Spencer (commented):
I reviewed this pass and I am not merging it yet.

- The branch still has a failing `format` check, so it does not meet the green-CI merge bar.
- Once the formatting issue is fixed and the PR is back to an all-green state, I can take another merge pass quickly.

Peyton-Spencer:
Triage: important P1 security fix with targeted regression coverage. The main remaining blocker is the failing format check; once formatted, this should move quickly.

Peyton-Spencer:
**Product Manager note**: This is a P1 security fix for #489 (discovery pairing hijack). The only failing check is `format` — tests and typecheck both pass. Running `bun run format` on the branch should unblock merge.

This PR should be prioritized given the severity of the underlying vulnerability.

PR #504, opened 2026-04-06
fix(security): cap trusted peer image proxy bodies

Peyton-Spencer:
## Summary

- cap authenticated discovery peer avatar and chat icon fetches at 5 MiB before buffering
- reject oversized `Content-Length` values early and enforce the same streamed byte limit when headers are absent or dishonest
- add regression tests for both early `Content-Length` rejection and streamed overflow handling

## Why

Trusted-peer image proxy helpers in `src/discovery/peer-client.ts` previously called `res.arrayBuffer()` with no byte cap. A malicious or compromised trusted peer could return a huge avatar or chat icon response and force the local instance to buffer it through discovery proxy routes.

## Validation

- `bun test src/discovery`
- `bun run typecheck`
- `bunx prettier --check "src/discovery/peer-client.ts" "src/discovery/peer-client.test.ts"`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced a 5 MiB size limit on peer-related image downloads to prevent oversized transfers
  * Improved image fetch error handling so size-limit and HTTP errors are surfaced instead of masked
  * Ensured downloads are cancelled when the size limit is exceeded

* **Tests**
  * Added tests validating size-limit enforcement for both content-length and streamed responses
  * Added tests confirming cancellation and proper error propagation when limits are exceeded

### DIFF
--- a/src/discovery/peer-client.test.ts
+++ b/src/discovery/peer-client.test.ts
@@ -4,6 +4,7 @@ import { createHash } from 'crypto';
 import { PeerClient, verifyPeerRequestSignature } from './peer-client.js';
 
 const originalFetch = globalThis.fetch;
+const MAX_PEER_PROXY_IMAGE_BYTES = 5 * 1024 * 1024;
 
 function sha256Hex(value: string): string {
   return createHash('sha256').update(value).digest('hex');
@@ -133,14 +134,25 @@ describe('PeerClient', () => {
   });
 
   it('rejects remote peer images that exceed the content-length cap', async () => {
+    let cancelled = false;
     globalThis.fetch = mock(() =>
       Promise.resolve(
-        new Response(new Uint8Array([1, 2, 3]), {
-          headers: {
-            'content-type': 'image/png',
-            'content-length': String(5 * 1024 * 1024 + 1),
+        new Response(
+          new ReadableStream<Uint8Array>({
+            start(controller) {
+              controller.enqueue(new Uint8Array([1, 2, 3]));
+            },
+            cancel() {
+              cancelled = true;
+            },
+          }),
+          {
+            headers: {
+              'content-type': 'image/png',
+              'content-length': String(MAX_PEER_PROXY_IMAGE_BYTES + 1),
+            },
           },
-        }),
+        ),
       ),
     ) as unknown as typeof globalThis.fetch;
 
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
## Security Review: Peer Proxy Image Body Cap

Solid security fix. The two-layer defense (Content-Length fast rejection + streaming byte limit via `readStreamWithByteLimit`) is the right approach. Both the avatar and chat icon proxy paths are covered, the existing `readStreamWithByteLimit` in `src/media.ts` properly cancels the stream on overflow, and the 5 MiB limit is reasonable for avatar/icon use cases.

### What works well

- **Content-Length check is advisory, not sole defense.** The streaming limit via `readStreamWithByteLimit` enforces the cap even when Content-Length is missing, zero, or lies. This is critical — a malicious peer can trivially omit or falsify the Content-Length header.
- **Stream cancellation on overflow.** `readStreamWithByteLimit` calls `reader.cancel()` before throwing, which tears down the TCP connection rather than leaving it draining.
- **The `allowedStatuses` refactor** eliminates the old blanket `catch { return null }` which was swallowing all errors including size limit violations. Now only 404 is explicitly allowed through, so the size cap errors propagate correctly.
- **Both proxy paths** (`getAgentAvatarImage` and `getChatIcon`) are covered.
- **Both rejection vectors** are tested: Content-Length early rejection and streaming overflow.

### Minor observations (non-blocking)

See inline comments for a couple of minor notes. None are blocking — this is good to merge.

**Verdict: APPROVE** — the fix is complete, the streaming path enforces the limit regardless of headers, and the tests cover both rejection modes.

Peyton-Spencer (commented):
## Security Review: Trusted Peer Image Proxy Body Cap

Thorough review of the 5 MiB cap on trusted peer image proxy responses. The approach is sound -- dual-layer defense (Content-Length early rejection + streamed byte counting) is the right pattern, and reusing the battle-tested `readStreamWithByteLimit` from `media.ts` keeps things DRY.

### What looks good

- **Two-layer enforcement**: `readBinaryResponseWithLimit` checks `Content-Length` first for fast rejection, then delegates to `readStreamWithByteLimit` for actual stream-level enforcement. This handles both honest and dishonest peers correctly.
- **Stream cancellation on overflow**: `readStreamWithByteLimit` calls `reader.cancel()` before throwing, so the TCP connection is not held open draining a multi-gigabyte response body.
- **Removing blanket try/catch**: The old code swallowed all errors as `null`. The new `allowedStatuses` pattern is strictly better -- 404 is properly handled as "no image", but actual server errors and oversized responses now correctly propagate.
- **`Number.isFinite` guard**: Correctly handles missing/unparseable Content-Length (returns `NaN` from `parseInt`, which is not finite, so the check is skipped and we fall through to stream enforcement).
- **Tests cover both rejection paths**: Content-Length early rejection and streamed overflow are tested separately with different error messages, which is useful for debugging.
- **Constant is well-named and scoped**: `MAX_PEER_PROXY_IMAGE_BYTES` is local to this module, appropriately separate from the 10 MiB `MAX_BINARY_DOWNLOAD_BYTES` in `media.ts`.

### Suggestions (non-blocking)

See inline comments for specific items. Overall this is a clean, well-scoped security fix. LGTM -- would approve if the GitHub token allowed it.

Peyton-Spencer:
Addressed the review follow-ups on this branch:

- cancel the response body when the `Content-Length` fast-path rejects an oversized image
- add exact-5-MiB success coverage for both the `Content-Length` path and streamed reads
- switch the streamed overflow regression test to a multi-chunk case so it exercises a more realistic chunked transfer

Validation:
- `bun test src/discovery`
- `bun run typecheck`
- `bunx prettier --check "src/discovery/peer-client.ts" "src/discovery/peer-client.test.ts"`

Peyton-Spencer (commented):
Triage: diff is tight and the added peer-client tests match the security hardening scope. All checks are green, so this looks ready for human security review and merge.

PR #512, opened 2026-04-08
test: cover WhatsApp auth flow branches

Peyton-Spencer:
## Summary
- add deterministic unit tests for `src/whatsapp-auth.ts`, covering CLI parsing, readline prompting, pairing-code scheduling, QR rendering, reconnect-on-515, success, and failure branches
- refactor the auth script to expose dependency-injectable helpers so the WhatsApp auth flow can be exercised without live Baileys sockets or process exits
- raise the suite by 12 tests from 1648 to 1660 total tests and keep `bun test` green in this branch

## Verification
- `bun test src/whatsapp-auth.test.ts`
- `bun test`

## Files Covered
- `src/whatsapp-auth.ts`
- `src/whatsapp-auth.test.ts`

## Remaining Coverage Gaps
- `src/simtest/index.ts` still lacks dedicated deterministic startup and shutdown tests
- `src/ipc.ts` still relies mostly on indirect coverage rather than focused helper-level tests

## Summary by CodeRabbit

* **New Features**
  * Added explicit CLI flag support for `--pairing-code` and `--phone` options to streamline authentication configuration.

* **Bug Fixes**
  * Improved error handling and reconnection logic during authentication failures.

* **Tests**
  * Added comprehensive test suite covering CLI parsing, authentication flows, QR code generation, and connection handling to ensure reliability.

### DIFF
--- a/src/whatsapp-auth.test.ts
+++ b/src/whatsapp-auth.test.ts
@@ -1,0 +1,272 @@
+import { beforeEach, describe, expect, it, mock } from 'bun:test';
+
+import { DisconnectReason } from '@whiskeysockets/baileys';
+
+import {
+  askQuestion,
+  authenticate,
+  connectSocket,
+  getCliOptions,
+} from './whatsapp-auth.js';
+
+type Handler = (...args: any[]) => void;
+
+function createSocketHarness(pairingCode = '123-456') {
+  const handlers: Record<string, Handler> = {};
+  const requestPairingCode = mock(async (_phoneNumber: string) => pairingCode);
+
+  return {
+    handlers,
+    socket: {
+      requestPairingCode,
+      ev: {
+        on: mock((event: string, handler: Handler) => {
+          handlers[event] = handler;
+        }),
+      },
+    },
+    requestPairingCode,
+  };
+}
+
+function createConnectDeps(overrides: Record<string, unknown> = {}) {
+  const timers: Array<() => void | Promise<void>> = [];
+  const socketHarness = createSocketHarness();
+  const saveCreds = mock(() => {});
+  const writeFileSync = mock(() => {});
+  const unlinkSync = mock(() => {});
+  const exit = mock((_code: number) => {});
+  const log = mock(() => {});
+  const error = mock(() => {});
+  const warn = mock(() => {});
+
+  const deps = {
+    fs: { writeFileSync, unlinkSync },
+    makeWASocket: mock((_config: unknown) => socketHarness.socket),
+    fetchLatestWaWebVersion: mock(async () => ({ version: [1, 2, 3] })),
+    makeCacheableSignalKeyStore: mock((_keys: unknown, _logger: unknown) => ({})),
+    useMultiFileAuthState: mock(async () => ({
+      state: { creds: { registered: false }, keys: {} },
+      saveCreds,
+    })),
+    qrcodeGenerate: mock((_qr: string, _options: { small: boolean }) => {}),
+    exit,
+    setTimeout: mock((handler: () => void | Promise<void>, _delay: number) => {
+      timers.push(handler);
+      return 1 as unknown as Timer;
+    }),
+    console: { log, error },
+    logger: { warn } as unknown,
+  };
+
+  return {
+    deps: { ...deps, ...overrides } as Parameters<typeof connectSocket>[3],
+    timers,
+    socketHarness,
+    writeFileSync,
+    unlinkSync,
+    exit,
+    log,
+    error,
+    warn,
+    saveCreds,
+  };
+}
+
+describe('getCliOptions', () => {
+  it('parses pairing mode and phone number flags', () => {
+    expect(
+      getCliOptions(['bun', 'src/whatsapp-auth.ts', '--pairing-code', '--phone', '14155551234']),
+    ).toEqual({
+      usePairingCode: true,
+      phoneArg: '14155551234',
+    });
+  });
+
+  it('returns falsey defaults when flags are absent', () => {
+    expect(getCliOptions(['bun', 'src/whatsapp-auth.ts'])).toEqual({
+      usePairingCode: false,
+      phoneArg: undefined,
+    });
+  });
+});
+
+describe('askQuestion', () => {
+  it('trims the answer and closes the readline interface', async () => {
+    const close = mock(() => {});
+    const createInterface = mock(() => ({
+      question: (_prompt: string, callback: (answer: string) => void) => {
+        callback('  14155551234  ');
+      },
+      close,
+    }));
+
+    await expect(askQuestion('Phone?', createInterface as never)).resolves.toBe(
+      '14155551234',
+    );
+    expect(close).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe('authenticate', () => {
+  it('cleans stale files, prompts in pairing mode, and connects with the answer', async () => {
+    const mkdirSync = mock(() => {});
+    const unlinkSync = mock(() => {
+      throw new Error('missing');
+    });
+    const askQuestionMock = mock(async () => '14155551234');
+    const connectSocketMock = mock(async (_phoneNumber?: string) => {});
+    const log = mock(() => {});
+
+    await authenticate(
+      { usePairingCode: true },
+      {
+        fs: { mkdirSync, unlinkSync },
+        askQuestion: askQuestionMock,
+        connectSocket: connectSocketMock,
+        console: { log },
+      },
+    );
+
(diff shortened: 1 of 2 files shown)


src/whatsapp-auth.ts
```diff
@@ -25,18 +24,78 @@ const AUTH_DIR = './store/auth';
...
+
+type ConnectSocketDeps = {
+  fs: Pick<typeof fs, 'writeFileSync' | 'unlinkSync'>;
+  makeWASocket: (config: Parameters<typeof makeWASocket>[0]) => SocketLike;
+  fetchLatestWaWebVersion: typeof fetchLatestWaWebVersion;
+  makeCacheableSignalKeyStore: typeof makeCacheableSignalKeyStore;
+  useMultiFileAuthState: typeof useMultiFileAuthState;
+  qrcodeGenerate: (qr: string, options: { small: boolean }) => void;
+  exit: (code: number) => void;
+  setTimeout: typeof setTimeout;
```
Peyton-Spencer, line 54: **🔴 Typecheck failure: `SocketLike.ev.on` signature is incompatible with real Baileys.**

The default `makeWASocket` returns a socket where `ev.on` is typed as:
```ts
<T extends keyof BaileysEventMap>(event: T, listener: (arg: BaileysEventMap[T]) => void) => void
```

But `SocketLike` declares:
```ts
ev: { on: (event: string, handler: (...args: any[]) => void) => void }
```

`string` is not assignable to `keyof BaileysEventMap`, so the assignment at line 65 (`defaultConnectSocketDeps.makeWASocket = makeWASocket`) fails typecheck.

Simplest fix: cast `makeWASocket` in the default deps — `makeWASocket as unknown as ConnectSocketDeps["makeWASocket"]` — since we only use `.requestPairingCode` and `.ev.on` from the return value.

src/whatsapp-auth.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    setTimeout: mock((handler: () => void | Promise<void>, _delay: number) => {
+      timers.push(handler);
+      return 1 as unknown as Timer;
+    }),
+    console: { log, error },
+    logger: { warn } as unknown,
+  };
+
+  return {
+    deps: { ...deps, ...overrides } as Parameters<typeof connectSocket>[3],
```
Peyton-Spencer: **🔴 Typecheck failure: `fetchLatestWaWebVersion` mock return type.**

The mock returns `{ version: [1, 2, 3] }` but the real function returns `{ version: WAVersion; isLatest: boolean; error?: unknown }`. TypeScript catches that `isLatest` is missing.

Fix: include the missing fields:
```ts
fetchLatestWaWebVersion: mock(async () => ({ version: [1, 2, 3] as [number, number, number], isLatest: true })),
```
Or cast: `mock(async () => ({ version: [1, 2, 3], isLatest: true }) as any)`

src/whatsapp-auth.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      throw new Error('missing');
+    });
+    const askQuestionMock = mock(async () => '14155551234');
+    const connectSocketMock = mock(async (_phoneNumber?: string) => {});
+    const log = mock(() => {});
+
+    await authenticate(
+      { usePairingCode: true },
+      {
+        fs: { mkdirSync, unlinkSync },
```
Peyton-Spencer, line 141: **🔴 Typecheck failure: `mkdirSync` mock return type mismatch.**

`mock(() => {})` returns `void`, but `fs.mkdirSync` with `{ recursive: true }` returns `string | undefined`. TypeScript flags this.

Fix:
```ts
const mkdirSync = mock((): string | undefined => undefined);
```

src/whatsapp-auth.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    });
+
+    await connectSocket(undefined, false, cliOptions, harness.deps);
+
+    expect(harness.writeFileSync).toHaveBeenCalledWith(
+      './store/auth-status.txt',
+      'already_authenticated',
+    );
+    expect(harness.exit).toHaveBeenCalledWith(0);
+    expect(harness.deps.makeWASocket).not.toHaveBeenCalled();
```
Peyton-Spencer, line 177: **🔴 Typecheck: `harness.deps` is possibly undefined (TS18048).**

The `createConnectDeps` return type infers `deps` from the spread `{ ...deps, ...overrides }` — when `overrides` has type `Record<string, unknown>`, the result gains an index signature that makes all properties potentially `undefined`.

Three occurrences (lines 160, 187, 240). Fix options:
1. Type `overrides` parameter as `Partial<ConnectSocketDeps>` instead of `Record<string, unknown>`
2. Add a non-null assertion on the return: `deps: { ...deps, ...overrides } as ConnectSocketDeps`
3. Use `satisfies` if you want to keep the assertion local

src/whatsapp-auth.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      'failed:logged_out',
+    );
+    expect(harness.exit).toHaveBeenCalledWith(1);
+  });
+
+  it('marks timed-out disconnects as qr timeouts', async () => {
+    const harness = createConnectDeps();
+
+    await connectSocket(undefined, false, cliOptions, harness.deps);
+    harness.socketHarness.handlers['connection.update']({
```
Peyton-Spencer, line 238: **🟡 Fragile microtask flushing.**

The reconnect-on-515 test uses two sequential `await Promise.resolve()` to flush the microtask queue before asserting. This is timing-dependent — if the internal reconnect chain gains an additional async step, this test silently passes without actually verifying the reconnect.

Consider `Bun.sleep(0)` which schedules after all pending microtasks, or make `connectSocket` return/signal when the reconnect fires so the test can await it directly.

src/whatsapp-auth.test.ts
```diff
@@ -0,0 +1,272 @@
+import { beforeEach, describe, expect, it, mock } from 'bun:test';
```
Peyton-Spencer, line 1: **🔴 Prettier formatting failure.**

CI `format:check` reports this file has style issues. Run `bunx prettier --write src/whatsapp-auth.test.ts` and amend.

src/whatsapp-auth.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      'failed:qr_timeout',
+    );
+    expect(harness.exit).toHaveBeenCalledWith(1);
+  });
+
+  it('reconnects after stream error 515 instead of failing the flow', async () => {
+    const harness = createConnectDeps();
+
+    await connectSocket('14155551234', false, cliOptions, harness.deps);
+    harness.socketHarness.handlers['connection.update']({
```
Peyton-Spencer, line 256: **🟢 Nice: success path tests both the status file write _and_ the deferred exit via the timer array.**

The pattern of capturing timers into an array and invoking them manually is clean — avoids needing fake timers and keeps the test deterministic. The QR file cleanup assertion via `unlinkSync` is a good detail to verify.

src/whatsapp-auth.ts
```diff
@@ -25,18 +24,78 @@ const AUTH_DIR = './store/auth';
...
+  prompt: string,
+  createInterface: typeof readline.createInterface = readline.createInterface,
+): Promise<string> {
+  const rl = createInterface({
     input: process.stdin,
     output: process.stdout,
   });
+
   return new Promise((resolve) => {
     rl.question(prompt, (answer) => {
```
Peyton-Spencer, line 103: **🟢 Good: `import.meta.main` guard.**

Replacing the top-level `authenticate().catch(...)` with an `import.meta.main` guard is the correct Bun idiom. This is what makes the module importable without side effects — critical for the test file to work at all.

Peyton-Spencer (commented):
Good test design — the dependency injection refactor makes the auth flow fully testable without Baileys sockets or process exits. The `import.meta.main` guard is the right Bun pattern. Branch coverage is thorough (7 `connectSocket` paths + CLI parsing + readline + authenticate flow).

Two CI blockers need fixing before merge: typecheck errors (7 diagnostics) and a Prettier formatting issue.

Peyton-Spencer:
Triage: good coverage addition with a sensible testability refactor in src/whatsapp-auth.ts, and CI is green. This looks ready for normal review/merge.

Peyton-Spencer:
Local validation does not match the green CI signal here. Running bun test and bunx tsc --noEmit on this branch fails with TypeScript errors in both src/whatsapp-auth.ts and src/whatsapp-auth.test.ts, including the ConnectSocketDeps mock typing, the makeWASocket return type, and several possibly-undefined harness.deps accesses. I am not comfortable merging until the branch typechecks locally.

PR #518, opened 2026-04-09
Fix ARM64 container browser bootstrap

Peyton-Spencer:
## Summary
- fix the ARM64 container build by skipping Chrome-for-Testing downloads and installing system Chromium instead
- copy the native agent-browser binary out of /root/.bun and wrap it from /usr/local/bin so the runtime no longer depends on root-owned paths
- keep the existing amd64 install path intact while preserving Playwright dependency setup

## Verification
- just build-all
- container run --rm --entrypoint /bin/sh omniclaw-agent-base:latest -c 'uname -m && ls -l /usr/local/bin/agent-browser /usr/local/bin/agent-browser-real && agent-browser --help >/tmp/agent-browser-help.txt && sed -n "1,20p" /tmp/agent-browser-help.txt'

## Summary by CodeRabbit

* **Chores**
  * Optimized browser agent installation and initialization process with improved architecture-specific handling.
  * Enhanced ARM64 system support to utilize native system browser when available for better performance.
  * Refined cross-platform browser compatibility and setup reliability.

### DIFF
--- a/container/Dockerfile.base
+++ b/container/Dockerfile.base
@@ -83,16 +83,31 @@ RUN bun install -g @anthropic-ai/claude-code
 # `agent-browser install --with-deps` calls sudo internally which doesn't exist, so we
 # run `npx playwright install-deps` directly as root instead (no sudo needed).
 # The JS wrapper at /usr/local/bin/agent-browser uses Node's existsSync() which returns
-# false for paths under /root/.bun/ due to EACCES on access() in the container sandbox,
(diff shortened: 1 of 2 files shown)


Peyton-Spencer:
Triage: CI is green and the scope is tight to the container bootstrap path. This looks mergeable from a queue-management standpoint. One thing to watch: if #507 lands first, rebase this so the agent-browser version does not step back during merge.

PR #525, opened 2026-04-10
fix(security): cap trusted peer error bodies

### LINKED ISSUE #523: security: cap trusted-peer discovery error bodies
## Summary

`PeerClient.fetch()` reads non-OK peer responses with `await res.text()` before throwing. Because trusted peers are remote instances and the response body is fully buffered with no byte cap, a compromised or malicious trusted peer can force the local OmniClaw process to allocate an arbitrarily large string on error paths.

## Severity

P2 / Medium

## Affected code

- `src/discovery/peer-client.ts:216`
- `src/discovery/routes.ts:861`
- `src/discovery/routes.ts:941`
- `src/discovery/routes.ts:1006`
- `src/discovery/routes.ts:1048`
- `src/discovery/routes.ts:1076`
- `src/discovery/routes.ts:1194`

## Root cause

`PeerClient.fetch()` handles unexpected status codes like this:

```ts
if (!res.ok && !allowedStatuses.includes(res.status)) {
  const body = await res.text().catch(() => '');
  throw new Error(
    `Peer API error: ${res.status} ${res.statusText} - ${body}`,
  );
}
```

Unlike the recent avatar/icon proxy hardening, this path has no `Content-Length` gate and no streamed byte limit. Any discovery proxy call that hits a non-OK response from a trusted peer will buffer the full body before returning a 502.

## Impact

A trusted peer can return a huge error body for proxied discovery endpoints such as:

- `GET /api/discovery/peers/:instanceId/agents`
- `GET /api/discovery/peers/:instanceId/stats`
- `GET /api/discovery/peers/:instanceId/context/layers`
- `PUT /api/discovery/peers/:instanceId/context/file`
- `GET /api/discovery/peers/:instanceId/context/compare`
- `POST /api/discovery/peers/:instanceId/context/pull`

That forces the local instance to allocate and copy the remote error body into memory, which can cause memory pressure or process instability. Because discovery trust extends across machines, this is a remote DoS within the trusted-peer boundary.

## Reproduction

1. Pair two OmniClaw instances so one appears as a trusted peer.
2. Make the remote peer return a very large error response body for a discovery route, for example `/api/agents` or `/api/stats`.
3. Request the corresponding local proxy route, such as `/api/discovery/peers/:instanceId/agents`.
4. Observe that `PeerClient.fetch()` calls `res.text()` and buffers the entire remote body before throwing.
5. With a sufficiently large body, the local process experiences high memory use and may become unstable.

## Recommended fix

- Add a bounded error-body reader in `PeerClient.fetch()`.
- Reject oversized `Content-Length` values before reading.
- Stream error bodies with a small cap (for example 8-64 KiB) and truncate the reported error text.
- Reuse the same helper for any future peer-response buffering paths.

Peyton-Spencer:
## Summary

- cap trusted-peer error response bodies in `PeerClient.fetch()` so proxy failures cannot buffer unbounded remote text
- reject oversized declared error bodies immediately and truncate streamed error text after an 8 KiB cap
- add focused coverage for both the `Content-Length` short-circuit and streamed truncation paths

## Testing

- `bun test src/discovery/peer-client.test.ts src/discovery/routes.test.ts`
- `bun run typecheck`
- `bun run format:check`

Closes #523

### DIFF
--- a/src/discovery/peer-client.test.ts
+++ b/src/discovery/peer-client.test.ts
@@ -11,6 +11,7 @@ function sha256Hex(value: string): string {
 }
 
 afterEach(() => {
+  mock.restore();
   globalThis.fetch = originalFetch;
 });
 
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Clean implementation. The dual-layer defense (Content-Length early reject + streamed byte cap) matches the pattern from PR #504. Good resource cleanup with body cancel and reader.releaseLock in finally. Tests cover both paths well. LGTM from mergemaster — will merge once CI is green.

PR #530, opened 2026-04-11
fix(discovery): cap trusted-peer JSON response bodies

Peyton-Spencer:
## Summary
- cap trusted-peer JSON responses in `PeerClient` before parsing so paired remote instances cannot force unbounded `res.json()` buffering
- reject oversized `Content-Length` values up front and enforce the same 1 MiB ceiling while streaming response bodies
- add regression tests for both header-based and streamed overflow on authenticated discovery responses

## Testing
- `bun test src/discovery/peer-client.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced a 1 MiB limit for JSON responses from peers to improve stability and prevent oversized downloads.
  * Peer responses that exceed the limit are now rejected and their downloads are terminated.

* **Tests**
  * Added tests verifying enforcement of the JSON response size limit for both streamed and non-streamed responses.

### DIFF
--- a/src/discovery/peer-client.ts
+++ b/src/discovery/peer-client.ts
@@ -18,6 +18,7 @@ const DEFAULT_TIMEOUT_MS = 10_000;
 const MAX_PEER_PROXY_IMAGE_BYTES = 5 * 1024 * 1024;
 const MAX_PEER_ERROR_BODY_BYTES = 8 * 1024;
 const MAX_PEER_JSON_RESPONSE_BYTES = 1024 * 1024;
+const PEER_JSON_RESPONSE_LABEL = 'Peer JSON response';
 const textDecoder = new TextDecoder();
 
 export interface PeerClientLike {
@@ -61,7 +62,6 @@ export class PeerClient implements PeerClientLike {
     return readJsonResponseWithLimit<PeerInfoResponse>(
       res,
       MAX_PEER_JSON_RESPONSE_BYTES,
-      'Peer JSON response',
     );
   }
 
@@ -75,7 +75,6 @@ export class PeerClient implements PeerClientLike {
     return readJsonResponseWithLimit<PairResponse>(
       res,
       MAX_PEER_JSON_RESPONSE_BYTES,
-      'Peer JSON response',
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/discovery/peer-client.ts
```diff
@@ -313,6 +343,24 @@ async function readErrorResponseWithLimit(
...
+    response.headers.get('content-length') || '',
+    10,
+  );
+  if (Number.isFinite(contentLength) && contentLength > maxBytes) {
+    response.body?.cancel().catch(() => {});
+    throw new Error(`${label} exceeded ${maxBytes} bytes`);
+  }
+
+  const bytes = await readStreamWithByteLimit(response.body, maxBytes);
+  return JSON.parse(textDecoder.decode(bytes)) as T;
```
Peyton-Spencer: Nit: the content-length fast-reject throws `"Peer JSON response exceeded 1048576 bytes"` (from the `label` parameter), but the streaming path in `readStreamWithByteLimit` throws `"Download exceeded 1048576 bytes"`. The tests correctly verify both messages, so this works, but it means callers/logs see different error strings depending on whether the server set `Content-Length` or not.

Not a blocker — just noting for future grep-ability. Could unify by catching the `readStreamWithByteLimit` error and rethrowing with the `label`, but that's probably overkill for a security fix.

src/discovery/peer-client.ts
```diff
@@ -313,6 +343,24 @@ async function readErrorResponseWithLimit(
...
+  response: Response,
+  maxBytes: number,
+  label: string,
+): Promise<T> {
+  const contentLength = Number.parseInt(
+    response.headers.get('content-length') || '',
+    10,
+  );
+  if (Number.isFinite(contentLength) && contentLength > maxBytes) {
+    response.body?.cancel().catch(() => {});
```
Peyton-Spencer, line 344: Ultra-minor: every call site passes `'Peer JSON response'` as the label. Could default the parameter (`label = 'Peer JSON response'`) to save the repetition at each call site. Not worth a follow-up on its own.

Peyton-Spencer (commented):
LGTM — clean, well-scoped security hardening. All JSON-parsing endpoints are covered (completePairing() returns void, avatar/icon use readBinaryResponseWithLimit(), streamLogs() returns raw Response). Two-layer defense (Content-Length fast-reject + streaming byte limit) is solid. Tests cover both rejection paths. Two minor nits inline, neither blocking. Ship it.

PR #533, opened 2026-04-12
feat(web): add password-based session authentication

Peyton-Spencer:
## Summary

Adds session-based password authentication to the Web UI, replacing the browser-native Basic Auth popup with a proper login page and cookie-based sessions.

- New `WEB_PASSWORD` env var enables session auth with a styled login page at `/login`
- HttpOnly session cookies (24-hour TTL, timing-safe password comparison, LRU eviction at 100 sessions)
- `/logout` endpoint clears the session and redirects to login
- API routes return 401 JSON; page routes redirect to `/login` when unauthenticated
- Session auth takes precedence over Basic Auth when both are configured
- Auth is skipped entirely when neither `WEB_PASSWORD` nor `WEB_UI_USER`/`WEB_UI_PASS` is set (dev/localhost mode)

Closes #532

## Test plan

- [x] 31 new tests (unit + integration) covering session store, cookie parsing, password verification, login page rendering, full login/logout flow
- [x] All 466 web tests pass (0 regressions)
- [x] Full suite: 1773 tests pass
- [x] TypeScript compiles cleanly (`tsc --noEmit`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added session-based password authentication for the web UI as an alternative to Basic Auth.
  * Added login page and logout functionality.
  * Introduced `WEB_PASSWORD` configuration option for simpler credential management.

* **Tests**
  * Added comprehensive test coverage for session authentication behavior.

PR #541, opened 2026-04-15
feat(security): add rate limiting to Web UI login endpoint

### LINKED ISSUE #540: security: add rate limiting to Web UI login endpoint
## Problem

PR #533 adds session-based password auth to the Web UI with a login page at `/login`. The current implementation uses timing-safe comparison and LRU session eviction, but has no rate limiting on login attempts.

Once the Web UI is exposed beyond localhost (the entire point of adding auth), the `/login` endpoint becomes vulnerable to brute-force password guessing. A simple password with no rate limiting is an invitation for automated attacks.

## Proposed approach

Add IP-based rate limiting to the `/login` POST endpoint:

1. **In-memory rate limiter** — track failed attempts per IP with a sliding window (e.g., 5 failures per minute)
2. **Exponential backoff response** — after the threshold, return 429 with a `Retry-After` header
3. **Auto-reset** — clear the counter after a successful login from that IP
4. **No external deps** — a simple `Map<string, { count, windowStart }>` is sufficient for a single-operator system

## Acceptance criteria

- [ ] Failed login attempts are tracked per source IP
- [ ] After N failures within a time window, further attempts return HTTP 429
- [ ] Successful login resets the counter for that IP
- [ ] Rate limit state does not persist across restarts (acceptable for single-operator use)
- [ ] Tests cover the rate limiting behavior

Peyton-Spencer:
## Summary

Adds IP-based rate limiting to the `/login` endpoint to prevent brute-force password guessing attacks. This is a security hardening follow-up to the session auth feature in PR #533.

- Sliding-window rate limiter tracks failed login attempts per source IP (5 failures / 60 seconds)
- After threshold, both `POST /login` and `GET /login` return HTTP 429 with `Retry-After` header
- Successful login resets the failure counter for that IP
- Background cleanup timer purges stale entries every 5 minutes (`.unref()` so it won't block process exit)
- Rate limiter disposed cleanly on server stop
- No external dependencies — simple `Map<string, { count, windowStart }>` implementation

Closes #540

**Depends on:** #533 (session auth)

## Test plan

- [x] 9 unit tests covering threshold, independent IP tracking, reset, expiry, cleanup, defaults
- [x] 5 integration tests covering 429 on excess failures, blocking correct password when rate-limited, counter reset after successful login, GET /login rate-limit message, no-op when session auth is disabled
- [x] All 1787 tests pass (0 regressions)
- [x] TypeScript compiles cleanly (`tsc --noEmit`)

### DIFF
--- a/src/web/rate-limit.test.ts
+++ b/src/web/rate-limit.test.ts
@@ -1,0 +1,276 @@
+import { describe, it, expect, afterEach } from 'bun:test';
+
+import { createRateLimiter } from './rate-limit.js';
+import { startWebServer, type WebServerHandle } from './server.js';
+import type { WebStateProvider, QueueStats } from './types.js';
+import type { Agent } from '../types.js';
+
+// ---- Fixtures ----
+
+function makeAgent(overrides: Partial<Agent> = {}): Agent {
+  return {
+    id: 'test-agent',
+    name: 'Test Agent',
+    folder: 'test-agent',
+    backend: 'apple-container',
+    agentRuntime: 'claude-agent-sdk',
+    isAdmin: false,
+    createdAt: '2026-01-01T00:00:00.000Z',
+    ...overrides,
+  };
+}
+
+const defaultStats: QueueStats = {
+  activeContainers: 0,
+  idleContainers: 0,
+  maxActive: 8,
+  maxIdle: 4,
+};
+
+function makeState(
+  overrides: Partial<WebStateProvider> = {},
+): WebStateProvider {
+  return {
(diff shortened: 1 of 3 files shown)


Peyton-Spencer (commented):
I’m not merging this yet.

This PR is currently based on `feat/web-session-auth` rather than `main`, and it only shows the CodeRabbit status right now. I want to see full CI on the final main-based diff before merging a login-path security change.

Once `#533` is merged or this branch is rebased onto `main` and the normal `format` / `test` / `typecheck` jobs are green, I can review it for merge.

Issue #505, opened 2026-04-06
tracking: Web UI Phase 3 — auth, live features, and polish


Peyton-Spencer:
## Context

Web UI Phase 1 (#157) and Phase 2 (#343) are both closed. Active web work continues without a tracking issue, making it harder to see progress and prioritize remaining features.

Recent/active Web UI PRs:
- #494: live execution status badges + task pause/resume on agent detail (open)
- #487: markdown rendering + live SSE messages in conversations (open, format CI failing)
- #469: remote logs and context editing error handling (merged)
- #468: versioned DB migration framework (merged)
- #463: SPA navigation regression tests (merged)

### Ship-ready (PRs open)
- [ ] Live agent execution status badges (#494)
- [ ] Conversation markdown rendering + live SSE messages (#487)

### Missing functionality
- [ ] Authentication / access control — the Web UI currently has no auth layer
- [ ] Task creation from the UI (currently only pause/resume/view)
- [ ] Conversation search and filtering
- [ ] Dark mode / theme toggle
- [ ] Mobile-responsive layout improvements

### Polish
- [ ] Loading states and error boundaries across all pages
- [ ] Consistent toast/notification system for user feedback
- [ ] Keyboard shortcuts for common operations

## Why this matters

The Web UI is the #1 strategic priority in the roadmap. Without a tracking issue, feature work drifts and it's hard to assess what "done" looks like for the next milestone.

Peyton-Spencer:
## Progress update — Apr 10-11 mergemaster sessions

Several Phase 3 items landed:

**Shipped:**
- [x] Live agent execution status badges (#494) — merged
- [x] Conversation markdown rendering + live SSE messages (#487) — merged
- [x] Loading states and error boundaries across all pages (#519) — merged
- [x] Conversation export (JSON and text) (#526) — merged

**Remaining from checklist:**
- [ ] Authentication / access control
- [ ] Task creation from the UI
- [ ] Conversation search and filtering
- [ ] Dark mode / theme toggle
- [ ] Mobile-responsive layout improvements
- [ ] Consistent toast/notification system
- [ ] Keyboard shortcuts

Peyton-Spencer:
## Phase 3 Progress Update (April 11)

Several items from the Phase 3 scope have been shipped since this issue was filed:

**Ship-ready (now merged)**
- [x] Live agent execution status badges (#494) — merged
- [x] Conversation markdown rendering + live SSE messages (#487) — merged
[...]

PR #544, opened 2026-04-15
feat(security): add rate limiting to Web UI login endpoint

### LINKED ISSUE #540: security: add rate limiting to Web UI login endpoint

Peyton-Spencer:
## Summary

- Adds fixed-window IP rate limiting for failed `/login` attempts when session auth is enabled.
- Returns `429` with `Retry-After` after repeated failures and resets the limiter after a successful login.
- Keeps the limiter disabled when session auth is off; `/login` is exercised directly in that regression test.

## Review Notes

- Rebased onto current `main` after PR #614.
- Addressed CodeRabbit's limiter naming concern by documenting the implementation as fixed-window instead of claiming sliding-window behavior.
- Addressed the disabled-auth test concern by repeatedly posting to `/login` and asserting it does not return `429`.

Closes #540

### DIFF
--- a/src/web/rate-limit.test.ts
+++ b/src/web/rate-limit.test.ts
@@ -1,0 +1,280 @@
+import { describe, it, expect, afterEach } from 'bun:test';
+
+import { createRateLimiter } from './rate-limit.js';
+import { startWebServer, type WebServerHandle } from './server.js';
+import type { WebStateProvider, QueueStats } from './types.js';
+import type { Agent } from '../types.js';
+
+// ---- Fixtures ----
+
+function makeAgent(overrides: Partial<Agent> = {}): Agent {
+  return {
+    id: 'test-agent',
+    name: 'Test Agent',
+    folder: 'test-agent',
+    backend: 'apple-container',
+    agentRuntime: 'claude-agent-sdk',
+    isAdmin: false,
+    createdAt: '2026-01-01T00:00:00.000Z',
+    ...overrides,
+  };
+}
+
+const defaultStats: QueueStats = {
+  activeContainers: 0,
+  idleContainers: 0,
+  maxActive: 8,
+  maxIdle: 4,
+};
+
+function makeState(
+  overrides: Partial<WebStateProvider> = {},
+): WebStateProvider {
+  return {
+    getAgents: () => ({ 'test-agent': makeAgent() }),
+    getChannelSubscriptions: () => ({}),
+    getTasks: () => [],
+    getTaskById: () => undefined,
+    getMessages: () => [],
+    getChats: () => [],
+    getQueueStats: () => defaultStats,
+    getQueueDetails: () => [],
+    getIpcEvents: () => [],
+    getTaskRunLogs: () => [],
+    getTaskRunPhaseEvents: () => [],
+    searchMessages: () => [],
+    createTask: () => {},
+    updateTask: () => {},
+    deleteTask: () => {},
+    calculateNextRun: () => null,
+    readContextFile: () => null,
+    writeContextFile: () => {},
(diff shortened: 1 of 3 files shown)


Peyton-Spencer:
**Product note:** CodeRabbit flagged a real security concern — the `'unknown'` IP fallback creates a shared rate-limit bucket that could lock out all users when `requestIP()` is unavailable. The sliding-window suggestion also makes sense for a login endpoint. These should be addressed before merge.

This PR addresses #540 and is part of Web UI Phase 3 (#505).

Peyton-Spencer:
Triage review: implementation looks appropriately scoped for `#540`, but I would hold merge until standard CI is attached on the PR.

- The code/test changes look coherent: dedicated rate limiter module, server integration, and focused tests.
- Current PR checks only show `CodeRabbit`; I am not seeing the normal `format` / `typecheck` / `test` suite on the PR.
- Given this touches auth-path request handling, I want the standard CI suite visible before calling it merge-ready.

If CI is just missing due to a GitHub-side glitch, rerun or refresh checks and this likely moves to ready quickly.

Peyton-Spencer (commented):
Blocking issue:

This adds login throttling, but the login handler still uses unbounded `req.formData()` before parsing the password. That leaves the endpoint exposed to the same pre-auth memory/CPU denial-of-service problem from oversized request bodies, so the brute-force protection would merge in with a body-parsing gap still open.

I’m not comfortable merging this as-is. It needs the bounded request-body reader on `/login` as well, or a rebase onto the body-cap fix, before it meets the security bar.

Peyton-Spencer:
Triage review:

The rate-limit behavior is useful, but this branch overlaps directly with `#549` in the `/login` path and should be rebased after the body-cap patch lands so both protections are exercised together.

I also only see CodeRabbit on the branch right now, so there is no current format/typecheck/test signal attached. Please rebase on latest `main` (or on top of `#549`), rerun CI, and then this is easier to evaluate for merge.

Peyton-Spencer:
Triage pass: this still looks directionally valid, but it is not currently merge-ready. The branch is DIRTY, the last non-CodeRabbit CI signal is from Apr 15, and base main has moved since then. Next step is to rebase or merge main into this branch and rerun CI so the implementation and test signal reflect the current base before review.

Peyton-Spencer:
Triage note: this still tracks #540 and remains relevant, but GitHub reports the branch as conflicting with `main` and only CodeRabbit is present in the status rollup. Next step is to rebase/merge `main`, resolve conflicts, and rerun the standard CI checks before review/merge.

PR #547, opened 2026-04-15
test(migrations): cover helper edge cases

Peyton-Spencer:
## Summary

- add direct unit coverage for `src/migrations.ts` helper behavior, including duplicate-column tolerance, defaulted column adds, and column-drop no-ops
- add a monkey-patched logger test to verify failed legacy column drops warn instead of aborting migrations
- add a deterministic baseline-migration regression test for clearing legacy heartbeat state and removing `stream_intermediates`

## Validation

- `bun test src/migrations.test.ts`
- `bunx tsc --noEmit`
- `bun test`

## Test Count

- before: 1781 tests
- after: 1787 tests
- added: 6 tests in `src/migrations.test.ts`

## Files Covered

- `src/migrations.ts`

## Remaining Gaps

- `src/index.ts` still has no direct unit coverage and remains the largest uncovered orchestration path
- `src/ipc.ts` is exercised indirectly by IPC tests, but helper-level branches still lack a dedicated unit test file
- `src/discovery/types.ts` and `src/web/types.ts` are type-only modules with no runtime behavior to cover

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for database migration operations, including column management and schema versioning with error handling validation.

### DIFF
--- a/src/migrations.test.ts
+++ b/src/migrations.test.ts
@@ -1,0 +1,152 @@
+import { afterEach, describe, expect, it, mock, spyOn } from 'bun:test';
+import { Database } from 'bun:sqlite';
+
+import { logger } from './logger.js';
+import {
+  addColumnIfNotExists,
+  allMigrations,
+  dropColumnIfExists,
+  getSchemaVersion,
+  runMigrations,
+} from './migrations.js';
+
+function getColumns(db: Database, table: string): string[] {
+  return (
+    db.query(`PRAGMA table_info(${table})`).all() as Array<{ name: string }>
+  ).map((row) => row.name);
+}
+
+afterEach(() => {
+  mock.restore();
+});
+
+describe('migration helpers', () => {
+  it('adds missing columns with defaults for existing rows', () => {
+    const db = new Database(':memory:');
+    db.exec(
+      'CREATE TABLE example (id INTEGER PRIMARY KEY, name TEXT NOT NULL)',
+    );
+    db.exec("INSERT INTO example (name) VALUES ('first')");
+
+    addColumnIfNotExists(db, 'example', 'enabled', 'INTEGER NOT NULL', '1');
+
+    const row = db
+      .query('SELECT enabled FROM example WHERE name = ?')
+      .get('first') as { enabled: number };
+
+    expect(getColumns(db, 'example')).toContain('enabled');
+    expect(row.enabled).toBe(1);
+
+    db.close();
+  });
+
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
Good test coverage for the migration helpers. Clean structure, sensible edge-case selection (duplicate columns, missing tables, column-drop no-ops, logger warn path). A few observations:

1. **Missing version assertion in baseline test** — the baseline migration test verifies heartbeat clearing and `stream_intermediates` removal but doesn't assert `getSchemaVersion(db) === 1` afterward. Adding that would catch regressions in the version-tracking machinery (which is arguably the most critical part of the migration framework).

2. **Idempotency test would add confidence** — `runMigrations` should be safe to call twice. A one-liner `expect(() => runMigrations(db, allMigrations)).not.toThrow()` after the existing baseline assertions would verify the "already at current version" short-circuit path.

3. Minor: `afterEach(mock.restore)` at module scope is fine, but since only the logger-warn test uses `spyOn`, scoping it inside that `describe` block would make the intent clearer.

Overall LGTM — these are suggestions, not blockers.

src/migrations.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      .get('dc:test') as { heartbeat: string | null };
+    const agentRow = db
+      .query('SELECT heartbeat FROM agents WHERE id = ?')
+      .get('agent-1') as { heartbeat: string | null };
+
+    expect(groupRow.heartbeat).toBeNull();
+    expect(agentRow.heartbeat).toBeNull();
+    expect(getColumns(db, 'registered_groups')).not.toContain(
+      'stream_intermediates',
+    );
```
Peyton-Spencer, line 153: Consider asserting the schema version was bumped after the migration:

```ts
expect(getSchemaVersion(db)).toBe(1);
```

This catches regressions in `setSchemaVersion` / `ensureVersionTable` without a separate test, and validates the full `runMigrations` lifecycle: version check → apply → stamp.

Optional: a follow-up `runMigrations(db, allMigrations)` call here (idempotency) would exercise the "no pending migrations" early-return path — currently the only untested branch in `runMigrations`.

Peyton-Spencer:
Triage review: merge-ready from queue hygiene perspective.

- Scope is tight: one new test file against `src/migrations.ts` helper behavior.
- CI is fully green (`format`, `typecheck`, `test`).
- Nothing in the diff changes runtime code paths, so regression risk is low.

I don't see a queue-level blocker here.

PR #549, opened 2026-04-16
fix(security): cap Web UI login request bodies

### LINKED ISSUE #548: security: cap Web UI login request bodies
## Summary

The session-auth login handler accepts unbounded form bodies. In `src/web/server.ts`, `POST /login` calls `req.formData()` directly, while the other authenticated write endpoints already use `readJsonBody(..., MAX_WEB_JSON_BODY_BYTES)` and reject oversized payloads with `413 Request body too large`.

This leaves `/login` as an easy memory/CPU denial-of-service target on deployments that expose the Web UI and enable `WEB_PASSWORD`.

## Severity

P2

## Affected code

- `src/web/server.ts:240-256`
- `src/request-body.ts:16-70`
- `src/web/routes.ts:439-454`
- `src/web/routes.ts:552-571`
- `src/web/routes.ts:873-888`
- `src/web/routes.ts:1030-1049`
- `src/web/routes.ts:1114-1138`

## Why this is a vulnerability

- `req.formData()` parses the full request body before password verification.
- `/login` is intentionally reachable before authentication, so any network client that can reach the Web UI can hit it.
- The codebase already treats oversized request bodies as a security concern elsewhere: webhook intake, trusted-peer auth, discovery routes, and authenticated Web UI JSON mutations all enforce explicit byte caps.
- Leaving the auth endpoint uncapped gives an attacker a simpler pre-auth exhaustion path than the already-hardened routes.

## Reproduction

1. Start OmniClaw with the Web UI exposed and session auth enabled via `WEB_PASSWORD`.
2. Send a very large request body to `/login`, for example a large `multipart/form-data` upload or a huge urlencoded `password=` value.
3. Observe that the server attempts to parse the full form payload instead of rejecting it early with `413`.
4. Repeat a few times concurrently to amplify memory pressure and CPU spent parsing bodies.

Example shape:

```bash
perl -e 'print "password=" . ("A" x (50*1024*1024))' | \
  curl -i \
    -X POST \
    -H 'Content-Type: application/x-www-form-urlencoded' \
    --data-binary @- \
    http://127.0.0.1:3000/login
```

## Expected behavior

The login route should apply the same bounded-body pattern as the other Web UI write routes and fail oversized bodies with `413 Request body too large` before form parsing.

## Suggested fix

- Add a dedicated max body size for login form submissions, likely matching the existing 1 MiB Web UI JSON cap.
- Parse the raw body through the shared bounded reader before constructing `URLSearchParams` or otherwise extracting the password field.
- Add regression tests covering both over-limit `Content-Length` and streaming bodies that exceed the limit mid-read.

Peyton-Spencer:
## Summary

Caps the `POST /login` request body to prevent pre-auth memory/CPU denial-of-service.

- Replaces unbounded `req.formData()` with the shared `readRequestBody()` bounded reader (1 MiB cap, matching other Web UI write endpoints)
- Parses the bounded body as `URLSearchParams` to extract the password field
- Returns HTTP 413 for oversized payloads before any form parsing
- Updates existing session-auth tests to use `URLSearchParams` (matches actual browser form submission encoding)
- 5 new regression tests: oversized body rejection, normal login, wrong password, missing password field, empty body

Closes #548

## Test plan

- [x] 5 new tests covering body cap enforcement and edge cases
- [x] All 36 session-auth tests pass (0 regressions)
- [x] Full suite: 1780 tests pass, 0 failures
- [x] TypeScript compiles cleanly (`tsc --noEmit`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added request body size validation for login submissions with HTTP 413 response for oversized payloads
  * Improved error handling for malformed login requests with appropriate HTTP status codes and messages

### DIFF
--- a/src/web/server.ts
+++ b/src/web/server.ts
@@ -4,6 +4,10 @@ import { ServerSentEventGenerator } from '@starfederation/datastar-sdk/web';
 
 import { logger } from '../logger.js';
 import {
+  readRequestBody,
+  RequestBodyTooLargeError,
+} from '../request-body.js';
+import {
   handleRequest,
   getRemotePeers,
   createRemotePeerResolver,
@@ -61,6 +65,7 @@ const PORT_ZERO_RETRY_ATTEMPTS = 10;
 const PORT_ZERO_FALLBACK_START = 40000;
 const PORT_ZERO_FALLBACK_SPAN = 20000;
 const MAX_PEER_AUTH_BODY_BYTES = 1024 * 1024;
+const MAX_LOGIN_BODY_BYTES = 1024 * 1024;
 
 interface RequestBodyHashResult {
   hash: string;
@@ -238,8 +243,23 @@ export function startWebServer(
           });
         }
         if (req.method === 'POST') {
-          const formData = await req.formData().catch(() => null);
-          const password = formData?.get('password');
+          let rawBody: string;
+          try {
+            rawBody = await readRequestBody(req, MAX_LOGIN_BODY_BYTES);
+          } catch (err) {
+            if (err instanceof RequestBodyTooLargeError) {
+              return new Response('Request body too large', {
+                status: 413,
+                headers: { 'Content-Type': 'text/plain' },
+              });
+            }
+            return new Response(renderLoginPage('Invalid request'), {
+              status: 400,
+              headers: { 'Content-Type': 'text/html; charset=utf-8' },
+            });
+          }
+          const params = new URLSearchParams(rawBody);
+          const password = params.get('password');
           if (
             typeof password === 'string' &&
             verifyPassword(password, sessionPassword)
--- a/src/web/server.ts
+++ b/src/web/server.ts
@@ -3,10 +3,7 @@ import { createHash } from 'crypto';
 import { ServerSentEventGenerator } from '@starfederation/datastar-sdk/web';
 
 import { logger } from '../logger.js';
-import {
-  readRequestBody,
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Local verification looks good on this one, but I’m not merging yet because the PR currently has no normal CI signal beyond CodeRabbit.

What I checked locally:
- `bun test src/web/session-auth.test.ts`
- `bunx tsc --noEmit`
- `bun test`

All of those passed for me. Once the standard CI jobs are present and green again, I’d be comfortable merging.

Peyton-Spencer:
Triage review:

The fix direction looks right. Switching `/login` off unbounded `req.formData()` and onto the shared bounded reader closes the pre-auth body-size gap with minimal surface area, and the regression tests cover the main branches.

I do not see an obvious code issue in the diff. The remaining concern is process state: the PR body claims full-suite green and clean typecheck, but the branch is only showing CodeRabbit at the moment and `main` is still red on unrelated checks. I would treat this as ready from a code-shape perspective, pending a real CI run on the branch.

Peyton-Spencer:
Triage: this looks superseded by #567, which targets the same login body-cap fix, includes the same core server/session-auth changes plus an extra request-body test, and already has green CI on the newer branch. Unless there is a reason to keep both open, this older PR can likely be closed in favor of #567 to keep the queue clean.

Peyton-Spencer:
Superseded by PR #567 which was merged and closed #548.

PR #566, opened 2026-04-18
test(discovery): cover network identity fallbacks

Peyton-Spencer:
## Summary
- add deterministic `network-identity` tests for the macOS `networksetup` PATH fallback plus Linux `iwgetid`/`nmcli` branches
- cover the unsupported-platform fast path so `detectCurrentNetwork()` no longer relies almost entirely on the existing macOS happy path tests

## Testing
- `bun test src/discovery/network-identity.test.ts`
- `bun test`
- `bun test --coverage`

## Test Count
- before: 1781 tests, 95 files, 4626 assertions
- after: 1785 tests, 95 files, 4632 assertions

## Coverage Notes
- improved `src/discovery/network-identity.ts` coverage to 95.00% functions / 91.28% lines
- current repo-wide coverage after this change: 88.08% functions / 85.86% lines

## Remaining Gaps
- `src/index.ts` (4.44% lines)
- `src/channels/discord.ts` (5.14% lines)
- `src/discovery/mdns.ts` (15.84% lines)
- `src/channels/slack.ts` (22.00% lines)
- `src/channels/telegram.ts` (22.78% lines)

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for network detection logic across platforms.
  * Validates macOS fallback behavior when the primary system utility fails and a secondary lookup provides the SSID.
  * Adds Linux cases covering primary/secondary command fallbacks and scenarios with no active SSID.
  * Confirms unsupported platforms return no network result and do not attempt external lookups.

### DIFF
--- a/src/discovery/network-identity.test.ts
+++ b/src/discovery/network-identity.test.ts
@@ -245,7 +245,9 @@ describe('network identity detection', () => {
       }
 
       if (cmd[0] === 'networksetup' && cmd.includes('-getairportnetwork')) {
-        return createProcess({ stdout: 'Current Wi-Fi Network: Fallback WiFi\n' });
+        return createProcess({
+          stdout: 'Current Wi-Fi Network: Fallback WiFi\n',
+        });
       }
 
       throw new Error(`Unexpected command: ${cmd.join(' ')}`);
@@ -253,9 +255,14 @@ describe('network identity detection', () => {
 
     const result = await detectCurrentNetwork();
 
-    expect(result).toEqual({ id: 'wifi:Fallback WiFi', label: 'Fallback WiFi' });
+    expect(result).toEqual({
+      id: 'wifi:Fallback WiFi',
+      label: 'Fallback WiFi',
+    });
     expect(
-      spawnSpy.mock.calls.some((call: unknown[]) => call[0]?.[0] === 'networksetup'),
+      spawnSpy.mock.calls.some(
+        (call: unknown[]) => call[0]?.[0] === 'networksetup',
+      ),
     ).toBe(true);
   });
 
--- a/src/discovery/network-identity.test.ts
+++ b/src/discovery/network-identity.test.ts
@@ -230,6 +230,94 @@ describe('network identity detection', () => {
       ),
     ).toBe(true);
   });
+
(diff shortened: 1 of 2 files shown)


src/discovery/network-identity.test.ts
```diff
@@ -230,6 +230,101 @@ describe('network identity detection', () => {
...
+        return createProcess({
+          stdout: 'Current Wi-Fi Network: Fallback WiFi\n',
+        });
+      }
+
+      throw new Error(`Unexpected command: ${cmd.join(' ')}`);
+    }) as typeof Bun.spawn);
+
+    const result = await detectCurrentNetwork();
+
```
Peyton-Spencer, line 258: Nit (non-blocking): The `spawnSpy.mock.calls.some(...)` assertion only checks that *at least one* call used bare `networksetup`. Since this test exercises the retry path for **both** `-listallhardwareports` and `-getairportnetwork`, you'd expect exactly 4 spawn calls (2 absolute-path throws + 2 bare retries). Adding a call-count assertion would make the retry behavior more explicit:

```ts
expect(spawnSpy).toHaveBeenCalledTimes(4);
```

Without it, if the retry logic silently regressed to only retrying one of the two calls, this test would still pass.

src/discovery/network-identity.test.ts
```diff
@@ -230,6 +230,101 @@ describe('network identity detection', () => {
...
+    expect(result).toEqual({ id: 'wifi:Office Linux', label: 'Office Linux' });
+  });
+
```
Peyton-Spencer, line 290: Optional gap (non-blocking): There's no test for the case where the PATH fallback *also* fails — i.e., both `/usr/sbin/networksetup` and bare `networksetup` throw. In that scenario `runCommand` should fall through to `ipconfig` → `system_profiler` → `wdutil`. That's a deeper integration path and not strictly necessary for this PR's scope, but worth noting as a future coverage opportunity.

Peyton-Spencer (commented):
Overall clean PR — adds meaningful deterministic coverage for the macOS PATH fallback, Linux `iwgetid`→`nmcli` cascade, and the unsupported-platform fast path. Tests are well-isolated and use the existing `createProcess` helper properly. Good work.

Two minor suggestions inline, neither blocking.

PR #567, opened 2026-04-18
fix(web): cap login request bodies

### LINKED ISSUE #548: security: cap Web UI login request bodies

Peyton-Spencer:
## Summary
- cap Web UI login request bodies before form parsing so `POST /login` cannot buffer arbitrary-size pre-auth form payloads
- reuse the shared bounded request-body path for form parsing and return `413` on oversized login requests
- add helper and auth regressions for both oversized `Content-Length` and streamed overflow cases

Closes #548.

## Testing
- `bun test src/request-body.test.ts src/web/session-auth.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Login requests with oversized form bodies now return HTTP 413 with a clear error message instead of failing silently.
  * Enhanced form data parsing with explicit size limit enforcement (1 MB maximum per login request).

* **Tests**
  * Added integration tests verifying proper rejection of oversized login form submissions.

### DIFF
--- a/src/request-body.ts
+++ b/src/request-body.ts
@@ -68,3 +68,19 @@ export async function readJsonBody<T>(
 ): Promise<T> {
   return JSON.parse(await readRequestBody(req, maxBodyBytes)) as T;
 }
+
+export async function readFormDataBody(
+  req: Request,
+  maxBodyBytes: number,
+): Promise<Awaited<ReturnType<Request['formData']>>> {
+  const body = await readRequestBody(req, maxBodyBytes);
+  const contentType =
+    req.headers.get('content-type') ??
+    'application/x-www-form-urlencoded;charset=UTF-8';
+
+  return new Request(req.url, {
+    method: req.method,
+    headers: { 'content-type': contentType },
+    body,
+  }).formData();
+}
--- a/src/request-body.ts
+++ b/src/request-body.ts
@@ -68,19 +68,3 @@ export async function readJsonBody<T>(
 ): Promise<T> {
   return JSON.parse(await readRequestBody(req, maxBodyBytes)) as T;
 }
-
-export async function readFormDataBody(
-  req: Request,
-  maxBodyBytes: number,
-): Promise<Awaited<ReturnType<Request['formData']>>> {
-  const body = await readRequestBody(req, maxBodyBytes);
-  const contentType =
-    req.headers.get('content-type') ??
-    'application/x-www-form-urlencoded;charset=UTF-8';
-
-  return new Request(req.url, {
-    method: req.method,
-    headers: { 'content-type': contentType },
-    body,
-  }).formData();
-}
--- a/src/web/server.ts
+++ b/src/web/server.ts
@@ -53,6 +53,7 @@ import {
   getSolidHandler,
   isMainProcessRoute,
 } from './solid-handler.js';
+import { readFormDataBody, RequestBodyTooLargeError } from '../request-body.js';
 
 const MAX_SSE_CLIENTS = 100;
 const MAX_LOG_LINES = 500;
@@ -61,6 +62,7 @@ const PORT_ZERO_RETRY_ATTEMPTS = 10;
 const PORT_ZERO_FALLBACK_START = 40000;
 const PORT_ZERO_FALLBACK_SPAN = 20000;
 const MAX_PEER_AUTH_BODY_BYTES = 1024 * 1024;
+const MAX_LOGIN_FORM_BODY_BYTES = 1024 * 1024;
 
 interface RequestBodyHashResult {
   hash: string;
@@ -238,7 +240,18 @@ export function startWebServer(
           });
         }
         if (req.method === 'POST') {
-          const formData = await req.formData().catch(() => null);
+          const formData = await readFormDataBody(
+            req,
+            MAX_LOGIN_FORM_BODY_BYTES,
+          ).catch((err: unknown) => {
+            if (err instanceof RequestBodyTooLargeError) {
+              return err;
+            }
+            return null;
+          });
+          if (formData instanceof RequestBodyTooLargeError) {
+            return new Response('Request body too large', { status: 413 });
+          }
           const password = formData?.get('password');
           if (
             typeof password === 'string' &&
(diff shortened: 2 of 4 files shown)


src/web/server.ts
```diff
@@ -61,6 +62,7 @@ const PORT_ZERO_RETRY_ATTEMPTS = 10;
 const PORT_ZERO_FALLBACK_START = 40000;
 const PORT_ZERO_FALLBACK_SPAN = 20000;
 const MAX_PEER_AUTH_BODY_BYTES = 1024 * 1024;
+const MAX_LOGIN_FORM_BODY_BYTES = 1024 * 1024;
```
Peyton-Spencer: Minor: The `.catch()` returning-error-as-value pattern is unconventional and makes the control flow harder to follow — `formData` can be `FormData`, `RequestBodyTooLargeError`, or `null`, requiring a type-narrowing check after.

A `try/catch` would be clearer:

```ts
let formData: FormData | null;
try {
  formData = await readFormDataBody(req, MAX_LOGIN_FORM_BODY_BYTES);
} catch (err) {
  if (err instanceof RequestBodyTooLargeError) {
    return new Response('Request body too large', { status: 413 });
  }
  formData = null;
}
```

This also matches how `readRequestBody` errors are handled elsewhere in this file (e.g. `hashRequestBodyWithLimit` on the peer-auth path).

src/request-body.ts
```diff
@@ -68,3 +68,19 @@ export async function readJsonBody<T>(
...
 }
+
+export async function readFormDataBody(
+  req: Request,
+  maxBodyBytes: number,
+): Promise<Awaited<ReturnType<Request['formData']>>> {
+  const body = await readRequestBody(req, maxBodyBytes);
+  const contentType =
+    req.headers.get('content-type') ??
+    'application/x-www-form-urlencoded;charset=UTF-8';
```
Peyton-Spencer: Nit: `Promise<Awaited<ReturnType<Request['formData']>>>` is just `Promise<FormData>`. The type gymnastics are harder to read for no benefit:

```ts
export async function readFormDataBody(
  req: Request,
  maxBodyBytes: number,
): Promise<FormData> {
```

src/request-body.ts
```diff
@@ -68,3 +68,19 @@ export async function readJsonBody<T>(
...
+
+  return new Request(req.url, {
+    method: req.method,
+    headers: { 'content-type': contentType },
```
Peyton-Spencer: Nit (design): Since the login form is always `application/x-www-form-urlencoded`, an alternative is to skip the new utility entirely and use `readRequestBody` + `new URLSearchParams(body)` inline at the call site — consistent with how `readJsonBody` is just `JSON.parse(await readRequestBody(...))`. That said, `readFormDataBody` is defensible as a reusable companion to `readJsonBody`, so this is a style preference not a blocker.

Peyton-Spencer (commented):
Security fix is correct — caps the unbounded `req.formData()` at `/login` with the shared bounded reader, addressing the pre-auth memory DoS. Test coverage is solid with both content-length fast-reject and streamed overflow paths tested at unit and integration levels.

3 suggestions inline (1 minor, 2 nits). Also note this overlaps with PR #549 which addresses the same issue (#548) using a simpler `readRequestBody` + `URLSearchParams` approach — worth deciding which to land and closing the other.

Peyton-Spencer:
Addressed the review direction in follow-up commit `067ef12`.

- Switched `/login` from the new `readFormDataBody()` helper to the simpler shared `readRequestBody()` + `URLSearchParams` path.
- Kept the oversized streamed-body regression coverage from this PR.
- Added an explicit `400 Bad request` path for non-size request read failures instead of collapsing them into invalid credentials.

This now matches the simpler shape from `#549` while keeping the stronger coverage already present here.

PR #570, opened 2026-04-19
test(discovery): cover runtime controller edge cases

Peyton-Spencer:
## Summary
- add deterministic coverage for `DiscoveryRuntimeController` persistence, trust-list transitions, detection failures, and polling lifecycle behavior
- raise targeted coverage for `src/discovery/runtime.ts` from the previously reported 16.48% lines to 98.80% lines and 93.75% functions in the focused coverage run
- increase runtime controller tests from 3 to 9 cases while keeping the full clean-worktree suite green

## Testing
- `bun test src/discovery/runtime.test.ts`
- `bun test --coverage src/discovery/runtime.test.ts`
- `bun test`

## Notes
- remaining low-coverage gaps from the focused report are still concentrated in `src/discovery/trust-store.ts`, `src/discovery/pairing-crypto.ts`, `src/db.ts`, `src/config.ts`, and `src/logger.ts`

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for runtime discovery functionality, including network detection, trust management, state transitions, and polling behavior.

### DIFF
--- a/src/discovery/runtime.test.ts
+++ b/src/discovery/runtime.test.ts
@@ -1,6 +1,7 @@
-import { beforeEach, describe, expect, it, mock } from 'bun:test';
+import { beforeEach, describe, expect, it, mock, spyOn } from 'bun:test';
 
 import { setRouterState, _initTestDatabase } from '../db.js';
+import { logger } from '../logger.js';
 import { DiscoveryRuntimeController } from './runtime.js';
 
 describe('DiscoveryRuntimeController', () => {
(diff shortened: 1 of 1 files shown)


src/discovery/runtime.test.ts
```diff
@@ -58,4 +59,177 @@ describe('DiscoveryRuntimeController', () => {
...
+    await controller.refresh();
+    controller.trustCurrentNetwork();
+
+    detector.mockImplementation(async () => {
+      throw new Error('wifi scan failed');
+    });
+
+    const snapshot = await controller.refresh();
+
+    expect(snapshot.currentNetwork).toBeNull();
```
Peyton-Spencer: **Suggestion (non-blocking):** The manual `.mock.calls.some()` scan works but gives opaque failure output — a failing test just says `expected true, received false` with no hint about what was actually logged.

`toHaveBeenCalledWith` is more direct and produces better diagnostics:

```ts
expect(warnSpy).toHaveBeenCalledWith(
  expect.objectContaining({ err: expect.any(Error) }),
  'Failed to detect current network',
);
```

Same result, but if it ever breaks you'll see the actual vs expected arguments in the diff.

src/discovery/runtime.test.ts
```diff
@@ -58,4 +59,177 @@ describe('DiscoveryRuntimeController', () => {
...
+    expect(snapshot.active).toBe(false);
```
Peyton-Spencer: **Nit (non-blocking):** If any assertion above this line throws, `mockRestore()` never runs and the spy leaks into subsequent tests. Same concern applies to the `setIntervalSpy`/`clearIntervalSpy` restores at the bottom of the file.

Safest pattern is a scoped cleanup:

```ts
const warnSpy = spyOn(logger, 'warn').mockImplementation(() => {});
try {
  // ... assertions ...
} finally {
  warnSpy.mockRestore();
}
```

Alternatively, if Bun's test runner supports `afterEach` cleanup hooks per-test, that would also work. Low risk for now since this is the last test that touches `logger`, but worth hardening if the file grows.

src/discovery/runtime.test.ts
```diff
@@ -58,4 +59,177 @@ describe('DiscoveryRuntimeController', () => {
...
+
+  it('throws when trusting a network before one has been detected', () => {
+    const controller = new DiscoveryRuntimeController({
+      initialEnabled: true,
+      detectCurrentNetwork: async () => null,
+    });
+
+    expect(() => controller.trustCurrentNetwork()).toThrow(
+      'No current Wi-Fi network detected',
+    );
```
Peyton-Spencer, line 113: **Observation (non-blocking):** This test packs two distinct behaviors into one case — deduplication on double-trust, and permissive-mode fallback when all trusts are removed. Both are worth covering, but the long name is a hint they could be separate cases.

Splitting would make failures more precise ("dedup broke" vs "permissive fallback broke") and the individual test names shorter. Not blocking — the combined test is correct as-is.

src/discovery/runtime.test.ts
```diff
@@ -58,4 +59,177 @@ describe('DiscoveryRuntimeController', () => {
...
+      pollIntervalMs: 4321,
+    });
+
+    controller.start();
+    controller.start();
+    controller.stop();
+    controller.stop();
+
+    expect(setIntervalSpy).toHaveBeenCalledTimes(1);
+    expect(setIntervalSpy).toHaveBeenCalledWith(expect.any(Function), 4321);
```
Peyton-Spencer: **Nit (non-blocking):** Same spy-restore-on-failure concern as the `warnSpy` above — if `expect(setIntervalSpy)` or `expect(clearIntervalSpy)` throws, the restores on lines 228–229 are skipped and the global `setInterval`/`clearInterval` stay patched.

Wrapping in `try/finally` would make this robust:

```ts
try {
  controller.start();
  controller.start();
  controller.stop();
  controller.stop();

  expect(setIntervalSpy).toHaveBeenCalledTimes(1);
  expect(setIntervalSpy).toHaveBeenCalledWith(expect.any(Function), 4321);
  expect(clearIntervalSpy).toHaveBeenCalledTimes(1);
} finally {
  setIntervalSpy.mockRestore();
  clearIntervalSpy.mockRestore();
}
```

Peyton-Spencer (commented):
Good coverage addition — takes `runtime.ts` from ~16% to ~99% lines. Tests are well-structured, logic traces correctly, and CI is green across all checks.

A few suggestions below, none blocking.

PR #571, opened 2026-04-19
fix(security): restrict discovery LAN admin trust

### LINKED ISSUE #557: security: discovery admin LAN trust can bypass auth for public clients
## Summary

`DISCOVERY_TRUST_LAN_ADMIN` currently trusts the server's own bind address when deciding whether to skip auth for discovery admin routes. That lets a public client inherit LAN trust whenever the Web UI is bound to a private IP, which breaks the intended "only private-network requests may bypass auth" boundary.

Severity: Medium

## Affected code

- `src/web/server.ts:971-990`
- `src/index.ts:2583-2597`

## Impact

When `DISCOVERY_TRUST_LAN_ADMIN=true` and no `WEB_PASSWORD` / `WEB_UI_USER`+`WEB_UI_PASS` is configured, privileged discovery admin routes can be reached without credentials if the listener is bound to a private address and the port is reachable from outside the LAN boundary.

Affected routes include discovery admin actions such as:

- `GET /api/discovery/requests`
- `POST /api/discovery/requests/:id/approve`
- `POST /api/discovery/requests/:id/reject`
- `POST /api/discovery/state`
- `POST /api/discovery/trusted-networks/current`
- `DELETE /api/discovery/trusted-networks/:id`
- peer/context proxy admin routes under `/api/discovery/peers/...`

This is configuration-dependent, but the current implementation does not actually verify that the *client* is on a trusted/private network before bypassing auth.

## Root cause

`isTrustedLanDiscoveryAdminRequest()` returns true when either of these is private:

- the actual socket remote address
- `listenerHostname`

The second condition is the problem:

```ts
return (
  isLoopbackOrPrivateAddress(remoteAddress) ||
  isLoopbackOrPrivateAddress(listenerHostname)
);
```

`listenerHostname` is the server's own configured bind host, not a property of the incoming client. If it is a LAN IP like `192.168.x.x`, any reachable client can satisfy the trust check even when `remoteAddress` is public.

## Reproduction

Code-level repro:

```ts
import { isTrustedLanDiscoveryAdminRequest } from './src/web/server.ts';

const req = new Request('http://example.com/api/discovery/requests');
Object.assign(req, { socket: { remoteAddress: '8.8.8.8' } });

console.log(
  isTrustedLanDiscoveryAdminRequest(
    req,
    '/api/discovery/requests',
    '192.168.1.10',
    true,
  ),
);
// actual: true
// expected: false
```

I also reproduced this directly in the repo with:

```bash
bun -e "import { isTrustedLanDiscoveryAdminRequest } from './src/web/server.ts'; const req = new Request('http://example.com/api/discovery/requests'); Object.assign(req,{socket:{remoteAddress:'8.8.8.8'}}); console.log(isTrustedLanDiscoveryAdminRequest(req,'/api/discovery/requests','192.168.1.10',true));"
```

Output:

```text
true
```

## Suggested fix

- Base the LAN trust decision on the actual socket peer address only.
- Keep the listener-host fallback only for the narrow case where the listener is loopback-bound *and* no socket address is available.
- Add a regression test that proves a public remote address is rejected even when `listenerHostname` is a private LAN address.

Peyton-Spencer:
## Summary

- fix `isTrustedLanDiscoveryAdminRequest()` so discovery admin auth bypass is based on the actual client socket peer instead of the Web UI listener bind host
- keep a loopback-only fallback when no socket address is available to preserve local-only behavior without extending LAN trust to public clients
- add a regression test covering a public remote address against a private listener hostname

## Testing

- `bun -e "import { isTrustedLanDiscoveryAdminRequest } from './src/web/server.ts'; const req = new Request('http://example.com/api/discovery/requests'); Object.assign(req,{socket:{remoteAddress:'8.8.8.8'}}); console.log(isTrustedLanDiscoveryAdminRequest(req,'/api/discovery/requests','192.168.1.10',true));"`

Closes #557

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened network discovery admin authentication by enforcing stricter peer address validation. Requests from public addresses are now properly rejected regardless of listener configuration.

* **Tests**
  * Added test coverage validating authentication rejection for requests originating from public network addresses.

### DIFF
--- a/src/web/server.test.ts
+++ b/src/web/server.test.ts
@@ -551,6 +551,26 @@ describe('isTrustedLanDiscoveryAdminRequest', () => {
     ).toBe(true);
   });
 
+  it('rejects public peers even when the listener is bound to a private LAN IP', () => {
+    const req = Object.assign(
+      new Request('http://evil.example/api/discovery/requests', {
+        headers: { Host: 'evil.example' },
+      }),
+      {
+        socket: { remoteAddress: '8.8.8.8' },
+      },
+    ) as Request;
+
+    expect(
+      isTrustedLanDiscoveryAdminRequest(
+        req,
+        '/api/discovery/requests',
+        '192.168.1.10',
+        true,
+      ),
+    ).toBe(false);
+  });
+
   it('allows loopback-bound servers when the socket address is unavailable', () => {
     const req = new Request('http://evil.example/api/discovery/requests', {
       headers: { Host: 'evil.example' },
--- a/src/web/server.ts
+++ b/src/web/server.ts
@@ -999,11 +999,26 @@ export function isTrustedLanDiscoveryAdminRequest(
     req as unknown as { socket?: { remoteAddress?: string } }
   ).socket?.remoteAddress;
 
-  // Only trust the actual socket peer or the configured listener host. URL/Host
-  // are attacker-controlled and must not influence auth decisions.
+  // Auth bypass must key off the actual peer address. The listener hostname is
+  // the server's own bind target, not a property of the client request.
+  if (remoteAddress) {
+    return isLoopbackOrPrivateAddress(remoteAddress);
+  }
+
+  // Bun may not always expose a socket address in tests or some deployments.
+  // Preserve the loopback-only fallback for local-only listeners.
+  return isLoopbackAddress(listenerHostname);
+}
+
+function isLoopbackAddress(address?: string): boolean {
+  if (!address) return false;
+  const normalized = address.toLowerCase();
+
   return (
-    isLoopbackOrPrivateAddress(remoteAddress) ||
-    isLoopbackOrPrivateAddress(listenerHostname)
+    normalized === '::1' ||
+    normalized === '::ffff:127.0.0.1' ||
+    normalized === 'localhost' ||
+    normalized.startsWith('127.')
   );
 }
 
@@ -1011,9 +1026,7 @@ function isLoopbackOrPrivateAddress(address?: string): boolean {
   if (!address) return false;
   const normalized = address.toLowerCase();
 
-  if (normalized === '::1' || normalized === '::ffff:127.0.0.1') return true;
-  if (normalized === 'localhost') return true;
-  if (normalized.startsWith('127.')) return true;
+  if (isLoopbackAddress(normalized)) return true;
 
   const ipv4 = normalized.startsWith('::ffff:')
     ? normalized.slice('::ffff:'.length)


src/web/server.ts
```diff
@@ -999,21 +999,34 @@ export function isTrustedLanDiscoveryAdminRequest(
...
+  // Auth bypass must key off the actual peer address. The listener hostname is
+  // the server's own bind target, not a property of the client request.
+  if (remoteAddress) {
+    return isLoopbackOrPrivateAddress(remoteAddress);
+  }
+
+  // Bun may not always expose a socket address in tests or some deployments.
+  // Preserve the loopback-only fallback for local-only listeners.
+  return isLoopbackAddress(listenerHostname);
+}
```
Peyton-Spencer, line 1011: Good separation. The two-branch structure reads clearly:

1. Socket peer exists → trust only if private/loopback
2. No socket info → trust only if the *server itself* is loopback-bound

The fallback intentionally narrows from `isLoopbackOrPrivateAddress` to `isLoopbackAddress`, which is the key hardening step — a server bound to `192.168.1.10` without socket info now correctly rejects instead of granting blanket trust.

src/web/server.test.ts
```diff
@@ -551,6 +551,26 @@ describe('isTrustedLanDiscoveryAdminRequest', () => {
...
+      isTrustedLanDiscoveryAdminRequest(
+        req,
+        '/api/discovery/requests',
+        '192.168.1.10',
+        true,
+      ),
+    ).toBe(false);
+  });
+
   it('allows loopback-bound servers when the socket address is unavailable', () => {
```
Peyton-Spencer, line 574: **Suggestion (non-blocking):** This test covers the exact reproduction from #557 — good. One complementary case worth adding: a private-LAN-bound server *without* socket info should also be rejected. This would guard against a future regression where someone changes the fallback back to `isLoopbackOrPrivateAddress`:

```ts
it('rejects private-LAN-bound servers when the socket address is unavailable', () => {
  const req = new Request('http://evil.example/api/discovery/requests', {
    headers: { Host: 'evil.example' },
  });

  expect(
    isTrustedLanDiscoveryAdminRequest(
      req,
      '/api/discovery/requests',
      '192.168.1.10',
      true,
    ),
  ).toBe(false);
});
```

This is the "no socket, LAN-bound" scenario — the fallback path that the old code got wrong.

Peyton-Spencer (commented):
Correct security fix — LGTM (can't self-approve via API). The root cause was clear: `listenerHostname` is a server property, not a client property, so it should never grant trust. The new logic properly gates on `remoteAddress` when available and falls back conservatively to loopback-only.

CI green across all checks. One optional test gap noted below.

PR #581, opened 2026-04-21
test(web): cover fallback and pairing failures

Peyton-Spencer:
## Summary
- Add a SolidStart fallback test that proves the web server falls back to the Datastar UI when the compiled bundle is missing.
- Add a WhatsApp auth failure-path test that verifies pairing-code setup logs an error and exits cleanly when `requestPairingCode()` throws.

## Verification
- `bun test`
- `bun test --coverage`

## Coverage Delta
- Test count: 1790 -> 1792
- Overall line coverage: 85.86% -> 85.95%
- Files improved:
  - `src/web/solid-handler.ts`: 93.15% -> 100.00% line coverage
  - `src/whatsapp-auth.ts`: 96.79% -> 98.08% line coverage

## Remaining Gaps
- `src/whatsapp-auth.ts` still has the `import.meta.main` process-exit wrapper uncovered at lines 238-239.
- Higher-value low-coverage areas still include `src/web/server.ts`, `src/web/routes.ts`, and `src/index.ts`.

## Summary by CodeRabbit

* **Tests**
  * Improved test coverage for error scenarios in authentication and bundle handling, ensuring graceful failure when external dependencies are unavailable.

### DIFF
--- a/src/web/solid-handler.test.ts
+++ b/src/web/solid-handler.test.ts
@@ -77,6 +77,17 @@ describe('solid-handler', () => {
     expect(isMainProcessRoute('/dashboard')).toBe(false);
   });
 
+  it('falls back to the Datastar UI when the SolidStart bundle is unavailable', async () => {
+    fs.rmSync(path.join(process.cwd(), 'webui', '.output'), {
+      recursive: true,
+      force: true,
+    });
+
+    const handler = await initSolidHandler(makeState());
+
+    expect(handler).toBeNull();
+  });
+
   it('initializes the proxy handler and forwards requests to the internal SolidStart server', async () => {
     fs.mkdirSync(solidOutputDir, { recursive: true });
     fs.writeFileSync(
--- a/src/whatsapp-auth.test.ts
+++ b/src/whatsapp-auth.test.ts
@@ -200,9 +200,11 @@ describe('connectSocket', () => {
 
   it('fails the flow when requesting a pairing code throws', async () => {
     const harness = createConnectDeps();
-    harness.socketHarness.requestPairingCode.mockImplementationOnce(async () => {
-      throw new Error('phone offline');
-    });
+    harness.socketHarness.requestPairingCode.mockImplementationOnce(
+      async () => {
+        throw new Error('phone offline');
+      },
+    );
 
     await connectSocket(
       '14155551234',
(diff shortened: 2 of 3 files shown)


src/web/solid-handler.test.ts
```diff
@@ -77,6 +77,17 @@ describe('solid-handler', () => {
     expect(isMainProcessRoute('/dashboard')).toBe(false);
   });
```
Peyton-Spencer, line 79: Nit: this `rmSync` is technically a no-op if the previous `afterEach` already cleaned up, but it makes the precondition explicit, which is fine.

One subtle thing to be aware of: `initSolidHandler` only assigns the module-level `solidHandler` on the *success* path. If the success test runs before this one and sets that module variable, calling `initSolidHandler` here returns `null` but doesn't clear the stale singleton. `getSolidHandler()` would still return the old handler. Not a regression from this PR (pre-existing), but worth noting if you ever add a `getSolidHandler()` assertion to this test.

src/whatsapp-auth.test.ts
```diff
@@ -198,6 +198,31 @@ describe('connectSocket', () => {
...
+        throw new Error('phone offline');
+      },
+    );
+
+    await connectSocket(
+      '14155551234',
+      false,
+      { usePairingCode: true },
+      harness.deps,
+    );
```
Peyton-Spencer, line 214: Good coverage of the `catch` branch. One thing worth locking in: the pairing-code failure path is the *only* failure exit that doesn't write to `STATUS_FILE` — every other `exit(1)` path (logged_out, timed_out, unknown) records the reason first. Consider adding:

```ts
// no status written — the pairing-code failure path exits without recording
expect(harness.writeFileSync).not.toHaveBeenCalled();
```

This would both document the current behavior and catch a future regression if someone adds a `writeFileSync` call to the catch block. That said, the missing status write itself might be a source-code gap worth a follow-up fix (write `failed:pairing_code` before `exit(1)`, consistent with the other branches).

Peyton-Spencer (commented):
Clean, focused additions — both tests correctly exercise their target failure paths and follow the existing harness patterns. Coverage delta looks right.

Two observations below (one nit, one source-level suggestion), neither blocking.

PR #582, opened 2026-04-21
fix(security): stream WhatsApp media downloads with limits

### LINKED ISSUE #576: security: WhatsApp media downloads buffer unbounded attachments before size rejection
## Summary

The WhatsApp adapter downloads media with Baileys' `downloadMediaMessage(..., 'buffer', ...)`, which buffers and decrypts the entire attachment before OmniClaw enforces its own byte cap. OmniClaw only rejects oversize payloads after the full `Buffer` has already been allocated, and the download path has no explicit timeout.

## Severity

- P1 / High

## Impact

- Any sender who can deliver media to a registered WhatsApp chat can force OmniClaw to allocate memory proportional to the full attachment size before rejection.
- A stalled or extremely slow media transfer can keep the message-processing path occupied indefinitely because there is no adapter-level timeout around `downloadMediaMessage()`.
- Repeated large media messages can cause memory pressure, long event-loop stalls, or process instability on deployments that expose WhatsApp ingress.

## Affected code

- `src/channels/whatsapp.ts:366-425`
- `src/channels/whatsapp.ts:669-689`

Relevant logic:

```ts
const buffer = (await downloadMediaMessage(
  msg,
  'buffer',
  {},
  {
    logger: { ...console, child: () => console } as any,
    reuploadRequest: this.sock.updateMediaMessage,
  },
)) as Buffer;

if (buffer.length > MAX_BINARY_DOWNLOAD_BYTES) {
  logger.warn(..., 'WhatsApp media exceeds size limit — discarding');
  return null;
}
```

## Why this is vulnerable

- The size check happens after Baileys has already returned the full `Buffer`.
- Unlike the Discord/Telegram/shared fetch-based media paths, this path does not stream through OmniClaw's bounded readers.
- The helper also lacks an explicit timeout, so the adapter depends on upstream library/network behavior instead of enforcing its own deadline.

## Reproduction

1. Run OmniClaw with a registered WhatsApp group or direct chat.
2. Send the bot a large image or image-like document that exceeds `MAX_BINARY_DOWNLOAD_BYTES`.
3. Observe that `downloadWhatsAppMedia()` still buffers the full payload before the adapter logs `WhatsApp media exceeds size limit — discarding`.
4. Repeat with multiple large uploads, or throttle the transfer aggressively, to observe memory growth and prolonged message-processing latency.

## Expected behavior

- Oversized WhatsApp media should be rejected before full buffering/decryption into memory.
- Media downloads should have an explicit timeout consistent with the other channel adapters.

## Suggested remediation

- Prefer a streaming/iterable Baileys download path so OmniClaw can enforce byte caps while reading.
- Add an adapter-level timeout around WhatsApp media downloads.
- If Baileys cannot provide a bounded streaming path, reject based on trustworthy metadata before download and treat unknown sizes conservatively.

Peyton-Spencer:
## Summary
- switch inbound WhatsApp media downloads to Baileys stream mode instead of buffering the full attachment first
- enforce the existing binary size cap while reading the decrypted media stream and abort stalled downloads after 15 seconds
- add regression coverage for successful bounded downloads, oversize rejection, and timeout handling

## Testing
- `bun test src/channels/whatsapp.test.ts`
- `bun run typecheck`

## Security
- fixes #576

## Release Notes

* **Tests**
  * Added tests for media download streaming behavior, size enforcement, and timeout handling.

* **Improvements**
  * Enhanced WhatsApp media downloads with streaming support and robust size enforcement.
  * Added timeout protection for media download operations.
  * Improved error logging for media download failures.

### DIFF
--- a/src/channels/whatsapp.test.ts
+++ b/src/channels/whatsapp.test.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import path from 'path';
+import { Readable } from 'stream';
 import {
   afterEach,
   describe,
@@ -12,6 +13,7 @@ import {
 
 import { GROUPS_DIR } from '../config.js';
 import {
+  MAX_BINARY_DOWNLOAD_BYTES,
   formatImageMarker,
   formatPlaceholder,
   formatTextFileMarker,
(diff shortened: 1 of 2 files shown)


src/channels/whatsapp.ts
```diff
@@ -34,6 +34,39 @@ import {
...
+  const chunks: Buffer[] = [];
+  let totalBytes = 0;
+
+  try {
+    for await (const chunk of stream) {
+      const bytes = Buffer.from(chunk);
+      totalBytes += bytes.length;
+
+      if (totalBytes > maxBytes) {
+        const error = new Error(`Download exceeded ${maxBytes} bytes`);
```
Peyton-Spencer, line 56: Minor: double `destroy()` on the oversize path. When `totalBytes > maxBytes`, `destroy` is called here, the error is thrown, and the `catch` block on line 139 calls `destroy` again with the same error. This is safe — `Readable.destroy()` is idempotent — but it's worth knowing it happens. If you want to tighten it up:

```ts
} catch (error) {
  if (!stream.destroyed) {
    stream.destroy?.(error instanceof Error ? error : undefined);
  }
  throw error;
}
```

Non-blocking — Node guarantees idempotency here.

src/channels/whatsapp.test.ts
```diff
@@ -942,6 +944,83 @@ describe('WhatsAppChannel.downloadWhatsAppMedia', () => {
     });
     expect(result).toBeNull();
   });
+
+  it('downloads streamed media under the byte limit', async () => {
+    const channel = makeChannel();
+    const mediaDownloader = mock(async (_msg: any, type: string, opts: any) => {
+      expect(type).toBe('stream');
```
Peyton-Spencer, line 951: Good boundary test with a single chunk at `MAX_BINARY_DOWNLOAD_BYTES + 1`. For completeness you could add a multi-chunk variant — e.g. two chunks that are individually under the limit but collectively exceed it:

```ts
const half = Math.ceil(MAX_BINARY_DOWNLOAD_BYTES / 2) + 1;
const stream = Readable.from([Buffer.alloc(half), Buffer.alloc(half)]);
```

This exercises the accumulation path in the `for await` loop. Optional — the current test already proves the size check works.

Peyton-Spencer (commented):
Solid security fix — streaming with early abort directly addresses the unbounded buffering vulnerability from #576. The `AbortSignal.timeout` + byte-capped stream reader is the right approach. Tests cover the three key paths (success, oversize, timeout). LGTM. Two minor observations below, neither blocking.

Peyton-Spencer:
I reviewed this one and do not see any blockers.

Streaming the Baileys download instead of buffering first is the right fix here, and the added tests cover the key bounded-download paths: successful download, oversize rejection, and timeout handling.

I cannot leave a formal approval because the authenticated GitHub account is the PR author, but from my side this looks ready to merge.

PR #556, opened 2026-04-16
test: cover index available groups, logger, and path security

Peyton-Spencer:
## Summary

- add deterministic `src/index.test.ts` coverage for available-group discovery, including channel filtering, sort order, exact registration, and legacy Telegram JID compatibility
- keep the earlier logger and Effect path-security test additions on this branch, expanding direct coverage on core orchestration helpers without production code changes
- document current repo-wide test status from this workspace: targeted changed files pass, but full `bun test` still reports unrelated existing failures concentrated in simtest, web, and discovery suites

## Verification

- `bun install`
- `bun test src/index.test.ts`
- `bun test` *(currently fails on pre-existing unrelated tests: 1557 passed, 242 failed, 1799 total)*

## Coverage Notes

- test count on this branch increased from `1795` to `1799` via the four new `src/index.test.ts` cases
- newly covered file: `src/index.ts` (`getAvailableGroups()` and legacy registration resolution paths)
- previously added coverage on this branch remains in `src/logger.ts` and `src/path-security.ts`

## Remaining Gaps

- full-suite failures remain in unrelated areas including `src/simtest/admin-api.test.ts`, `src/simtest/fake-state.test.ts`, `src/web/server.test.ts`, `src/web/system.test.ts`, `src/discovery/routes.test.ts`, and `src/discovery/peer-client.test.ts`
- this change does not address the broader red full-suite baseline in the current workspace

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for logging functionality with new pretty-mode output validation tests.
  * Added comprehensive test suite for path security validation using Effect-based APIs.
  * Introduced new tests for group availability and filtering logic.

Issue #550, opened 2026-04-16
test: web module mocks leak across files and break 242 tests on main


Peyton-Spencer:
# test: web module mocks leak across files and break 242 tests on main

## Summary

`bun test` on `main` is failing with 242 test failures. The highest-confidence root cause is `src/web/index.test.ts` using `mock.module()` for `./server.js` and `./ipc-events.js`, then later test files observing those mocked exports instead of the real implementations.

## Reproduction

1. Sync and install:
   ```bash
   git fetch --all --prune
   bun install
   ```
2. Run:
   ```bash
   bun test
   ```

## Observed result

- `bun test` reports:
  - `1545 pass`
  - `242 fail`
  - `Ran 1787 tests across 98 files`
- Representative failures:
  - `src/web/ipc-events.test.ts`
    - `TypeError: new IpcEventBuffer().push is not a function`
  - many `src/web/*.test.ts` failures
    - `TypeError: undefined is not an object (evaluating 'handle.port')`
    - `ConnectionRefused` when fetching `http://localhost:3000/...`
  - `src/simtest/admin-api.test.ts`
    - route tests fail after the mocked web server export is observed instead of the real implementation

## Why this looks like mock leakage

- `src/web/index.test.ts` contains:
  - `mock.module('./server.js', () => ({ startWebServer }))`
  - `mock.module('./ipc-events.js', () => ({ IpcEventBuffer }))`
- The mocked `IpcEventBuffer` class in that test has no `push()` method.
- The later `ipc-events` failures match that exact shape.
- The later route/server failures are also consistent with `startWebServer` being replaced by a no-op mock, leaving `handle` unset.

## Expected result

- `bun test` should pass on `main`
- `src/web/index.test.ts` should not affect later files

## Suggested fix direction

- Remove cross-file `mock.module()` usage from `src/web/index.test.ts`, or isolate that test so mocked modules cannot leak into the broader run
- Re-run the full suite, not just the export test, because this regression currently masks the real health of the web test surface

## Notes

- This looks like a recent Web UI test regression rather than an environmental flake; the failure signature is deterministic and internally consistent

Peyton-Spencer:
Re-ran `bun test` on current `main` after syncing and installing in `/workspace/group/omniclaw-repo`.

This issue is still valid. Current counts are:

```text
1553 pass
242 fail
1795 total
```

[...]

PR #586, opened 2026-04-22
test(ipc): cover format_mention response writes

Peyton-Spencer:
## Summary
- Add deterministic `ipc.ts` coverage for `format_mention` response file writes, including sanitized request IDs and platform-specific mention formatting.
- Verify fallback behavior when the user registry is unreadable so agents still receive usable mention responses instead of broken IPC output.

## Testing
- `bun test src/ipc-format-mention-response.test.ts`
- `bun test`

## Test Count
- Before: 1563 passing tests
- After: 1568 passing tests
- Delta: +5 tests

## Files Covered
- `src/ipc.ts`
- `src/ipc-format-mention-response.test.ts`

## Remaining Coverage Gaps
- `src/ipc.ts` still has watcher and task-processing branches that are only partially covered through broader integration-style tests.
- `src/discovery/types.ts` and `src/web/types.ts` are type-only modules without runtime behavior to unit test directly.
- `src/backends/local-backend.isolated.ts` contains isolated test-only coverage and is not part of the main suite's normal source/test pairing.

## Notes
- `bunx tsc --noEmit` still reports pre-existing type errors in `src/discovery/network-identity.test.ts` and `src/logger.test.ts` unrelated to this change.

## Tests

* **Expanded test coverage** across multiple functionality areas including group chat management, inter-process communication handling, logging, and file path security validation.
* **Enhanced quality assurance** with comprehensive test suites that verify system behavior and improve reliability.

### DIFF
--- a/src/ipc-format-mention-response.test.ts
+++ b/src/ipc-format-mention-response.test.ts
@@ -1,0 +1,174 @@
+import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+import { processMessageIpc, type IpcDeps, type MessageResult } from './ipc.js';
+import type { IpcMessagePayload, RegisteredGroup } from './types.js';
+
+const MAIN_GROUP: RegisteredGroup = {
+  name: 'Main',
+  folder: 'main',
+  trigger: 'always',
+  added_at: '2024-01-01T00:00:00.000Z',
+};
+
+const OTHER_GROUP: RegisteredGroup = {
+  name: 'Other',
+  folder: 'other-group',
+  trigger: '@Other',
+  added_at: '2024-01-01T00:00:00.000Z',
+};
+
+describe('processMessageIpc: format_mention response files', () => {
+  // Keep response-file assertions in a separate file because bun:test module
+  // mocks leak across files, and group-queue.test.ts globally mocks fs.
+  let tmpDir: string;
+  let groups: Record<string, RegisteredGroup>;
+  let deps: IpcDeps;
+
+  beforeEach(() => {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'ipc-format-mention-'));
+    groups = {
+      'main@g.us': MAIN_GROUP,
+      'other@g.us': OTHER_GROUP,
+    };
+    deps = {
+      sendMessage: async () => 'sent-1',
+      notifyGroup: () => {},
+      registeredGroups: () => groups,
+      registerGroup: () => {},
+      updateGroup: () => {},
+      syncGroupMetadata: async () => {},
+      getAvailableGroups: () => [],
+      writeGroupsSnapshot: () => {},
+    };
+  });
+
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
The core contribution — `ipc-format-mention-response.test.ts` — is solid and fills a genuine gap. The existing `ipc-messages.test.ts` format_mention tests explicitly note they can't verify response files on disk due to the fs mock leakage issue (#550). This new file, running in isolation, validates the actual file writes: sanitized requestIds, platform-specific mention formats, corrupt-registry fallback, and the empty-after-sanitization blocking path. Well structured.

A few observations on the rest of the PR below.

src/index.test.ts
```diff
@@ -0,0 +1,109 @@
+import { beforeEach, describe, expect, it } from 'bun:test';
```
Peyton-Spencer: 3 of 4 tests here substantially overlap with `routing.test.ts` (which already covers filtering, ordering, registration marking, Discord DM exclusion, and Telegram legacy JID resolution). The Slack filtering test (line 95) is the only genuinely new coverage.

Consider either:
- Moving just the Slack test into `routing.test.ts` and dropping this file, or
- Adding a comment explaining why a separate `index.test.ts` is preferred over extending `routing.test.ts`

Not blocking — defense-in-depth is fine if intentional, but as-is it reads like accidental duplication.

src/ipc-format-mention-response.test.ts
```diff
@@ -0,0 +1,172 @@
+import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
```
Peyton-Spencer, line 1: Nice. This works around the mock leakage problem documented in the existing `ipc-messages.test.ts` (`group-queue.test.ts` globally mocks `fs`, and bun:test mocks leak across files). Running response-file assertions in a dedicated file is the right call until #550 is resolved.

Minor: a one-line comment at the top referencing the mock-leakage reason (like the note in `ipc-messages.test.ts:368-369`) would help future readers understand why this file exists separately.

src/logger.test.ts
```diff
@@ -458,5 +502,20 @@ describe('logger', () => {
       expect(badSubscriber).toHaveBeenCalledTimes(1);
       expect(goodSubscriber).toHaveBeenCalledTimes(1);
     });
```
Peyton-Spencer, line 460: This relies on `logger` (the module-level singleton) having a level that allows `info` events to reach subscribers. Since subscribers receive all log entries regardless of the output level filter, this should be safe — but it is the same concern flagged on PR #556. Worth confirming that `logger.subscribe()` is truly level-independent so this test stays deterministic.

src/path-security.test.ts
```diff
@@ -163,3 +170,62 @@ describe('assertPathWithin', () => {
     );
   });
 });
```
Peyton-Spencer, line 165: Clean Effect API coverage. Good use of `Effect.either` + `Either.isLeft`/`isRight` to capture typed errors without throwing. The `_tag` assertion on `PathTraversalError` confirms the tagged-error pattern works end-to-end.

Issue #587, opened 2026-04-22
security: remote image cache remains vulnerable to DNS rebinding SSRF


Peyton-Spencer:
## Summary

`serveCachedRemoteImage()` is still vulnerable to DNS rebinding SSRF. The code validates a hostname up front with `validateRemoteImageUrl()`, but later performs the actual `fetch()` by hostname again. An attacker-controlled domain can therefore pass validation while it resolves to a public IP, then rebind to loopback, RFC1918, or link-local targets before the image is fetched.

This is explicitly acknowledged in the source comment at `src/web/image-cache.ts`:

> This still has a DNS rebinding/TOCTOU gap because fetch() resolves the hostname again.

## Severity

P1 security bug.

The impact is stored SSRF from the OmniClaw host into internal services such as:

- loopback-only web services
- RFC1918 LAN services
- cloud metadata endpoints like `169.254.169.254`

## Affected code

- `src/web/image-cache.ts:103-145`
- `src/web/image-cache.ts:236-339`
- `src/web/routes.ts:1065-1073`
- `src/web/routes.ts:977-1019`

## Reproduction

1. Configure a custom avatar URL through the web UI using an attacker-controlled hostname such as `https://avatar.attacker.test/pic.png`.
2. Ensure the hostname initially resolves to a public IP so `validateRemoteImageUrl()` accepts it.
3. After the URL is stored, change DNS for that hostname to `127.0.0.1`, an RFC1918 address, or `169.254.169.254`.
4. Load the avatar image endpoint, for example `/api/agents/<agentId>/avatar/image`.
5. OmniClaw re-resolves the hostname during `fetch()` and sends the request to the rebound internal target.

## Why the current defenses are insufficient

- Write-time validation only checks the hostname before persistence.
- Fetch-time validation calls `validateRemoteImageUrl()` again, but `fetch()` still performs a fresh DNS resolution afterward.
- That leaves a classic time-of-check/time-of-use gap for rebinding.

## Suggested fix direction

One of these needs to happen before the issue is actually closed:

- resolve the hostname once and connect directly to the verified IP while preserving the original `Host` header / SNI requirements safely, or
- disallow hostname-based remote image fetches entirely and only permit direct public IP URLs, or
- proxy remote image fetches through a hardened allowlist-based fetcher that pins the resolved address and rejects redirects / re-resolution to private ranges.

Tests should cover a hostname that validates as public but whose fetch step attempts to connect to a private address afterward.

PR #588, opened 2026-04-23
fix(security): close DNS rebinding SSRF in remote image cache

Peyton-Spencer:
## Summary

Closes the DNS rebinding / TOCTOU SSRF vulnerability described in #587.

`serveCachedRemoteImage` previously called `validateRemoteImageUrl` (which resolves DNS and checks the addresses are public), then passed the original hostname URL to `fetch()` — which re-resolves DNS independently. An attacker-controlled domain could pass validation while pointing to a public IP, then rebind to `127.0.0.1`, RFC1918, or cloud metadata endpoints before the fetch executes.

### Fix

- New `resolveAndValidateRemoteImageUrl` resolves DNS once and returns both the validation result and the resolved IP addresses
- New `fetchWithPinnedDns` uses `node:http`/`node:https` with a custom `lookup` callback that always returns the pre-validated address — the TCP connection goes to the exact IP we checked, with no DNS re-resolution
- `serveCachedRemoteImage` routes hostname URLs through pinned fetch, direct-IP URLs through standard `fetch`, and test-injected `fetchImpl` when provided
- `validateRemoteImageUrl` kept as a backward-compatible wrapper (used in `routes.ts` write-time validation)

### New exports

| Function | Purpose |
|---|---|
| `resolveAndValidateRemoteImageUrl` | Validates + returns resolved addresses |
| `fetchWithPinnedDns` | HTTP(S) fetch pinned to a specific IP |
| `RemoteImageUrlValidationResult` | Return type for the above |

### Tests (9 new, 32 total for this file)

- `resolveAndValidateRemoteImageUrl` returns addresses for hostnames, empty for direct IPs, error for blocked
- `fetchWithPinnedDns` connects to pinned address with correct Host header
- `fetchWithPinnedDns` times out correctly
- `serveCachedRemoteImage` blocks when DNS resolves to private address
- `serveCachedRemoteImage` passes `lookupHostAddresses` through to validation

### Test results

- 1799 pass, 0 fail across 95 files
- `tsc --noEmit` clean

## Test plan

- [x] `bun test src/web/image-cache.test.ts` — 32 pass, 0 fail
- [x] `bun test` — 1799 pass, 0 fail
- [ ] CI green on PR

## Summary by CodeRabbit

* **New Features**
  * Remote image validation now includes DNS pre-resolution and returns resolved addresses; requests to private/loopback addresses are blocked.
  * Remote image fetches can use a pinned IP for the connection while preserving the original hostname in the request; request timeouts supported.

* **Tests**
  * Added unit and integration tests for DNS validation, pinned-IP fetching behavior, header preservation, blocking cases, and timeout handling.

### DIFF
--- a/src/web/image-cache.test.ts
+++ b/src/web/image-cache.test.ts
@@ -1,5 +1,6 @@
 import { randomUUID } from 'crypto';
 import fs from 'fs';
+import http from 'node:http';
 import path from 'path';
 
 import { describe, expect, it } from 'bun:test';
@@ -8,8 +9,10 @@ import { DATA_DIR } from '../config.js';
 import { logger } from '../logger.js';
 import {
   describeImageUrl,
+  fetchWithPinnedDns,
   readStreamWithCap,
   type RemoteImageFetch,
+  resolveAndValidateRemoteImageUrl,
   serveCachedRemoteImage,
   validateRemoteImageUrl,
 } from './image-cache.js';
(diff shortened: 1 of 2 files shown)


Peyton-Spencer:
## Mergemaster Review — HOLD (NEEDS_WORK)

The core approach (DNS pre-resolution + pinned fetch via custom `lookup` callback) is the right pattern. However, three gaps need addressing before merge:

### GAP A — CRITICAL: `fetchImpl` bypasses DNS pinning

Lines 412-428 give `fetchImpl` precedence over `fetchWithPinnedDns`. In production, `routes.ts` passes `state.fetchRemoteImage` as `fetchImpl` for all three call sites (avatars, chat icons, guild icons). **All production fetches bypass DNS pinning entirely**, re-opening the exact TOCTOU window this PR aims to close.

**Fix**: When `validation.resolvedAddresses.length > 0`, always use `fetchWithPinnedDns`. Only fall back to `fetchImpl` for direct-IP URLs where no pinning is needed.

### GAP B — MAJOR: Empty DNS resolution treated as valid

If `resolveHostAddresses()` returns `[]`, the function returns `{ error: null, resolvedAddresses: [] }`. Back in `serveCachedRemoteImage`, `resolvedAddresses.length > 0` is false, so the code falls through to direct `fetch()` — which re-resolves DNS independently.

**Fix**: After `await resolveHostAddresses(hostname)`, check `addresses.length === 0` and return an error. Fail closed.

### GAP C — MAJOR (functional): No redirect handling

`fetchWithPinnedDns` uses `http.request`/`https.request` which do NOT follow redirects. Images behind CDN redirects will silently break. Not a direct SSRF hole, but a functional regression. Document the tradeoff or add per-hop DNS-validated redirect following.

### Test gaps
- Test at line 513-554 calls `fetchWithPinnedDns` directly but never exercises `serveCachedRemoteImage`
- No test for empty-addresses edge case
- No test proving `fetchImpl` doesn't bypass pinning for hostname URLs

All three CodeRabbit findings are valid.

Peyton-Spencer (commented):
## Blocking: `fetchImpl` bypasses DNS pinning

The critical issue (also flagged by CodeRabbit at line 428) is that **all production callers in `src/web/routes.ts` pass `fetchImpl`**, which takes precedence at line 416 and skips the pinned DNS fetch entirely.

This means the DNS rebinding fix is effectively bypassed for every real request path. The SSRF vulnerability remains open because:
1. `serveCachedRemoteImage()` validates the hostname and pins DNS
2. But then `fetchImpl` (provided by the caller) re-resolves DNS independently
3. An attacker can pass validation, then DNS rebinds to an internal IP before `fetchImpl` makes the actual request

**Fix needed:** Either:
- Remove `fetchImpl` support for hostname URLs (force all hostname fetches through `fetchWithPinnedDns`)
- Or thread the pinned IP address through to `fetchImpl` so it can't re-resolve

This needs a design change, not just a tweak. Holding merge until resolved.

Peyton-Spencer:
Focused verification passed, but I’m holding merge on one likely regression:  uses / directly, so it no longer follows upstream redirects the way  did before. That can turn previously valid remote image URLs into failures if they 301/302 to a CDN/object URL. Please either preserve redirect-follow behavior in the pinned path or add an explicit regression test showing redirects are intentionally unsupported.

Peyton-Spencer:
Focused verification passed, but I’m holding merge on one likely regression:

`fetchWithPinnedDns` uses Node's `http`/`https` request path directly, so it no longer follows upstream redirects the way `fetch()` did before. That can turn previously valid remote image URLs into failures if they return a 301/302 to the actual CDN or object URL.

Please either preserve redirect-follow behavior in the pinned path or add an explicit regression test demonstrating that redirects are intentionally unsupported.

PR #594, opened 2026-04-23
test(control-plane): cover route and startup branches

Peyton-Spencer:
## Summary
- add deterministic control-plane tests for fixed-time health responses, task listing, and 400/404 task-action branches
- monkey-patch `Bun.serve` and `logger.info` to verify `startControlPlaneServer()` wires the fetch handler and startup logging correctly
- increase overall test count from 1851 to 1856 and coverage from 88.50%/86.48% to 88.64%/86.79%

## Verification
- `bunx tsc --noEmit`
- `bun test`
- `bun test --coverage`

## Coverage Notes
- covered file: `src/control-plane.ts`
- before: `src/control-plane.ts` at 88.89% funcs / 71.74% lines
- after: `src/control-plane.ts` at 100.00% funcs / 97.83% lines

## Remaining Gaps
- `src/github-linked.ts` remains lightly covered at 25.00% funcs / 24.49% lines
- `src/simtest/admin-api.ts` and `src/simtest/discovery-sim.ts` still have useful simulation-test headroom
- large runtime-heavy modules like `src/index.ts` and channel adapters still need more targeted extraction or seam-based tests

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for health endpoints, task listing operations, error handling and response codes, and server startup procedures.

### DIFF
--- a/src/control-plane.test.ts
+++ b/src/control-plane.test.ts
@@ -1,10 +1,41 @@
-import { describe, expect, it, mock } from 'bun:test';
+import { afterEach, describe, expect, it, mock } from 'bun:test';
 
 import {
   createControlPlaneFetch,
+  startControlPlaneServer,
   type ControlPlaneDeps,
 } from './control-plane.js';
+import { logger } from './logger.js';
 import type { RegisteredGroup, ScheduledTask } from './types.js';
+
+const RealDate = Date;
+const realBunServe = Bun.serve;
+const realLoggerInfo = logger.info;
+
+function installFixedDate(iso: string) {
+  const fixedTime = new RealDate(iso).getTime();
+  class FixedDate extends RealDate {
+    constructor(value?: string | number | Date) {
+      if (value !== undefined) {
+        super(value);
+        return;
+      }
+      super(fixedTime);
+    }
+
+    static override now() {
+      return fixedTime;
+    }
+  }
+
+  globalThis.Date = FixedDate as DateConstructor;
+}
+
+afterEach(() => {
+  globalThis.Date = RealDate;
+  Bun.serve = realBunServe;
+  logger.info = realLoggerInfo;
+});
 
 function makeTask(id: string, status: ScheduledTask['status']): ScheduledTask {
   return {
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
Solid test-only PR. The new tests are well-structured, deterministic, and meaningfully close the coverage gap on `control-plane.ts` (71.74% → 97.83% lines, 100% functions). CI is green across all checks.

**What's good:**
- `installFixedDate` is a clean approach — subclassing `Date` to pin `new Date()` without breaking `new Date(existingValue)` callers.
- Top-level `afterEach` cleanup is a sensible safety net for the globals being monkey-patched.
- The new route tests (`task list`, `400 on non-not_found failure`, `404 on GET to action route`, `404 catch-all`) close real coverage gaps — these were all distinct branches in `createControlPlaneFetch`.
- The `startControlPlaneServer` test correctly extracts the fetch handler from the `Bun.serve` options and invokes it directly, avoiding a real server bind. The `logger.info` assertion verifies the structured log shape exactly.

**Minor observations (non-blocking):**

1. The "returns 404 for unsupported methods on task actions" test name suggests a 405-like behavior, but the production code actually falls through to the generic 404 catch-all (the route matcher only considers `POST && regex`). The test correctly matches production behavior — just noting that if the product ever wants proper 405 responses for known paths + wrong methods, this would need updating.

2. The remaining ~2% uncovered lines are the Zod validation failure branch after `taskActionSchema.safeParse`. This is effectively unreachable because the regex already constrains `action` to the enum values and `([^/]+)` guarantees a non-empty `taskId`. Could be worth a comment in the test file noting this is dead code by construction, but not a blocker.

3. The `serveMockFn` type cast (`as unknown as typeof Bun.serve`) is fine for a test seam — the real `Bun.serve` return type is too complex to fully mock for a unit test.

Approving — clean work, correct seam choices, good coverage improvement.

src/control-plane.test.ts
```diff
@@ -1,11 +1,42 @@
...
+import { logger } from './logger.js';
 import type { RegisteredGroup, ScheduledTask } from './types.js';
 
+const RealDate = Date;
+const realBunServe = Bun.serve;
+const realLoggerInfo = logger.info;
+
+function installFixedDate(iso: string) {
+  const fixedTime = new RealDate(iso).getTime();
+  class FixedDate extends RealDate {
```
Peyton-Spencer, line 17: Nit: `Date.now()` override is defined but never exercised by any test. Fine as a defensive measure — just noting it for completeness.

src/control-plane.test.ts
```diff
@@ -1,11 +1,42 @@
...
+      super(fixedTime);
+    }
+
+    static override now() {
+      return fixedTime;
+    }
+  }
+
+  globalThis.Date = FixedDate as DateConstructor;
+}
```
Peyton-Spencer, line 32: Minor: this test name says "unsupported methods" but the production behavior is a generic 404 catch-all (not a 405). Accurate to the code — just noting the semantic gap in case you ever want to distinguish "known path, wrong method" from "truly unknown route."

Peyton-Spencer:
Verified locally in a clean PR worktree: - \bun test v1.3.11 (af24e281) passes (10/10)- \ passesI did not find correctness issues in the new test seam; it exercises the public fetch/server entrypoints rather than private helpers.

Peyton-Spencer:
Verified locally in a clean PR worktree:

- `bun test src/control-plane.test.ts` passes (10/10)
- `bun run typecheck` passes

I did not find correctness issues in the new test seam; it exercises the public fetch/server entrypoints rather than private helpers.

PR #619, opened 2026-05-02
feat(web): add CPU cores and load average to system health

Peyton-Spencer:
## Summary

Adds CPU cores and 1m/5m/15m load averages to the Web UI System page. Lightweight observability addition that uses `os.cpus()` and `os.loadavg()` (no new deps, no extra cost).

- New `cpu` block in `HealthData` returned by `/api/health`: `count`, `load_1m`, `load_5m`, `load_15m`
- New "cpu" metric card on `/system` with stable IDs (`sys-cpu-count`, `sys-cpu-load-1m`, `sys-cpu-load-5m`, `sys-cpu-load-15m`)
- The existing `systemScript()` updater (5s polling against `/api/health`) now refreshes the new fields when present
- Backwards compatible — the script guards on `if(h.cpu)` so older serialized health payloads don't break the live update

## Why

`memory` and `containers` are already exposed but operators have no signal on host CPU pressure. Load average is a cheap, well-understood metric and surfaces saturation when many agent containers are spawning at once.

## Test plan

- [x] `bun test` — 1904 pass, 0 fail (5 new assertions across 1 new test + augmented coverage)
- [x] `bunx prettier --check` on modified files — clean
- [ ] Manual: open `/system`, verify the cpu card renders with cores + 3 load values, watch them update every 5s

## Phase 3 fit

Slots into the "live features / polish" track in #505 — System page now has parity with what an `htop` glance would tell you.

## Summary by CodeRabbit

* **New Features**
  * System health monitoring now displays CPU metrics, including core count and 1-minute, 5-minute, and 15-minute load averages.

* **Tests**
  * Added test coverage for CPU metrics validation and UI rendering.


Peyton-Spencer (commented):
LGTM. Reviewed the CPU health payload and system card wiring. Local checks passed: `bun test src/web/`, `bun run typecheck`, and a direct `os.loadavg()` sanity check in this Linux container returned three finite values.

src/web/system.ts
```diff
@@ -91,6 +100,12 @@ export function buildHealthData(
       heap_used_mb: Math.round((mem.heapUsed / 1024 / 1024) * 10) / 10,
       heap_total_mb: Math.round((mem.heapTotal / 1024 / 1024) * 10) / 10,
     },
+    cpu: {
```
Peyton-Spencer, line 103: Confirmed the CPU payload is safe here: `os.loadavg()` is read once, each index has a `?? 0` fallback before rounding, and the rendered card IDs (`sys-cpu-count`, `sys-cpu-load-1m`, `sys-cpu-load-5m`, `sys-cpu-load-15m`) match the `systemScript()` selectors.

PR #621, opened 2026-05-02
test(web): cover session expiry and network wifi rendering

Peyton-Spencer:
## Summary

- Add deterministic `Date.now` tests for session expiry, purge, and max-session eviction.
- Add network page rendering coverage for current/trusted Wi-Fi state and escaped network identifiers.

## Test Count / Coverage

- Baseline tracked suite: `1904 pass, 0 fail`, coverage `88.79% funcs / 87.17% lines`.
- After changes tracked suite: `1907 pass, 0 fail`, coverage `88.97% funcs / 87.38% lines`.
- Files now fully covered by this PR: `src/web/session-auth.ts`, `src/web/network.ts`.

## Verification

- `bun test src/web/session-auth.test.ts src/web/network.test.ts`
- `bun test $(git ls-files '*test.ts')`
- `bun test --coverage $(git ls-files '*test.ts')`

## Note

The shared workspace contains unrelated pre-existing untracked test files, so plain `bun test` also picks them up and fails outside this PR. The tracked test suite passes cleanly.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for network content rendering with proper identifier escaping.
  * Expanded session store test coverage for expired session handling and capacity eviction scenarios.

### DIFF
--- a/src/web/network.test.ts
+++ b/src/web/network.test.ts
@@ -138,8 +138,16 @@ describe('renderNetworkContent', () => {
       runtime: makeRuntime({
         currentNetwork: { id: 'ssid:office<&>', label: 'Office <Wi-Fi>' },
         trustedNetworks: [
-          { id: 'ssid:home<&>', label: 'Home <Wi-Fi>' },
-          { id: 'ssid:lab', label: 'Lab Network' },
+          {
+            id: 'ssid:home<&>',
+            label: 'Home <Wi-Fi>',
+            trustedAt: '2026-03-01T00:00:00.000Z',
+          },
+          {
+            id: 'ssid:lab',
+            label: 'Lab Network',
+            trustedAt: '2026-03-02T00:00:00.000Z',
+          },
         ],
       }),
       peers: [],
--- a/src/web/network.test.ts
+++ b/src/web/network.test.ts
@@ -151,6 +151,8 @@ describe('renderNetworkContent', () => {
     );
     expect(html).toContain('<strong>Home &lt;Wi-Fi&gt;</strong>');
     expect(html).toContain('ssid:home&lt;&amp;&gt;');
+    expect(html).toContain('<strong>Lab Network</strong>');
+    expect(html).toContain('ssid:lab');
     expect(html).toContain('data-network-action="untrust-network"');
     expect(html).not.toContain('No trusted Wi-Fi networks yet.');
   });
--- a/src/web/network.test.ts
+++ b/src/web/network.test.ts
@@ -128,6 +128,32 @@ describe('renderNetworkContent', () => {
       'Discovery controls are unavailable in this environment.',
     );
   });
+
+  it('renders current and trusted Wi-Fi networks with escaped identifiers', () => {
+    const html = renderNetworkContent({
+      instanceId: 'local-id',
+      instanceName: 'Main Node',
+      discoveryAvailable: true,
+      discoveryEnabled: true,
+      runtime: makeRuntime({
+        currentNetwork: { id: 'ssid:office<&>', label: 'Office <Wi-Fi>' },
+        trustedNetworks: [
+          { id: 'ssid:home<&>', label: 'Home <Wi-Fi>' },
+          { id: 'ssid:lab', label: 'Lab Network' },
+        ],
+      }),
+      peers: [],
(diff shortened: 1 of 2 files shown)


src/web/network.test.ts
```diff
@@ -128,6 +128,32 @@ describe('renderNetworkContent', () => {
...
+      pendingRequests: [],
+    });
+
+    expect(html).toContain(
+      'Current Wi-Fi: <strong>Office &lt;Wi-Fi&gt;</strong>',
+    );
+    expect(html).toContain('<strong>Home &lt;Wi-Fi&gt;</strong>');
+    expect(html).toContain('ssid:home&lt;&amp;&gt;');
+    expect(html).toContain('data-network-action="untrust-network"');
+    expect(html).not.toContain('No trusted Wi-Fi networks yet.');
```
Peyton-Spencer, line 165: [nit] Consider also asserting the second trusted entry renders, e.g. `expect(html).toContain('<strong>Lab Network</strong>')` and `expect(html).toContain('ssid:lab')`. Right now only the escaped row is asserted, so a regression that drops non-escaped entries (or only emits the first network) wouldn't be caught.

src/web/session-auth.test.ts
```diff
@@ -113,6 +113,50 @@ describe('SessionStore', () => {
...
+  it('invalidates expired sessions during validation and purge', () => {
+    const now = spyOn(Date, 'now');
+    try {
+      now.mockReturnValue(1_000);
+      const store = createSessionStore();
+      const token = store.create();
+      const purgeToken = store.create();
+      expect(store.size).toBe(2);
+
+      now.mockReturnValue(1_000 + 24 * 60 * 60 * 1000);
```
Peyton-Spencer, line 127: [nit] This pins the boundary at exactly `expiresAt`, which is the `<=` edge in `session-auth.ts:74`. Worth a comment so the intent (boundary expiry, not just "some time later") is obvious to future readers, e.g. `// advance to exactly expiresAt — store uses <= so this should expire`.

src/web/session-auth.test.ts
```diff
@@ -113,6 +113,50 @@ describe('SessionStore', () => {
...
+      const newest = store.create();
+
+      expect(store.size).toBe(100);
+      expect(store.validate(tokens[0])).toBe(false);
+      expect(store.validate(tokens[1])).toBe(true);
+      expect(store.validate(newest)).toBe(true);
+    } finally {
+      now.mockRestore();
+    }
+  });
```
Peyton-Spencer, line 160: [nit] Coverage gap worth a follow-up: the `create()` path calls `purge()` *before* checking capacity. A test that fills the store with expired sessions and then creates one more would verify purge runs first and eviction is skipped (currently both tests exercise eviction or purge in isolation, not the interaction).

Peyton-Spencer (commented):
Nice deterministic coverage — `Date.now` spying is a clean way to drive the TTL/eviction edges. A couple of small follow-ups inline.

src/web/network.test.ts
Peyton-Spencer: Fixed in bec1468 by asserting both `Lab Network` and `ssid:lab` render.

src/web/session-auth.test.ts
Peyton-Spencer: Fixed in bec1468 with an inline comment documenting the exact `expiresAt` boundary.

src/web/session-auth.test.ts
Peyton-Spencer: Fixed in bec1468 with a capacity test that fills the store with expired sessions, creates one more, and verifies purge leaves only the new session.

Peyton-Spencer:
Triage note: this PR is otherwise small and focused, but it is not ready to merge while CI typecheck is failing. Current metadata shows format/test passing and typecheck failing; next step is to inspect and fix the `typecheck` job failure before requesting final review.

Peyton-Spencer:
Fixed the typecheck failure in `cd39b55` by adding required `trustedAt` timestamps to the `TrustedNetwork` fixtures in `src/web/network.test.ts`.

Verification:
- `bun run typecheck`
- `bun test src/web/network.test.ts`

PR #622, opened 2026-05-02
fix(security): cap execution broker payloads

Peyton-Spencer:
## Summary

- Cap split-execution broker request files at 64 KiB before JSON parsing.
- Cap captured stdout/stderr at 1 MiB per stream and stop the exec process when exceeded.
- Add regression coverage for oversized broker requests and output streams.

## Security Impact

The execution broker processes IPC request files and subprocess output from agent-controlled execution flows. Without local caps, a compromised or buggy agent could force the orchestrator to allocate excessive memory by writing a huge request file or producing unbounded stdout/stderr.

## Tests

- `bun test src/backends/local-backend.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **Tests**
  * Added execution broker parsing test suite validating size limit constraints and error handling.

* **Bug Fixes**
  * Implemented output size limits for execution broker to prevent excessive resource consumption.
  * Enhanced error handling with proper exit codes when output limits are exceeded.

### DIFF
--- a/src/backends/local-backend.test.ts
+++ b/src/backends/local-backend.test.ts
@@ -8,7 +8,10 @@ import {
   buildExecInvocationArgs,
   buildExecutionContainerArgs,
   buildVolumeMounts,
+  ExecBrokerOutputLimitError,
   LocalBackend,
+  parseExecRequest,
+  readExecBrokerText,
   resolveGitHubTokenForContainer,
 } from './local-backend.js';
 
@@ -797,6 +800,38 @@ describe('LocalBackend', () => {
         '-lc',
         'pwd',
       ]);
+    });
+  });
+
+  describe('execution broker parsing', () => {
+    const log = {
+      warn() {},
+    } as any;
+
+    it('rejects oversized execution request files before parsing JSON', () => {
+      const fixture = createFixture();
+      try {
+        fs.mkdirSync(fixture.ipcDir, { recursive: true });
+        const requestPath = path.join(fixture.ipcDir, 'oversized.json');
+        fs.writeFileSync(requestPath, 'x'.repeat(64 * 1024 + 1));
+
+        expect(parseExecRequest(requestPath, log)).toBeNull();
+      } finally {
+        fixture.cleanup();
+      }
+    });
+
+    it('limits execution broker stream capture', async () => {
+      const stream = new ReadableStream<Uint8Array>({
+        start(controller) {
+          controller.enqueue(new Uint8Array(1024 * 1024 + 1));
+          controller.close();
+        },
+      });
+
+      await expect(readExecBrokerText(stream, 'stdout')).rejects.toThrow(
+        ExecBrokerOutputLimitError,
+      );
     });
   });
 
--- a/src/backends/local-backend.test.ts
+++ b/src/backends/local-backend.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, describe, expect, it } from 'bun:test';
+import { afterEach, describe, expect, it, mock, spyOn } from 'bun:test';
 import fs from 'node:fs';
 import path from 'node:path';
 
@@ -13,6 +13,7 @@ import {
   parseExecRequest,
   readExecBrokerText,
   resolveGitHubTokenForContainer,
+  runExecBrokerRequest,
 } from './local-backend.js';
 
 function uniqueId(): string {
@@ -821,6 +822,18 @@ describe('LocalBackend', () => {
       }
     });
 
+    it('returns null when an execution request file cannot be statted', () => {
+      const fixture = createFixture();
+      try {
+        fs.mkdirSync(fixture.ipcDir, { recursive: true });
+        const requestPath = path.join(fixture.ipcDir, 'missing.json');
+
+        expect(parseExecRequest(requestPath, log)).toBeNull();
+      } finally {
+        fixture.cleanup();
+      }
+    });
+
     it('limits execution broker stream capture', async () => {
       const stream = new ReadableStream<Uint8Array>({
         start(controller) {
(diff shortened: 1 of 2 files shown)


src/backends/local-backend.ts
```diff
@@ -937,19 +985,25 @@ async function runExecBrokerRequest(
...
-  ]);
+  let stdout = '';
+  let stderr = '';
+  let exitCode = 125;
+  try {
+    [stdout, stderr, exitCode] = await Promise.all([
+      readExecBrokerText(proc.stdout, 'stdout'),
+      readExecBrokerText(proc.stderr, 'stderr'),
+      proc.exited,
+    ]);
```
Peyton-Spencer, line 1003: **[nit] Sibling reader can keep pulling after one stream throws**

When `readExecBrokerText(proc.stdout, ...)` rejects, `Promise.all` rejects immediately, but the `proc.stderr` reader inside the other concurrent `readExecBrokerText` call is still in its `while` loop. We rely on `proc.kill()` (SIGTERM) closing the stdio pipes so its `read()` returns `done`. That's usually fine, but if the child ignores SIGTERM (e.g., a wedged shell wrapper) we'd keep buffering up to another 1 MiB on the sibling stream before the `>` guard fires, and the orphan reader stays parked on `read()` until then.

Two small follow-ups would tighten this:
1. `await reader.cancel()` (or capture both readers and `cancel()` the loser when one throws) before/alongside `proc.kill()`.
2. Escalate to SIGKILL if `proc.exited` doesn't resolve within a short timeout.

Not a blocker for this PR — the cap still holds — but worth a follow-up issue.

src/backends/local-backend.ts
```diff
@@ -849,10 +851,56 @@ async function spawnExecutionContainer(
...
+  } finally {
+    reader.releaseLock();
+  }
+}
+
+export function parseExecRequest(
   requestPath: string,
   log: typeof logger,
 ): ExecRequest | null {
+  const { size } = fs.statSync(requestPath);
```
Peyton-Spencer: **[nit] `fs.statSync` will throw if the request file disappears between rename and parse**

The broker loop renames `requestPath` → `processingPath` (line 1050) and then calls `parseExecRequest(processingPath, log)`. If the file is gone (e.g., FS quirk, manual cleanup, container teardown), `statSync` throws, which is caught by the outer `try/catch` in `startExecBroker` and surfaces as a synthetic `Malformed execution request` response — that's probably the desired behavior, but the warning log path in `parseExecRequest` is bypassed in that case.

Minor: consider `try { stat } catch { return null; }` so the size-check failure mode mirrors the malformed-JSON failure mode (logged warning + null), or add a comment that this throw is intentionally caught upstream.

src/backends/local-backend.ts
```diff
@@ -849,10 +851,56 @@ async function spawnExecutionContainer(
...
+
+  try {
+    while (true) {
+      const { value, done } = await reader.read();
+      if (done) break;
+      bytes += value.byteLength;
+      if (bytes > MAX_EXEC_BROKER_OUTPUT_BYTES) {
+        throw new ExecBrokerOutputLimitError(label);
+      }
+      text += decoder.decode(value, { stream: true });
```
Peyton-Spencer, line 882: **[nit] Decoder text accumulates one chunk past the threshold check is fine, but the test only exercises a single-chunk over-limit case**

The guard correctly throws *before* `text += decoder.decode(...)` on the over-limit chunk, so the buffered `text` never crosses 1 MiB. Good.

For regression coverage, consider an additional test that streams *many* small chunks summing to just over the cap (e.g., 1024 chunks of 1 KiB + 1 byte) — that's the realistic agent-output shape and exercises the cumulative `bytes += value.byteLength` path rather than the single-large-chunk path. The current test would still pass even if the implementation had checked `value.byteLength` instead of cumulative `bytes`.

src/backends/local-backend.ts
```diff
@@ -937,19 +985,25 @@ async function runExecBrokerRequest(
...
+      readExecBrokerText(proc.stderr, 'stderr'),
+      proc.exited,
+    ]);
+  } catch (err) {
+    proc.kill();
+    await proc.exited.catch(() => undefined);
+    writeExecBrokerResponse(responseDir, request.id, {
+      stdout: '',
+      stderr: err instanceof Error ? err.message : 'Execution broker failed',
+      exitCode,
```
Peyton-Spencer, line 1020: **[nit] `exitCode` defaults to 125 only inside the catch block**

The `let exitCode = 125;` initializer is dead in the success path (overwritten by `Promise.all` destructuring). It's only used when the `try` throws *before* `proc.exited` resolves into a real value. Reads cleanly enough as-is; just flagging that the `125` fallback is the broker's invariant for "we synthesized this response because the child never returned a real exit". Worth a one-line comment for future readers, IMO.

src/backends/local-backend.test.ts
```diff
@@ -800,6 +803,38 @@ describe('LocalBackend', () => {
...
+        const requestPath = path.join(fixture.ipcDir, 'oversized.json');
+        fs.writeFileSync(requestPath, 'x'.repeat(64 * 1024 + 1));
+
+        expect(parseExecRequest(requestPath, log)).toBeNull();
+      } finally {
+        fixture.cleanup();
+      }
+    });
+
+    it('limits execution broker stream capture', async () => {
```
Peyton-Spencer, line 837: [nit] Consider asserting that the synthetic over-limit response is also surfaced through `runExecBrokerRequest`'s response writer (i.e., a `.stderr` file containing the limit message and `.exitcode` of `125`). The unit-level coverage on the limit error is good; an integration-style assertion would lock in the broker contract for a misbehaving agent.

Peyton-Spencer (commented):
Security review — looks solid overall. The two caps are well-placed (size pre-check before `JSON.parse`, byte-count guard in the streaming reader), tests assert both boundary behaviors, and the error path now returns a synthetic `125` response instead of letting the broker loop drop the request silently. Verified locally — `bun test src/backends/local-backend.test.ts` → 22 pass / 0 fail.

A few non-blocking observations left as line comments. Approving in spirit; happy to stamp once you take a look.

src/backends/local-backend.ts
Peyton-Spencer: Added SIGKILL escalation if the child does not exit after SIGTERM. I left explicit reader cancellation out of this small follow-up because stream closure after forced termination keeps the cap invariant without threading reader handles through the broker.

src/backends/local-backend.ts
Peyton-Spencer: Fixed. parseExecRequest now catches stat failures, logs a warning, and returns null consistently with other malformed request paths.

src/backends/local-backend.ts
Peyton-Spencer: Fixed. Added a many-small-chunks regression test that crosses the cumulative output cap by one byte.

src/backends/local-backend.ts
Peyton-Spencer: Fixed. Added a one-line comment documenting 125 as the synthetic broker response fallback before the child returns an exit code.

src/backends/local-backend.test.ts
Peyton-Spencer: Fixed. Added broker-level coverage that an over-limit output writes empty stdout, the limit error to stderr, and exitcode 125.

Peyton-Spencer (commented):
Verified the follow-up commit `4f3915d` locally — `bun test src/backends/local-backend.test.ts` → 25 pass / 0 fail, `bun run tsc --noEmit` clean.

All five threads addressed:
- `parseExecRequest` stat failures now log + return null consistently with malformed-JSON path.
- SIGKILL escalation after a 1s grace window covers the wedged-child case (functionally subsumes the sibling-reader cancellation concern — kill closes the pipes).
- `exitCode = 125` initializer documented.
- New cumulative many-small-chunks test locks in the cumulative byte counter.
- New broker response-file test asserts the `125` + cap-error stderr surface.

LGTM — can't formally approve as I share the PR author identity, but ready to merge from my side.

PR #629, opened 2026-05-03
feat: load declarative agent topology

### LINKED ISSUE #57: feat: declarative agent topology config file
## Summary

Agent configuration is currently stored in SQLite `registered_groups` table, managed imperatively through IPC `register_group` commands. As the system grows in complexity (6 backends, 4 channels, heartbeats, server folders, etc.), a declarative config format would make it easier to understand and reproduce the full agent topology.

## Proposed Design

A YAML or TOML file (e.g., `agents.yaml`) that declares the full topology:

```yaml
agents:
  main:
    backend: apple-container
    channels:
      - jid: "123@s.whatsapp.net"
        trigger: "@Omni"
    heartbeat:
      enabled: true
      interval: "1800000"
  omniclaw-discord:
    backend: sprites
    channels:
      - jid: "dc:123456"
        trigger: "@PeytonOmni"
        discordGuildId: "789"
```

SQLite remains the runtime state store (messages, sessions, task runs), but agent topology comes from the config file. Changes to the file are applied on startup or reload.

## Why

- Makes the full system topology visible in one place
- Easier to version control and review changes
- Reproducible setup without replaying IPC commands
- Aligns with the "small enough to understand" philosophy

## References

- See `docs/ROADMAP.md` (Long-term: Declarative Agent Configuration)

Peyton-Spencer:
## Summary

- Add `agents.yaml` topology loading at startup, controlled by `AGENTS_CONFIG_PATH`.
- Validate declarative agent/channel registrations with Zod and upsert them through the existing `registerGroup` path.
- Document the config format and add parser/application tests.

## Tests

- `bun test src/agent-topology.test.ts`
- `bun run typecheck`
- `bun test`

Fixes #57

## Release Notes

* **New Features**
  * Added support for declarative agent topology via `agents.yaml` configuration file
  * Added `AGENTS_CONFIG_PATH` environment variable to specify custom topology file location (defaults to `agents.yaml`)

* **Documentation**
  * Added declarative topology documentation to README with sample configuration structure
  * Updated `.env.example` with new configuration settings

* **Tests**
  * Added comprehensive test coverage for agent topology loading and application

### DIFF
--- a/src/agent-topology.test.ts
+++ b/src/agent-topology.test.ts
@@ -1,0 +1,156 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { afterEach, describe, expect, it } from 'vitest';
+
+import {
+  applyDeclarativeAgentTopology,
+  loadDeclarativeAgentTopology,
+} from './agent-topology.js';
+import type { RegisteredGroup } from './types.js';
+
+const tempDirs: string[] = [];
+
+afterEach(() => {
+  for (const dir of tempDirs.splice(0)) {
+    fs.rmSync(dir, { recursive: true, force: true });
+  }
+});
+
+function writeTopology(source: string): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'omniclaw-topology-'));
+  tempDirs.push(dir);
+  const filePath = path.join(dir, 'agents.yaml');
+  fs.writeFileSync(filePath, source);
+  return filePath;
+}
+
+describe('loadDeclarativeAgentTopology', () => {
+  it('returns no registrations when the topology file is absent', () => {
+    const filePath = path.join(os.tmpdir(), 'missing-agents.yaml');
+
+    expect(loadDeclarativeAgentTopology(filePath)).toEqual([]);
+  });
+
+  it('maps agents.yaml channels to registered groups', () => {
+    const filePath = writeTopology(`
+agents:
+  omniclaw-discord:
+    name: OCPeyton
+    backend: docker
+    agentRuntime: opencode
+    description: Infrastructure agent
+    serverFolder: servers/omniaura
+    channels:
+      - jid: "dc:123456"
+        name: omniclaw
+        trigger: "@OCPeyton"
+        requiresTrigger: true
+        discordGuildId: "789"
+        channelFolder: servers/omniaura/omniclaw
+`);
+
+    const registrations = loadDeclarativeAgentTopology(filePath);
+
+    expect(registrations).toHaveLength(1);
+    expect(registrations[0].jid).toBe('dc:123456');
+    expect(registrations[0].group).toMatchObject({
+      name: 'omniclaw',
+      folder: 'omniclaw-discord',
+      trigger: '@OCPeyton',
+      backend: 'docker',
+      agentRuntime: 'opencode',
+      description: 'Infrastructure agent',
+      discordGuildId: '789',
+      serverFolder: 'servers/omniaura',
+      channelFolder: 'servers/omniaura/omniclaw',
+    });
+    expect(registrations[0].group.added_at).toMatch(/^\d{4}-\d{2}-\d{2}T/);
+  });
+
+  it('preserves existing added_at timestamps for stable upserts', () => {
+    const filePath = writeTopology(`
+agents:
+  main:
+    channels:
+      - jid: "123@s.whatsapp.net"
+        trigger: "@Omni"
+`);
+    const existing: Record<string, RegisteredGroup> = {
+      '123@s.whatsapp.net': {
+        name: 'Old name',
+        folder: 'main',
+        trigger: '@Old',
+        added_at: '2026-01-01T00:00:00.000Z',
+      },
+    };
+
+    const registrations = loadDeclarativeAgentTopology(filePath, existing);
+
+    expect(registrations[0].group.added_at).toBe('2026-01-01T00:00:00.000Z');
+    expect(registrations[0].group.trigger).toBe('@Omni');
+  });
+
+  it('rejects unsafe folder names', () => {
+    const filePath = writeTopology(`
+agents:
+  ../main:
+    channels:
+      - jid: "123@s.whatsapp.net"
+        trigger: "@Omni"
+`);
+
+    expect(() => loadDeclarativeAgentTopology(filePath)).toThrow(
+      /Invalid key in record/,
+    );
+  });
+
+  it('rejects duplicate channel jids', () => {
+    const filePath = writeTopology(`
+agents:
+  main:
+    channels:
+      - jid: "dc:123"
+        trigger: "@Omni"
+  other:
+    channels:
+      - jid: "dc:123"
+        trigger: "@Other"
+`);
+
+    expect(() => loadDeclarativeAgentTopology(filePath)).toThrow(
+      /duplicate channel jid dc:123/,
+    );
+  });
+});
+
+describe('applyDeclarativeAgentTopology', () => {
+  it('registers each loaded topology channel', () => {
+    const filePath = writeTopology(`
+agents:
+  main:
+    name: Omni
+    channels:
+      - jid: "123@s.whatsapp.net"
+        trigger: "@Omni"
+      - jid: "dc:456"
+        trigger: "@Omni"
+`);
+    const registered: Array<{ jid: string; group: RegisteredGroup }> = [];
+
+    const count = applyDeclarativeAgentTopology({
+      filePath,
+      existingGroups: {},
+      registerGroup: (jid, group) => registered.push({ jid, group }),
+    });
+
+    expect(count).toBe(2);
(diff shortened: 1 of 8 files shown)


Peyton-Spencer (commented):
Review for #57 declarative topology loader. Overall direction looks solid — narrow Zod schema, file-existence-as-opt-out, and upserting through the existing `registerGroup` path so SQLite stays the runtime store. A few observations and follow-ups inline; the heartbeat gap is the main one I'd want addressed before merge.

Not blocking: camelCase/snake_case duplication, deletion semantics, and `containerConfig` validation can land as follow-ups if you'd rather keep this PR tight.

src/agent-topology.ts
```diff
@@ -0,0 +1,172 @@
...
+    agentRuntime: agentRuntimeSchema.optional(),
+    agent_runtime: agentRuntimeSchema.optional(),
+    description: z.string().min(1).optional(),
+    containerConfig: containerConfigSchema.optional(),
+    container_config: containerConfigSchema.optional(),
+    serverFolder: z.string().min(1).optional(),
+    server_folder: z.string().min(1).optional(),
+    agentContextFolder: z.string().min(1).optional(),
+    agent_context_folder: z.string().min(1).optional(),
+    heartbeat: z.object({}).passthrough().optional(),
```
Peyton-Spencer: **Heartbeat accepted but silently dropped.** Schema parses `heartbeat: z.object({}).passthrough().optional()` here, but the registration loop on lines 107–148 never reads `agent.heartbeat`, and `RegisteredGroup` (src/types.ts:62) has no heartbeat field. Issue #57's example specifically shows:

```yaml
heartbeat:
  enabled: true
  interval: "1800000"
```

So users following the issue will configure heartbeat blocks that have zero effect. Two clean options:
1. Drop `heartbeat` from the schema in this PR and add it as a tracked follow-up issue — `.strict()` will then reject the field with a clear error.
2. Wire it through `RegisteredGroup` and the task scheduler now.

Either is fine; silently accepting-and-discarding is the worst option.

src/agent-topology.ts
```diff
@@ -0,0 +1,172 @@
...
+    server_folder: z.string().min(1).optional(),
+    channelFolder: z.string().min(1).optional(),
+    channel_folder: z.string().min(1).optional(),
+    categoryFolder: z.string().min(1).optional(),
+    category_folder: z.string().min(1).optional(),
+    autoRespondToQuestions: z.boolean().optional(),
+    auto_respond_to_questions: z.boolean().optional(),
+    autoRespondKeywords: z.array(z.string().min(1)).optional(),
+    auto_respond_keywords: z.array(z.string().min(1)).optional(),
+  })
```
Peyton-Spencer, line 49: **Dual camelCase + snake_case surface.** This (and `agentSchema`) accepts both `discordGuildId` and `discord_guild_id`, etc., for ~7 fields. The README only documents camelCase. Picking up the first present alias (`channel.discordGuildId || channel.discord_guild_id`) silently hides cases where both are set with different values.

Suggestion: pick one (camelCase, matching `RegisteredGroup`) and reject the rest via `.strict()`. If you want to keep both, at minimum:
- document snake_case in the README,
- add a test that asserts precedence when both are set,
- consider rejecting when both keys are present with different values.

src/agent-topology.ts
```diff
@@ -0,0 +1,172 @@
...
+const folderSchema = z
+  .string()
+  .regex(/^[a-z0-9][a-z0-9_-]*$/i, 'must be a safe folder name');
+
+const backendSchema = z.enum(['apple-container', 'docker']);
+const agentRuntimeSchema = z.enum(['claude-agent-sdk', 'opencode', 'codex']);
+const containerConfigSchema = z
+  .object({})
+  .passthrough()
+  .transform((value) => value as ContainerConfig);
```
Peyton-Spencer, line 34: **`containerConfig` bypasses validation.** `z.object({}).passthrough().transform(v => v as ContainerConfig)` accepts any shape and unsafely casts. That defeats the Zod safety the rest of the schema provides. Worth a real schema for `ContainerConfig` (memory limits, mounts, etc.) — even a partial one — so a typo like `memmory: "4G"` doesn't silently get persisted into the DB.

src/agent-topology.ts
```diff
@@ -0,0 +1,172 @@
+import fs from 'fs';
+import { parseDocument } from 'yaml';
+import { z } from 'zod';
+
+import { logger } from './logger.js';
+import type { BackendType, ContainerConfig, RegisteredGroup } from './types.js';
+
+const folderSchema = z
+  .string()
+  .regex(/^[a-z0-9][a-z0-9_-]*$/i, 'must be a safe folder name');
```
Peyton-Spencer: **Folder regex is case-insensitive (`/i`).** Folder names map to filesystem paths and container IDs. Allowing `Main`/`main`/`MAIN` invites collisions on case-insensitive filesystems (macOS default) and inconsistent directory naming. Consider dropping the `/i` so folders must be lowercase.

src/agent-topology.ts
```diff
@@ -0,0 +1,172 @@
...
+export function applyDeclarativeAgentTopology(options: {
+  filePath: string;
+  existingGroups: Record<string, RegisteredGroup>;
+  registerGroup: (jid: string, group: RegisteredGroup) => void;
+}): number {
+  const registrations = loadDeclarativeAgentTopology(
+    options.filePath,
+    options.existingGroups,
+  );
+  for (const registration of registrations) {
```
Peyton-Spencer, line 170: **Upsert-only — no awareness of removed entries.** The PR description and README note this as expected, but issue #57's stated goal of "Reproducible setup without replaying IPC commands" implies the YAML is the source of truth. Removing a channel from `agents.yaml` leaves it live in the DB indefinitely with no log signal.

Minimum: log a startup warning when `existingGroups` contains JIDs not present in the loaded topology. That gives operators a chance to notice drift without requiring a destructive sync. A full reconcile (delete-on-absent) can come later behind a flag.

src/config.ts
```diff
@@ -161,6 +161,10 @@ const configSchema = z
       optionalTrimmedString,
       z.string().optional(),
     ),
+    AGENTS_CONFIG_PATH: z.preprocess(
+      optionalTrimmedString,
+      z.string().default('agents.yaml'),
+    ),
```
Peyton-Spencer, line 167: **No explicit disable form.** `AGENTS_CONFIG_PATH` defaults to `'agents.yaml'` and `optionalTrimmedString` collapses empty strings back to the default. The only "off switch" is not creating the file. Probably fine, but worth a README note that the file is opt-in by existence (some users will assume an env-var control).

src/agent-topology.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(count).toBe(2);
+    expect(registered.map((entry) => entry.jid)).toEqual([
+      '123@s.whatsapp.net',
+      'dc:456',
+    ]);
+    expect(registered.every((entry) => entry.group.folder === 'main')).toBe(
+      true,
+    );
+  });
+});
```
Peyton-Spencer, line 312: **Coverage gaps worth a follow-up:**
- `containerConfig` parsing (currently uncovered).
- Both-cases-set precedence when `discordGuildId` and `discord_guild_id` are both present.
- Unknown top-level keys at agent level — would confirm `.strict()` rejects typos like `chanels:`.
- YAML parse errors surface with the `Invalid agent topology YAML:` prefix.
- `applyDeclarativeAgentTopology` returns 0 (and doesn't throw / doesn't call `registerGroup`) when the file is missing.

Peyton-Spencer:
Addressed Clayton's review in `66c6300`.

Changes:
- Removed `heartbeat` from the accepted schema so it fails loudly until scheduler wiring exists.
- Switched declarative topology to camelCase-only keys and kept `.strict()` rejection for aliases/typos.
- Made folder names lowercase-only to avoid filesystem/container naming collisions.
- Added partial `containerConfig` validation for known fields instead of passthrough casting.
- Added a startup warning when existing DB registrations are absent from a loaded topology file.
- Expanded tests for missing files, heartbeat rejection, snake_case rejection, unknown keys, containerConfig typos, YAML parse errors, and drift behavior.

Verification:
- `bun test src/agent-topology.test.ts`
- `bun run typecheck`
- `bun test`

Peyton-Spencer (commented):
Confirmed all review items in `66c6300`:

- `heartbeat` removed from schema; `.strict()` now rejects it with a clear error (test added).
- snake_case aliases dropped across agent + channel schemas (test added).
- `containerConfig` now has a real Zod shape with `.strict()` — typos like `memmory` are caught (test added).
- Folder regex no longer case-insensitive (test added).
- Drift detection: `applyDeclarativeAgentTopology` warns on `existingGroups` JIDs absent from the YAML.
- Test coverage filled: containerConfig happy-path, unknown keys (top-level + nested), YAML parse error prefix, file-absent returns 0 without calling registerGroup.

Nice turnaround. LGTM.

PR #632, opened 2026-05-03
fix(discord): sync slash commands for codex bots

Peyton-Spencer:
## Summary

- Sync Discord slash commands from canonical channel subscriptions instead of only the legacy `registeredGroups` map.
- Allows Codex-backed Discord bots to register slash commands even when Claude is the primary registered group for the channel.
- Keeps legacy `registeredGroups` fallback for older single-group registrations.

## Verification

- `bun run typecheck`
- `bun test ./src/channels/discord.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Discord slash command synchronization to build command groups from live channel subscriptions and agent data, deduplicate entries, and correctly scope commands per guild while preserving legacy single-bot behavior.
  
* **Tests**
  * Added tests covering Discord slash command refresh flows, including multi-bot sync, subscription-derived command folders, and legacy single-bot retention.

### DIFF
--- a/container/agent-runner/src/__tests__/codex-runtime.test.ts
+++ b/container/agent-runner/src/__tests__/codex-runtime.test.ts
@@ -1,6 +1,4 @@
 import { afterEach, describe, expect, it } from 'bun:test';
-import fs from 'fs';
-import path from 'path';
 
 import {
   buildCodexAppServerArgs,
@@ -8,15 +6,12 @@ import {
   buildCodexThreadResumeParams,
   buildCodexThreadStartParams,
   buildCodexTurnStartParams,
-  expandSlashCommandPrompt,
   extractAssistantTextFromItem,
   extractTextFromCodexContent,
   isRecoverableThreadResumeError,
 } from '../codex-runtime.js';
 
 const ORIGINAL_ENV = { ...process.env };
-const TEST_COMMANDS_DIR = path.join(process.cwd(), '.tmp-codex-slash-commands');
-const TEST_COMMAND_FILE = path.join(TEST_COMMANDS_DIR, 'codexslashtest.md');
 
 function resetEnv(): void {
   for (const key of Object.keys(process.env)) {
@@ -30,11 +25,6 @@ function resetEnv(): void {
 }
 
 afterEach(() => {
-  try {
-    fs.rmSync(TEST_COMMANDS_DIR, { recursive: true, force: true });
-  } catch {
-    /* ignore */
-  }
   resetEnv();
 });
 
@@ -54,7 +44,6 @@ describe('buildCodexEnv', () => {
 
   it('preserves github auth for Codex shell commands', () => {
     process.env.GITHUB_TOKEN = 'gh-token';
-    delete process.env.GH_TOKEN;
 
     const env = buildCodexEnv({} as any);
 
@@ -180,27 +169,6 @@ describe('buildCodexTurnStartParams', () => {
         networkAccess: 'restricted',
       },
     });
-  });
-});
-
-describe('expandSlashCommandPrompt', () => {
-  it('expands Claude-style slash command files before sending to Codex', () => {
-    process.env.OMNICLAW_SLASH_COMMAND_ROOTS = TEST_COMMANDS_DIR;
-    fs.mkdirSync(TEST_COMMANDS_DIR, { recursive: true });
-    fs.writeFileSync(
-      TEST_COMMAND_FILE,
-      'Run the shared workflow with: $ARGUMENTS\n',
-    );
-
-    expect(expandSlashCommandPrompt('/codexslashtest repo cleanup')).toBe(
-      'Run the shared workflow with: repo cleanup',
-    );
-  });
-
-  it('leaves unknown slash commands unchanged for native Codex handling', () => {
-    expect(expandSlashCommandPrompt('/__test_codex_missing arg')).toBe(
-      '/__test_codex_missing arg',
-    );
   });
 });
 
--- a/container/agent-runner/src/__tests__/codex-runtime.test.ts
+++ b/container/agent-runner/src/__tests__/codex-runtime.test.ts
@@ -1,4 +1,6 @@
 import { afterEach, describe, expect, it } from 'bun:test';
+import fs from 'fs';
+import path from 'path';
 
 import {
   buildCodexAppServerArgs,
@@ -6,12 +8,15 @@ import {
   buildCodexThreadResumeParams,
   buildCodexThreadStartParams,
   buildCodexTurnStartParams,
+  expandSlashCommandPrompt,
   extractAssistantTextFromItem,
   extractTextFromCodexContent,
   isRecoverableThreadResumeError,
 } from '../codex-runtime.js';
 
 const ORIGINAL_ENV = { ...process.env };
+const TEST_COMMANDS_DIR = path.join(process.cwd(), '.tmp-codex-slash-commands');
+const TEST_COMMAND_FILE = path.join(TEST_COMMANDS_DIR, 'codexslashtest.md');
 
 function resetEnv(): void {
   for (const key of Object.keys(process.env)) {
@@ -25,6 +30,11 @@ function resetEnv(): void {
 }
 
 afterEach(() => {
+  try {
+    fs.rmSync(TEST_COMMANDS_DIR, { recursive: true, force: true });
+  } catch {
+    /* ignore */
+  }
   resetEnv();
 });
 
@@ -44,6 +54,7 @@ describe('buildCodexEnv', () => {
 
   it('preserves github auth for Codex shell commands', () => {
     process.env.GITHUB_TOKEN = 'gh-token';
+    delete process.env.GH_TOKEN;
 
     const env = buildCodexEnv({} as any);
 
@@ -169,6 +180,27 @@ describe('buildCodexTurnStartParams', () => {
         networkAccess: 'restricted',
       },
     });
+  });
+});
+
+describe('expandSlashCommandPrompt', () => {
+  it('expands Claude-style slash command files before sending to Codex', () => {
+    process.env.OMNICLAW_SLASH_COMMAND_ROOTS = TEST_COMMANDS_DIR;
+    fs.mkdirSync(TEST_COMMANDS_DIR, { recursive: true });
+    fs.writeFileSync(
+      TEST_COMMAND_FILE,
+      'Run the shared workflow with: $ARGUMENTS\n',
+    );
+
+    expect(expandSlashCommandPrompt('/codexslashtest repo cleanup')).toBe(
+      'Run the shared workflow with: repo cleanup',
+    );
+  });
+
+  it('leaves unknown slash commands unchanged for native Codex handling', () => {
+    expect(expandSlashCommandPrompt('/__test_codex_missing arg')).toBe(
+      '/__test_codex_missing arg',
+    );
   });
 });
 
--- a/container/agent-runner/src/codex-runtime.ts
+++ b/container/agent-runner/src/codex-runtime.ts
@@ -88,6 +88,9 @@ const OUTPUT_END_MARKER = '---OMNICLAW_OUTPUT_END---';
 const IPC_POLL_MS = 500;
 const APP_SERVER_REQUEST_TIMEOUT_MS = 30_000;
 const TURN_TIMEOUT_MS = 1_800_000; // 30 min
+const SLASH_COMMAND_FILE_MAX_BYTES = 64 * 1024;
+const SLASH_COMMAND_NAME_RE =
+  /^[A-Za-z0-9][A-Za-z0-9_-]*(?::[A-Za-z0-9][A-Za-z0-9_-]*)?$/;
 
 const RECOVERABLE_THREAD_RESUME_ERROR_SNIPPETS = [
   'not found',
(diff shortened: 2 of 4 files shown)


container/agent-runner/src/codex-runtime.ts
```diff
@@ -876,6 +879,136 @@ function waitForIpcMessage(): Promise<string | null> {
...
+  if (!SLASH_COMMAND_NAME_RE.test(name)) return null;
+
+  const trailingPrompt =
+    firstLineEnd === -1 ? '' : trimmed.slice(firstLineEnd).trim();
+  const args = [inlineArgs.trim(), trailingPrompt].filter(Boolean).join('\n\n');
+  return { name, arguments: args };
+}
+
+function slashCommandSearchRoots(): string[] {
+  const configuredRoots = process.env.OMNICLAW_SLASH_COMMAND_ROOTS?.split('\n')
```
Peyton-Spencer: [P2] **`OMNICLAW_SLASH_COMMAND_ROOTS` uses newline as the only separator.**

Newline-delimited env vars are awkward to set from a shell or a typical container env file (you basically need `$'...\n...'` or YAML literal blocks). The conventional shapes for path lists in env vars are `:` (PATH-like) or `;`.

Also, since this is undocumented anywhere outside this function, future readers won't know the contract. Could you either:

1. Switch to `:` (or `path.delimiter`) as the separator, _or_
2. Accept both newline _and_ `:` and document the precedence in a short JSDoc on `slashCommandSearchRoots()` plus a line in the PR description / CLAUDE.md?

Not blocking — just want to make sure ops can actually drive this knob without surprises.

container/agent-runner/src/codex-runtime.ts
```diff
@@ -876,6 +879,136 @@ function waitForIpcMessage(): Promise<string | null> {
...
+    '/workspace/category/.claude/commands',
+    '/workspace/server/.claude/commands',
+    '/workspace/agent/.claude/commands',
+    '/home/bun/.claude/commands',
+    '/workspace/group/.codex/prompts',
+    '/workspace/category/.codex/prompts',
+    '/workspace/server/.codex/prompts',
+    '/workspace/agent/.codex/prompts',
+    '/home/bun/.codex/prompts',
+  ];
```
Peyton-Spencer: [P3] **Search root precedence puts `.claude/commands` ahead of `.codex/prompts`.**

Intentional? The five `.claude/commands` paths all win over the five `.codex/prompts` paths with the same scope. That makes sense if you want "Claude commands work everywhere" as the default, but it also means a Codex-specific override placed in `.codex/prompts/foo.md` will be **shadowed** by any `.claude/commands/foo.md` of the same name.

For the codex runtime specifically, I'd argue the more intuitive ordering is _scope-major_ (`group → category → server → agent → home`) and _kind-minor_ with `.codex/prompts` winning over `.claude/commands` within each scope, since this is the Codex backend. Curious what shape you had in mind.

container/agent-runner/src/codex-runtime.ts
```diff
@@ -876,6 +879,136 @@ function waitForIpcMessage(): Promise<string | null> {
...
+  const args = invocation.arguments;
+  const expanded = commandPrompt
+    .replaceAll('$ARGUMENTS', args)
+    .replaceAll('{{ARGUMENTS}}', args);
+
+  if (expanded === commandPrompt && args) {
+    return `${expanded}\n\nArguments:\n${args}`;
+  }
+  return expanded;
+}
```
Peyton-Spencer: [P3] **Implicit `Arguments:` fallback may surprise users.**

If a command file does **not** contain `$ARGUMENTS` / `{{ARGUMENTS}}` and the user supplied trailing args, you append:

```
<command body>

Arguments:
<args>
```

This is a nice ergonomic touch, but it diverges from Claude's slash-command semantics where unused args are silently dropped. Worth either:

- Documenting this fallback in the PR description / a doc note, or
- Dropping it for parity with Claude.

If you keep it, consider only triggering when args are non-trivial (e.g. >0 chars after trim — already handled — and maybe gating behind a flag). Not blocking.

container/agent-runner/src/codex-runtime.ts
```diff
@@ -1045,6 +1178,7 @@ export async function runCodexRuntime(
         break;
       }
 
+      prompt = expandSlashCommandPrompt(prompt);
```
Peyton-Spencer: [P2] **Slash expansion runs _after_ pending IPC messages are appended to the prompt.**

At line 1167 above (outside this PR's diff), `prompt += '\n' + formatIpcMessages(pendingResult.messages);` mutates the initial prompt to look like:

```
/foo bar baz
<ipc message 1>
<ipc message 2>
```

`parseSlashCommandInvocation` only inspects the **first line** for the command, so the slash command is still detected — but everything after the first line ends up in `trailingPrompt`, gets joined into `invocation.arguments`, and is then injected wherever `$ARGUMENTS` appears in the command file (or as an `Arguments:` trailer if no placeholder exists).

That means a queued IPC message can silently end up _inside_ a slash-command-expanded prompt template, possibly under a different role/heading than intended.

A couple of options:

1. Expand the slash command **before** appending IPC messages, then re-attach IPC content as plain trailing text (so it stays out of `$ARGUMENTS`).
2. Restrict slash expansion to prompts where the entire string (post-trim) starts with `/` and has no embedded newlines — i.e. only treat single-line invocations as slash commands.

Option 1 matches Claude's behavior best.

container/agent-runner/src/codex-runtime.ts
```diff
@@ -876,6 +879,136 @@ function waitForIpcMessage(): Promise<string | null> {
...
+    for (const candidate of slashCommandCandidateNames(name)) {
+      const paths = [
+        path.resolve(absoluteRoot, `${candidate}.md`),
+        path.resolve(absoluteRoot, candidate, 'index.md'),
+      ];
+      for (const candidatePath of paths) {
+        if (
+          isPathWithin(candidatePath, absoluteRoot) &&
+          fs.existsSync(candidatePath) &&
+          fs.statSync(candidatePath).isFile()
```
Peyton-Spencer: [P3] **Minor TOCTOU between `existsSync` / `statSync` / `readFileSync`.**

`resolveSlashCommandFile` checks `fs.existsSync` + `fs.statSync(...).isFile()`, then `expandSlashCommandPrompt` re-runs `fs.statSync` (line 992) and finally `fs.readFileSync` (line 998). All on the same path. Inside the container with a read-only-ish mount this is fine in practice, but you can collapse it to a single `fs.readFileSync` wrapped in try/catch + a length check on the resulting string, which removes the race window entirely and is fewer syscalls.

Purely a cleanup suggestion — feel free to ignore.

container/agent-runner/src/codex-runtime.ts
```diff
@@ -876,6 +879,136 @@ function waitForIpcMessage(): Promise<string | null> {
...
+  const firstLine =
+    firstLineEnd === -1 ? trimmed : trimmed.slice(0, firstLineEnd);
+  const match =
+    /^\/([A-Za-z0-9][A-Za-z0-9_-]*(?::[A-Za-z0-9][A-Za-z0-9_-]*)?)(?:\s+([\s\S]*))?$/.exec(
+      firstLine,
+    );
+  if (!match) return null;
+
+  const [, name, inlineArgs = ''] = match;
+  if (!SLASH_COMMAND_NAME_RE.test(name)) return null;
```
Peyton-Spencer: [nit] `SLASH_COMMAND_NAME_RE` here is structurally a subset of the inline regex on lines 909-911 (the captured group `[A-Za-z0-9][A-Za-z0-9_-]*(?::[A-Za-z0-9][A-Za-z0-9_-]*)?` already enforces it). The second check is unreachable as written. Fine to keep as defense-in-depth, but worth either dropping it or adding a comment explaining the intent.

container/agent-runner/src/__tests__/codex-runtime.test.ts
```diff
@@ -44,6 +54,7 @@ describe('buildCodexEnv', () => {
 
   it('preserves github auth for Codex shell commands', () => {
     process.env.GITHUB_TOKEN = 'gh-token';
```
Peyton-Spencer, line 46: [P3] `delete process.env.GH_TOKEN;` looks unrelated to slash commands. Was this an existing test isolation bug surfacing in CI under a new env shape? If so, worth either:

- A short comment explaining why `GH_TOKEN` specifically needs scrubbing (so it doesn't get reverted later), or
- Moving the cleanup into `resetEnv()` so all tests benefit, not just this one.

Not blocking but it caught my eye since the rest of the diff is purely additive.

container/agent-runner/src/__tests__/codex-runtime.test.ts
```diff
@@ -1,17 +1,22 @@
...
   buildCodexThreadResumeParams,
   buildCodexThreadStartParams,
   buildCodexTurnStartParams,
+  expandSlashCommandPrompt,
   extractAssistantTextFromItem,
   extractTextFromCodexContent,
   isRecoverableThreadResumeError,
 } from '../codex-runtime.js';
 
 const ORIGINAL_ENV = { ...process.env };
```
Peyton-Spencer, line 14: [nit] `TEST_COMMANDS_DIR = path.join(process.cwd(), '.tmp-codex-slash-commands')` is fine for a single-file test, but if these tests ever run in parallel (`bun test` does in some configs) two workers could race on the same directory. Consider `path.join(os.tmpdir(), 'omniclaw-codex-slash-commands-' + crypto.randomUUID())` or similar for isolation. Low priority.

Peyton-Spencer (commented):
Solid first cut at Codex slash-command parity. CI is green and the behavior is well-isolated behind `expandSlashCommandPrompt`. A few observations below as file/line comments — most are minor; the IPC-drain interaction and the `OMNICLAW_SLASH_COMMAND_ROOTS` separator are worth a closer look.

None of these are blocking — happy to see this land.

Peyton-Spencer (commented):
Registration side looks good and is now backend-agnostic. One important gap re: "works for all three backends":

- **`/resume` and `/sessions` are Claude-only.** `handleSessionCommand` in `src/index.ts` hardcodes the path `DATA_DIR/sessions/<runtimeFolder>/.claude/projects/-workspace-group/*.jsonl`. For Codex (`agentRuntime: 'codex'`) and OpenCode bots this directory will never exist, so `/sessions` always returns "No sessions found" and `/resume` always returns "Session not found". The PR registers these system commands for those bots but they're effectively dead.
  - Two options: (a) filter `system: true` builtins out of `buildDiscordSlashCommandPayloads` when no group in the bot's slice uses `agentRuntime: 'claude-agent-sdk'`, or (b) add backend dispatch to `handleSessionCommand` (Codex thread state, OpenCode session store).

- **Test gap:** the new test only covers the `agentRuntime: 'codex'` registration case. An OpenCode-runtime variant (and the legacy fallback that CodeRabbit flagged) would lock the contract.

Non-blocking otherwise — the synthetic-message path for non-system flows correctly fan-outs through the normal pipeline, so prompt commands like `/mergemaster` will work for all three runtimes.

src/channels/discord.ts
```diff
@@ -472,6 +468,59 @@ export class DiscordChannel implements Channel {
...
+          (group) =>
+            group.discordBotId === this.botId && !!group.discordGuildId,
+        )
+      : [];
+    for (const group of legacyGroups) {
+      groups.set(
+        `${group.discordGuildId}:${group.folder}:legacy:${group.channelFolder || ''}`,
+        group,
+      );
+    }
```
Peyton-Spencer, line 529: [P2] System commands `/resume` and `/sessions` get registered for every bot here regardless of `agentRuntime`, but `handleSessionCommand` in `src/index.ts:340-347` only knows how to read Claude's `.claude/projects/-workspace-group/*.jsonl` session files. For a Codex or OpenCode bot, both commands always fail ("No sessions found").

Consider filtering `system: true` builtins out of `buildDiscordSlashCommandPayloads` when none of the bot's groups have `agentRuntime: 'claude-agent-sdk'`, OR adding a backend dispatch to `handleSessionCommand` so Codex/OpenCode get their own session listing/restore.

src/channels/discord.test.ts
```diff
@@ -230,6 +230,74 @@ describe('Discord download guards', () => {
...
+    };
+
+    await channel.refreshSlashCommands();
+
+    expect(setCommands).toHaveBeenCalledTimes(1);
+    const commands = setCommands.mock.calls[0]?.[0] as Array<{
+      name: string;
+    }>;
+    expect(commands.map((command) => command.name)).toContain('mergemaster');
+  });
```
Peyton-Spencer, line 339: [nit] Only the Codex registration path is exercised. Adding an `agentRuntime: 'opencode'` case (and the no-`discordBotId` legacy single-bot case CodeRabbit flagged) would prevent regression of the "works for all three backends" contract this PR is targeting.

Peyton-Spencer:
Requesting one fix before merge.

The new subscription-backed slash sync reconstructs `RegisteredGroup` with `serverFolder: agent.serverFolder`, `categoryFolder: sub.categoryFolder`, and `channelFolder: sub.channelFolder` directly. That misses the existing derived Discord context layer behavior used elsewhere in the router. In normal registrations, `setChannelSubscription()` does not persist `channelFolder`, so a Discord subscription with only `channelJid` + `discordGuildId` should still derive `servers/<guild>/channels/<channel>`. Without that, `buildDiscordSlashCommandPayloads()` will only see built-ins plus agent/server folders and will skip channel-scoped `discord-commands.json` files for Codex-backed subscriptions.

Please build these groups through the same `resolveContextLayers({ channelJid, discordGuildId, serverFolder, categoryFolder, channelFolder })` path used by `buildRegisteredGroupFromSubscription()` / `refreshRegisteredGroupsFromCanonicalState()`, and add a test where a subscription has no explicit `channelFolder` but a command file exists in the derived Discord channel folder.

Verified locally:

- `bun run typecheck`
- `bun test ./src/channels/discord.test.ts`

Peyton-Spencer:
Handled the requested feedback.

- Subscription-derived slash groups now use `resolveContextLayers()` so Discord channel folders are derived as `servers/<guild>/channels/<channel>` when `channelFolder` is not persisted.
- Legacy single-bot registered groups without `discordBotId` are preserved during slash sync.
- Added regressions for derived channel-scoped `discord-commands.json` discovery and legacy single-bot group fallback.

Verified:

- `bun test ./src/channels/discord.test.ts`
- `bun run typecheck`

Peyton-Spencer:
Mergemaster hold for the 2026-05-03 window.

This PR is green, but I am not treating it as merge-ready:

- It appears to overlap with #635, which is the current lead slash-command fix.
- The latest CodeRabbit review still flags a runtime execution-path gap for derived channel-folder commands.
- The PR has older review discussion around backend/system-command behavior that should be reconciled before landing any slash-command changes.

Requested next step: decide whether #632 is superseded by #635. If superseded, close it after confirming #635 covers the required behavior. If it remains needed, resolve the runtime-path comment and request a fresh review on the updated head.

Peyton-Spencer:
Mergemaster audit follow-up.

I am still holding this PR. #635 looks like the cleaner lead implementation, but #632 is not fully superseded yet.

Items in #632 that still matter:

- Tests for Codex bot subscription registration.
- Tests for derived Discord channel-folder command registration.
- Tests for legacy single-bot `registeredGroups` without `discordBotId`.
- Legacy `registeredGroups` merge behavior; #635 currently bypasses legacy groups when `slashCommandGroups` is provided.

Remaining blocker in #632 itself:

- The latest runtime execution-path gap still needs resolution or explicit deferral: `resolveGroupForChannel()` can use raw subscription folders while registration uses derived context layers, so a derived channel-folder command may register successfully but fail during slash execution.

Recommended queue decision: keep #632 open until #635 either absorbs the missing coverage/legacy behavior or explicitly documents that those behaviors are being deferred.

Peyton-Spencer:
Superseded by #635. Follow-up Codex slash-command test coverage is tracked in #639.

PR #637, opened 2026-05-03
feat: add SDLC Discord driver commands

Peyton-Spencer:
## Summary

Adds SDLC-oriented Discord slash flows for agent coordination:

- `/product-driver`
- `/issue-driver`
- `/research-driver`
- `/implement-driver`
- `/qa-driver`

Also adds per-agent `agents.yaml` control for Discord command availability:

```yaml
discordCommands:
  enabled:
    - product-driver
    - issue-driver
  disabled:
    - mergemaster
```

Omitted `discordCommands` keeps the current default: all built-in and workspace command-file flows are enabled. `enabled` acts as an allowlist when present, and `disabled` is applied after that.

## Notes

- Prompt shape follows the same lightweight slash-command pattern as the existing flows: explicit role, scoped objective, expected deliverables, and clear constraints.
- I used current Claude slash-command docs and public skill/security writeups as input, but did not vendor or execute third-party skill content. The security angle is intentional: marketplace prompt packs should be treated as untrusted and adapted manually.
- The command filter applies after built-ins and workspace `discord-commands.json` files are loaded, so it works for both built-in and custom flows.
- Adds a schema migration for persisted `registered_groups.discord_commands`.

## Verification

- `bun test src/db.test.ts src/discord-command-flows.test.ts src/agent-topology.test.ts src/migrations.test.ts src/db.migrations.test.ts`
- `bun run typecheck`
- `bun run format:check`
- `bun test`

## Summary by CodeRabbit

* **New Features**
  * Added five new Discord slash commands: `/product-driver`, `/issue-driver`, `/research-driver`, `/implement-driver`, and `/qa-driver`.
  * Introduced per-agent Discord command filtering, allowing administrators to enable or disable specific commands via configuration.

* **Documentation**
  * Updated README with comprehensive documentation on all available Discord slash commands and configuration instructions.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -39,6 +39,7 @@ interface RegisteredGroupRow {
   description: string | null;
   auto_respond_to_questions: number | null;
   auto_respond_keywords: string | null;
+  discord_commands: string | null;
 }
 
 /** Row type for agents table SELECT * queries */
@@ -148,6 +149,11 @@ function mapRowToRegisteredGroup(
       row.auto_respond_keywords,
       { jid: row.jid },
       'auto_respond_keywords',
+    ),
+    discordCommands: safeJsonParse<RegisteredGroup['discordCommands']>(
+      row.discord_commands,
+      { jid: row.jid },
+      'discord_commands',
     ),
   };
 }
@@ -1236,8 +1242,8 @@ export function getRegisteredGroup(
 /** Insert or replace a registered group entry. */
 export function setRegisteredGroup(jid: string, group: RegisteredGroup): void {
   db.query(
-    `INSERT OR REPLACE INTO registered_groups (jid, name, folder, trigger_pattern, added_at, container_config, requires_trigger, discord_bot_id, discord_guild_id, server_folder, backend, description, auto_respond_to_questions, auto_respond_keywords)
-     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+    `INSERT OR REPLACE INTO registered_groups (jid, name, folder, trigger_pattern, added_at, container_config, requires_trigger, discord_bot_id, discord_guild_id, server_folder, backend, description, auto_respond_to_questions, auto_respond_keywords, discord_commands)
+     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
   ).run(
     jid,
     group.name,
@@ -1255,6 +1261,7 @@ export function setRegisteredGroup(jid: string, group: RegisteredGroup): void {
     group.autoRespondKeywords
       ? JSON.stringify(group.autoRespondKeywords)
       : null,
+    group.discordCommands ? JSON.stringify(group.discordCommands) : null,
   );
 }
 
--- a/src/db.ts
+++ b/src/db.ts
@@ -3,6 +3,7 @@ import { randomUUID } from 'crypto';
 import fs from 'fs';
 import path from 'path';
 
+import { discordCommandsSchema } from './agent-topology.js';
 import { DATA_DIR, STORE_DIR } from './config.js';
 import { TrustStore } from './discovery/trust-store.js';
 import { logger } from './logger.js';
@@ -123,6 +124,23 @@ function normalizeAgentRuntime(
   return 'claude-agent-sdk';
 }
 
+function parseDiscordCommands(
+  value: string | null,
+  jid: string,
+): RegisteredGroup['discordCommands'] {
+  const parsed = safeJsonParse<unknown>(value, { jid }, 'discord_commands');
+  if (parsed === undefined) return undefined;
+  const result = discordCommandsSchema.safeParse(parsed);
+  if (!result.success) {
+    logger.warn(
+      { jid, issues: result.error.issues },
+      'Invalid discord_commands in DB, ignoring',
+    );
+    return undefined;
+  }
+  return result.data;
+}
+
 /** Map a database row to a RegisteredGroup object (without jid field) */
 function mapRowToRegisteredGroup(
   row: RegisteredGroupRow,
@@ -150,11 +168,7 @@ function mapRowToRegisteredGroup(
       { jid: row.jid },
       'auto_respond_keywords',
     ),
-    discordCommands: safeJsonParse<RegisteredGroup['discordCommands']>(
-      row.discord_commands,
-      { jid: row.jid },
-      'discord_commands',
-    ),
+    discordCommands: parseDiscordCommands(row.discord_commands, row.jid),
   };
 }
 
--- a/src/discord-command-flows.ts
+++ b/src/discord-command-flows.ts
@@ -67,6 +67,130 @@ const BUILTIN_COMMANDS: DiscordFlowDefinition[] = [
     ],
   },
   {
+    name: 'product-driver',
+    description: 'Drive product discovery and issue creation',
+    prompt:
+      'You are product driver for {{repo}} for the next {{duration_minutes}} minutes.\n\nGoal: {{goal}}\n\nYour job is to turn ambiguous product direction into concrete product progress. Start by reading the repo context, open issues, recent PRs, product docs, and relevant user conversations. De-duplicate before filing anything.\n\nExpected output:\n- identify the highest-leverage product gaps or opportunities\n- file clear issues with user problem, scope, acceptance criteria, and test notes when the repo needs tracking issues\n- direct small exploratory PRs only when code discovery is needed to clarify the product path\n- keep implementation delegated or narrowly exploratory; do not drift into broad feature work\n- leave a short queue update naming filed issues, delegated owners, blockers, and next decisions needed',
+    options: [
+      {
+        name: 'repo',
+        description: 'Repository or product area to drive',
+        type: 'string',
+        defaultValue: 'the target repo',
+      },
+      {
+        name: 'goal',
+        description: 'Product outcome or customer problem',
+        type: 'string',
+        defaultValue: 'meaningful product progress',
+      },
+      {
+        name: 'duration_minutes',
+        description: 'How long to stay in product-driver mode',
+        type: 'integer',
+        defaultValue: 60,
+      },
+    ],
+  },
+  {
+    name: 'issue-driver',
+    description: 'Triage, dedupe, and file actionable issues',
+    prompt:
+      'You are issue driver for {{repo}} for the next {{duration_minutes}} minutes.\n\nGoal: {{goal}}\n\nWork the issue queue like a product-minded engineer. Inspect existing issues and recent PRs first, then decide whether to update, close, split, prioritize, or file new issues. Prefer fewer high-quality issues over many vague ones.\n\nIssue quality bar:\n- concrete problem statement and why it matters\n- reproduction or evidence when applicable\n- proposed scope, non-goals, acceptance criteria, and verification plan\n- priority recommendation and dependencies\n- links to relevant files, PRs, conversations, or docs\n\nDo not implement fixes unless explicitly asked. End with a concise issue ledger and recommended next owners.',
+    options: [
+      {
+        name: 'repo',
+        description: 'Repository or project issue queue',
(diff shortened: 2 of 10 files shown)


Peyton-Spencer:
Mergemaster hold for the 2026-05-03 window.

Current state: PR #637 is clean and PR CI is green on head `c6267595f0db7e578b388460c248d7e3a424823e`:

- `format`: pass
- `typecheck`: pass
- `test`: pass
- `validate-skill-only`: pass
- CodeRabbit status: pass

Not merge-ready yet. CodeRabbit posted actionable review findings that should be resolved before approval/merge:

- `src/discord-command-flows.ts`: `enabled: []` currently behaves like allow-all because the allowlist only applies when `enabled.size > 0`. An explicitly present empty allowlist should be locked down as deny-all if that is the intended config semantics.
- `src/migrations.ts`: `migration1` appears to have been modified to include `registered_groups.discord_commands`; historical migrations should stay immutable and the new column should be introduced only by a new migration.
- `src/db.ts`: persisted `discordCommands` JSON is parsed without runtime shape validation before downstream code treats fields as string arrays.
- `src/index.ts`: synthesized registered groups may not preserve `discordCommands`, so command filtering can be lost on derived groups.
- `src/discord-command-flows.test.ts`: add a regression for the explicit `enabled: []` allowlist case.

Requested next step: fix or explicitly rebut each actionable item, then rerun CI and request a fresh review. This PR is otherwise correctly scoped to the new SDLC slash commands and per-agent command filters.

src/migrations.ts
```diff
@@ -288,7 +287,8 @@ const migration1: Migration = {
         backend TEXT,
         description TEXT,
         auto_respond_to_questions INTEGER DEFAULT 0,
-        auto_respond_keywords TEXT
+        auto_respond_keywords TEXT,
+        discord_commands TEXT
```
Peyton-Spencer: **Keep migration1 immutable.** Adding `discord_commands TEXT` to migration1's CREATE TABLE mutates a migration that has already been applied to existing DBs. The `addColumnIfNotExists` calls below are idempotent so this works in practice, but it sets a precedent that breaks the migration framework's core invariant: each migration runs exactly once and its `up` is fixed.

The column is already correctly added by `migration2` (line 504) for upgrading DBs. Drop it from migration1's CREATE TABLE so historical replay reproduces the pre-migration2 schema.

Confirming CodeRabbit's flag here.

src/discord-command-flows.ts
```diff
@@ -255,6 +379,19 @@ function loadFlowFile(filePath: string): DiscordFlowDefinition[] {
...
+function isCommandEnabledForGroup(
+  command: DiscordFlowDefinition,
+  group: RegisteredGroup,
+): boolean {
+  const config = group.discordCommands;
+  if (!config) return true;
+
+  const enabled = new Set(config.enabled || []);
+  const disabled = new Set(config.disabled || []);
+  if (enabled.size > 0 && !enabled.has(command.name)) return false;
```
Peyton-Spencer: **Allowlist semantics drop when `enabled` is `[]`.** The `enabled.size > 0` check means an empty allowlist falls through and every command is allowed. From a config standpoint, `enabled: []` reads as "no commands allowed" — the current behavior is the opposite.

Suggested: distinguish `enabled === undefined` (no allowlist, all allowed) from `enabled === []` (deny all). The Zod schema already preserves the distinction because `optional()` keeps `undefined` separate from `[]`.

```ts
const enabled = config.enabled;
if (enabled !== undefined && !enabled.includes(command.name)) return false;
```

Confirming CodeRabbit's flag here.

src/db.ts
```diff
@@ -149,6 +150,11 @@ function mapRowToRegisteredGroup(
       { jid: row.jid },
       'auto_respond_keywords',
     ),
+    discordCommands: safeJsonParse<RegisteredGroup['discordCommands']>(
```
Peyton-Spencer: **Validate `discord_commands` shape at the read boundary.** `safeJsonParse` only catches JSON syntax errors. A row with `{"enabled": "oops"}` parses fine and crashes later when `new Set(config.enabled || [])` runs in `isCommandEnabledForGroup`.

The Zod schema from `agent-topology.ts` (`discordCommandsSchema`) is exactly what's needed — export it and `safeParse` here, log + return undefined on failure (mirrors the corrupt-JSON warn). That keeps the agent-topology validator and the DB validator in sync.

Confirming CodeRabbit's flag here.

src/discord-command-flows.ts
```diff
@@ -292,7 +429,9 @@ export function getDiscordFlowDefinitionsForGroup(
     }
   }
 
-  return [...commands.values()].sort((a, b) => a.name.localeCompare(b.name));
+  return [...commands.values()]
+    .filter((command) => isCommandEnabledForGroup(command, group))
```
Peyton-Spencer, line 446: **P1 — workspace JSON can override system builtins.** The `commands.set(command.name, command)` call a few lines above (line 428) will replace a builtin if a workspace `discord-commands.json` defines a command with the same name as `resume` or `sessions`.

`SYSTEM_COMMAND_NAMES` (line 503) is computed once at module load from `BUILTIN_COMMANDS`, so the host's `handleSessionCommand` dispatch still treats `/resume` as system, but the actual definition served to Discord (and any rendered prompt) is the workspace one. A malicious or accidental workspace file could:

1. Replace `/resume` with `{ name: "resume", description: "...", prompt: "<arbitrary text sent to agent>" }`.
2. End up with the host short-circuiting to system handling on the agent path, or with a real prompt being injected if any code path renders the workspace definition.

Fix options (either is fine):
- In `loadFlowFile` / `normalizeFlow`, refuse any workspace command whose name is in `SYSTEM_COMMAND_NAMES` (`logger.warn` + skip).
- In the loop on line 427-429, skip the override when `commands.get(command.name)?.system === true`.

This is a P1 because system command names are part of our control plane and shouldn't be redefinable from a workspace-mounted file.

Peyton-Spencer (commented):
## Review summary

Solid feature: SDLC drivers + per-agent allow/deny config + Zod validation + tests. CI is green and 1933 tests pass locally. Before approving I want to land four substantive items — three already raised by CodeRabbit (which I'm independently confirming) and one new finding I haven't seen flagged elsewhere.

**Substantive (please address before merge):**
1. `migrations.ts:291` — mutating migration1's CREATE TABLE breaks the immutable-migration contract; column is already added by migration2.
2. `discord-command-flows.ts:391` — `enabled: []` short-circuits the allowlist check and degrades to "all allowed". Either treat empty as deny-all (recommended) or document.
3. `db.ts:153` — `safeJsonParse` only catches JSON syntax errors; reuse `discordCommandsSchema` from agent-topology to validate shape at the DB read boundary.
4. **(new, P1, correctness/security)** `discord-command-flows.ts` — workspace `discord-commands.json` can override builtin **system** commands (`/resume`, `/sessions`). The host's `SYSTEM_COMMAND_NAMES` is computed at module load from `BUILTIN_COMMANDS`, so a workspace override would still be treated as system by the host while serving an attacker-controlled prompt to Discord. Inline comment with details.

**Nits (non-blocking):**
- `discord-command-flows.ts` `buildDiscordSlashCommandPayloads` — JSON.stringify-based conflict detection across groups is O(n·m) and fragile to key ordering. Follow-up issue.
- `index.ts:470` — `fallback?.discordCommands` is fine today because the declarative topology path persists via `setRegisteredGroup`, but consider also pulling from `agent.discordCommands` once `Agent` carries that field, in case a future code path constructs a group without DB round-trip.

**Verification:**
- Pulled `c626759`, confirmed diff matches PR description (10 files, +297/-10).
- Read end-to-end: `types.ts` → `agent-topology.ts` (Zod) → `db.ts` (persist/load) → `discord-command-flows.ts` (filter + render) → `migrations.ts`.
- PR CI confirmed green: format, typecheck, test (1933 pass), validate-skill-only, CodeRabbit.

Will approve once #4 is resolved (P1) and #1-3 are either fixed or explicitly deferred with a follow-up issue.

Peyton-Spencer:
Addressed the review blockers in `0afe767c77cfe2ff3ffa13ccb07edde16b83ce85`.

Changes:
- Restored migration1 immutability by removing `registered_groups.discord_commands` from the v1 `CREATE TABLE` and v1 compatibility additions; migration2 is now the only migration that adds the column.
- Changed `enabled: []` semantics to deny all commands instead of falling through to allow all.
- Reused the exported `discordCommandsSchema` at the DB read boundary so malformed persisted JSON is ignored with a warning instead of flowing into `Set` construction.
- Prevented workspace `discord-commands.json` files from overriding built-in system commands like `/resume` and `/sessions`.
- Preserved `discordCommands` during canonical registered-group synthesis.

Regression coverage added:
- empty `enabled: []` allowlist
- custom `/resume` override rejection
- malformed persisted `discord_commands`

Verification:
- `bun test src/db.test.ts src/discord-command-flows.test.ts src/agent-topology.test.ts src/migrations.test.ts src/db.migrations.test.ts src/index.test.ts`
- `bun run typecheck`
- `bun run format:check`
- `bun test` — 1936 pass

GitHub PR checks are green on the updated head.

Peyton-Spencer:
Mergemaster update.

PR #637 has advanced to head `0afe767c77cfe2ff3ffa13ccb07edde16b83ce85`.

Current state:

- `mergeStateStatus`: CLEAN
- `format`: pass
- `typecheck`: pass
- `test`: pass
- `validate-skill-only`: pass
- CodeRabbit status: pass, though the follow-up review was rate-limited

Cody's follow-up commit explicitly addresses the previously listed blockers:

- migration1 immutability restored; migration2 owns `discord_commands`
- `enabled: []` now means deny-all
- DB readback validates `discordCommands`
- workspace custom commands cannot override host system commands like `/resume` and `/sessions`
- synthesized registered groups preserve `discordCommands`
- regression coverage added for the above

Holding merge for re-review / formal approval. Clayton's latest human review is still COMMENTED and `reviewDecision` is empty.

Peyton-Spencer (commented):
LGTM — all four review blockers addressed cleanly with regression coverage. (GitHub won't let me leave an APPROVED review since this token is the PR author's; treating this as a sign-off equivalent.)

1. *migration1 immutability* — `discord_commands` removed from migration1 CREATE TABLE and from the `addColumnIfNotExists` block; only migration2 owns the column now.
2. *Empty allowlist semantics* — `config.enabled !== undefined` distinguishes `[]` (deny-all) from `undefined` (no allowlist). New test `treats an empty enabled list as deny-all` locks the contract.
3. *DB shape validation* — new `parseDiscordCommands` in `db.ts` runs `discordCommandsSchema.safeParse` and warns + returns undefined on shape mismatch. Test `ignores malformed persisted Discord command filters` covers it.
4. *System command override* — `BUILTIN_SYSTEM_COMMAND_NAMES` extracted; `getDiscordFlowDefinitionsForGroup` skips workspace commands that collide with system names with a warn log. Test `does not let custom command files override system commands` covers it.

Bonus: `refreshRegisteredGroupsFromCanonicalState` now carries `discordCommands` from the legacy/fallback row, addressing the canonical-synthesis fragility I called out as a nit.

Verification on `0afe767c`:
- Read full diff against `c626759`.
- `bun test src/discord-command-flows.test.ts src/db.test.ts src/agent-topology.test.ts` — 72 pass, 0 fail.
- PR-triggered CI green: format, typecheck, test, validate-skill-only.

Ready for a non-mirroring human/agent reviewer to formally approve and merge.

Peyton-Spencer:
Approved after re-review of head `0afe767c77cfe2ff3ffa13ccb07edde16b83ce85`.

Verified:
- GitHub checks are green: format, typecheck, test, validate-skill-only, CodeRabbit.
- Focused local tests pass in an isolated worktree:
  `bun test src/discord-command-flows.test.ts src/db.test.ts src/agent-topology.test.ts`
  Result: 72 pass, 0 fail.
- Prior blockers are addressed in code and covered by regression tests:
  migration1 restored with `discord_commands` owned by migration2, `enabled: []`
  deny-all behavior, DB readback validation for malformed command config,
  preservation through synthesized groups, and custom command files cannot
  override built-in system commands.

No remaining merge blockers from my review.

PR #647, opened 2026-05-04
feat(discord): add session command namespace

Peyton-Spencer:
## Summary

- add a native `/session` Discord slash command with `new`, `resume`, `list`, `current`, `end`, and `rename` subcommands
- keep `/resume` and `/sessions` as deprecated aliases that delegate to `/session resume` and `/session list`
- persist session display names/end markers and queue names/forks for the next emitted session id
- pass `forkSession` through the host/container protocol so `/session new resume_from:` can branch Claude sessions

Closes #646.

## Verification

- `bun test src/discord-command-flows.test.ts src/db.test.ts`
- `bun test src/db.migrations.test.ts src/migrations.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **New Features**
  * Added `/session` Discord command with subcommands (`new`, `resume`, `list`, `current`, `end`, `rename`)
  * Enabled session naming and session branching capabilities

* **Documentation**
  * Updated documentation for new `/session` command structure
  * Marked `/resume` and `/sessions` commands as deprecated (available for one release with deprecation notice)

PR #648, opened 2026-05-04
feat: add Discord session command group

### LINKED ISSUE #646: feat: consolidate Discord session control into native /session subcommand group
## Problem

Session control in Discord is currently split across two flat top-level commands (`src/discord-command-flows.ts:121-139`):

- `/resume [session_id]` — resume a previous session in this channel
- `/sessions` — list recent sessions

There is no Discord-native way to **start a fresh session** explicitly, and the existing surface is two unrelated commands instead of one cohesive namespace. As we lean harder on session-as-a-primitive (mergemaster, taskbooker, scheduler windows + ad-hoc agent runs), users need first-class control over the session lifecycle from the channel itself.

## Proposal

Consolidate session control into a single Discord-native `/session` command using **slash command subcommand groups** (Discord's native nested option type).

Proposed surface:

```
/session new       [name?] [resume_from?]   — start a fresh agent session in this channel
/session resume    [session_id?]            — resume a previous session (omit id → picker)
/session list      [limit?]                 — list recent sessions
/session current                            — show the active session (id, started, last activity)
/session end       [session_id?]            — end current or specified session
/session rename    <session_id> <name>      — give a session a human-readable name
```

Notes:
- Discord renders subcommands as a true dropdown in the slash UI, so this is much more discoverable than 6 sibling commands.
- `/resume` and `/sessions` would be aliased/deprecated for one release, then removed.
- `new` should accept an optional `resume_from` so "branch off this session" is one click.

## Implementation Sketch

The current registration layer assumes a flat command shape:

- `DiscordFlowDefinition` (`src/discord-command-flows.ts`) only models top-level `options[]`, no nested subcommands.
- The discord.js builder path that consumes these definitions needs to support `addSubcommand()` / `addSubcommandGroup()`.
- System commands (`system: true`) bypass the prompt-template path and call into channel-local session handlers — those handlers need to be split per-subcommand.

Rough work breakdown:

1. Extend `DiscordFlowDefinition` to optionally carry `subcommands: DiscordFlowDefinition[]` (mutually exclusive with `options` at the top level when subcommands are present).
2. Update the Discord registration code (wherever `BUILTIN_COMMANDS` gets converted to `SlashCommandBuilder`) to walk the subcommand tree.
3. Add a `session` system flow with the six subcommands above; route each to the existing session manager.
4. Keep `resume`/`sessions` registered as deprecated aliases that delegate to `/session resume` and `/session list` respectively (one release of overlap).
5. Update `discord-command-flows.test.ts` to cover the nested shape + alias behavior.
6. Doc the new surface in `README.md` and any user-facing flow docs.

## Acceptance Criteria

- [ ] `/session` appears in Discord with a native subcommand picker
- [ ] All six subcommands work end-to-end against the existing session store
- [ ] `/resume` and `/sessions` still work but log a deprecation warning to the channel
- [ ] Tests cover subcommand parsing, alias delegation, and the system-handler routing
- [ ] README + flow docs reflect the new namespace

## Out of Scope

- Cross-channel session sharing (separate concern)
- Session persistence/restoration across container restarts (tracked elsewhere)
- Renaming the underlying session id format

---

Requested by Peyton in #omniclaw 2026-05-04. Wants this to feel discord-native rather than a flat command list.

Peyton-Spencer:
## Summary

- Adds a native `/session` Discord command with `new`, `resume`, `list`, `current`, `end`, and `rename` subcommands.
- Keeps legacy `/resume` and `/sessions` aliases, routing them through the new handler with deprecation notices.
- Adds session clearing, pending friendly names, fork-from staging, and Claude SDK `forkSession` forwarding for `/session new resume_from`.
- Documents the new command surface and adds tests for payload registration plus Discord handler routing.

Closes #646.

## Verification

- `bun test src/discord-command-flows.test.ts src/channels/discord.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun test`

### DIFF
--- a/src/channels/discord.test.ts
+++ b/src/channels/discord.test.ts
@@ -312,6 +312,87 @@ describe('Discord slash flows', () => {
     expect(interaction.followUp).not.toHaveBeenCalled();
   });
 
+  it('forwards /session new resume_from to the host session handler', async () => {
+    let received:
+      | {
+          command: string;
+          resumeFrom?: string;
+          name?: string;
+        }
+      | undefined;
+    const channel = new DiscordChannel({
+      token: 'test-token-not-used',
+      botId: 'PRIMARY',
+      onSessionCommand: (command, _chatJid, _group, options) => {
+        received = {
+          command,
+          resumeFrom: options?.resumeFrom,
+          name: options?.name,
+        };
+        return { message: 'queued' };
+      },
+    });
+
+    setAgent({
+      id: 'clayton-discord',
+      name: 'Clayton',
+      folder: 'clayton-discord',
+      backend: 'apple-container',
+      agentRuntime: 'claude-agent-sdk',
+      isAdmin: false,
+      createdAt: '2026-05-01T00:00:00.000Z',
+    });
+    setChannelSubscription({
+      channelJid: 'dc:1474995286903361772',
+      agentId: 'clayton-discord',
+      trigger: '@Clayton',
+      requiresTrigger: true,
+      priority: 0,
+      isPrimary: true,
+      discordBotId: 'PRIMARY',
+      discordGuildId: '753336633083953213',
+      createdAt: '2026-05-01T00:00:00.000Z',
+    });
+
+    const interaction = {
+      id: '1499999999999999999',
+      commandName: 'session',
+      channelId: '1474995286903361772',
+      guildId: '753336633083953213',
+      inGuild: () => true,
+      memberPermissions: { has: () => true },
+      options: {
+        getSubcommand: () => 'new',
+        getString: (name: string) =>
+          name === 'resume_from'
+            ? '123e4567-e89b-12d3-a456-426614174000'
+            : name === 'name'
+              ? 'launch triage'
+              : null,
+        getInteger: () => null,
+        getBoolean: () => null,
+      },
+      deferReply: mock(async () => {}),
+      editReply: mock(async () => {}),
+      followUp: mock(async () => {}),
+    };
+
+    await (
+      channel as unknown as {
+        handleSlashCommand: (input: typeof interaction) => Promise<void>;
+      }
+    ).handleSlashCommand(interaction);
+
+    expect(received).toEqual({
+      command: 'new',
+      resumeFrom: '123e4567-e89b-12d3-a456-426614174000',
+      name: 'launch triage',
+    });
+    expect(interaction.editReply).toHaveBeenCalledWith({
+      content: 'queued',
+    });
+  });
+
   it('routes legacy /sessions to /session list with deprecation metadata', async () => {
     let received:
       | {
--- a/src/channels/discord.test.ts
+++ b/src/channels/discord.test.ts
@@ -230,6 +230,161 @@ describe('Discord download guards', () => {
 });
 
 describe('Discord slash flows', () => {
+  it('routes /session subcommands to the host session handler', async () => {
+    let received:
+      | {
+          command: string;
+          sessionId?: string;
+          name?: string;
+        }
+      | undefined;
+    const channel = new DiscordChannel({
+      token: 'test-token-not-used',
+      botId: 'PRIMARY',
+      onSessionCommand: (command, _chatJid, _group, options) => {
+        received = {
+          command,
+          sessionId: options?.sessionId,
+          name: options?.name,
+        };
+        return { message: 'renamed' };
+      },
+    });
+
+    setAgent({
+      id: 'clayton-discord',
+      name: 'Clayton',
+      folder: 'clayton-discord',
+      backend: 'apple-container',
+      agentRuntime: 'claude-agent-sdk',
+      isAdmin: false,
+      createdAt: '2026-05-01T00:00:00.000Z',
+    });
+    setChannelSubscription({
+      channelJid: 'dc:1474995286903361772',
+      agentId: 'clayton-discord',
+      trigger: '@Clayton',
+      requiresTrigger: true,
+      priority: 0,
+      isPrimary: true,
+      discordBotId: 'PRIMARY',
+      discordGuildId: '753336633083953213',
+      createdAt: '2026-05-01T00:00:00.000Z',
+    });
+
+    const interaction = {
+      id: '1499999999999999999',
+      commandName: 'session',
+      channelId: '1474995286903361772',
+      guildId: '753336633083953213',
+      inGuild: () => true,
+      memberPermissions: { has: () => true },
+      options: {
+        getSubcommand: () => 'rename',
+        getString: (name: string) =>
+          name === 'session_id'
+            ? '123e4567-e89b-12d3-a456-426614174000'
+            : name === 'name'
+              ? 'launch triage'
+              : null,
+        getInteger: () => null,
+        getBoolean: () => null,
+      },
+      deferReply: mock(async () => {}),
+      editReply: mock(async () => {}),
+      followUp: mock(async () => {}),
+    };
+
+    await (
+      channel as unknown as {
+        handleSlashCommand: (input: typeof interaction) => Promise<void>;
+      }
+    ).handleSlashCommand(interaction);
+
+    expect(received).toEqual({
+      command: 'rename',
+      sessionId: '123e4567-e89b-12d3-a456-426614174000',
+      name: 'launch triage',
+    });
+    expect(interaction.editReply).toHaveBeenCalledWith({
+      content: 'renamed',
+    });
+    expect(interaction.followUp).not.toHaveBeenCalled();
+  });
+
+  it('routes legacy /sessions to /session list with deprecation metadata', async () => {
+    let received:
+      | {
+          command: string;
+          deprecatedAlias?: string;
+        }
+      | undefined;
+    const channel = new DiscordChannel({
+      token: 'test-token-not-used',
+      botId: 'PRIMARY',
+      onSessionCommand: (command, _chatJid, _group, options) => {
+        received = {
+          command,
+          deprecatedAlias: options?.deprecatedAlias,
+        };
+        return { message: 'listed' };
+      },
+    });
+
+    setAgent({
+      id: 'clayton-discord',
+      name: 'Clayton',
+      folder: 'clayton-discord',
+      backend: 'apple-container',
+      agentRuntime: 'claude-agent-sdk',
+      isAdmin: false,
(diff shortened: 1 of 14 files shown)


Peyton-Spencer:
@Clayton please review when you have a pass available. Please leave feedback directly in the GitHub PR, using file/line comments where applicable.

src/index.ts
```diff
@@ -367,49 +388,215 @@ function handleSessionCommand(
...
+    if (resumeFrom) {
+      const validation = validateSessionId(resumeFrom);
+      if (validation) return { message: validation };
+      const sessionFile = path.join(sessionsDir, `${resumeFrom}.jsonl`);
+      if (!fs.existsSync(sessionFile)) {
+        return {
+          message: `Session \`${resumeFrom}\` not found. Use \`/session list\` to list available sessions.`,
+        };
+      }
+      sessions[runtimeFolder] = resumeFrom;
```
Peyton-Spencer, line 595: [P1] `/session new resume_from:X` writes `sessions[runtimeFolder] = resumeFrom` and persists it via `setSession(...)`, while only marking the fork intent in the in-memory `pendingForkSessions` set (line 440).

Two failure modes fall out of that:

1. **Restart between `/session new resume_from:X` and the next user message.** On boot, `pendingForkSessions` is empty (it is never persisted) but the DB-backed session row still points at `X`. The next message resumes `X` in place instead of forking, silently corrupting the source session that the operator was trying to preserve.
2. **Race with another lane on the same `runtimeFolder`** (scheduled task, IPC-triggered task, etc.). Whatever fires first consumes `forkSession` and clears the pending flag at runAgent.ts L2014/L2119; the user's next interactive message then resumes the forked session instead of forking again.

At minimum, please persist the fork intent next to the session row (e.g. add a `pending_fork` column or a sibling table) and clear it in the same place `setSession` is updated. If persistence is too much for this PR, please at least guard the in-memory state with a `runtimeFolder`-keyed sentinel that includes the source `sessionId` so we can detect drift on boot and warn.

src/index.ts
```diff
@@ -207,6 +211,8 @@ async function shutdown(): Promise<void> {
 
 let lastTimestamp = '';
 let sessions: Record<string, string> = {};
+const pendingForkSessions = new Set<string>();
+const pendingSessionNames = new Map<string, string>();
```
Peyton-Spencer: [P2] Both `pendingForkSessions` and `pendingSessionNames` are module-scoped maps that never get rehydrated. Combined with the comment on line 438, this means **any** `/session new` (with or without `resume_from`) that is followed by a bot restart loses the user-visible promise — the next message either resumes the prior session (no resume_from) or replays the source session in place (with resume_from).

For pending names specifically, this is silently lossy: the user does `/session new name:"launch triage"`, the bot restarts, the next message starts a fresh session, the SDK emits a new `newSessionId`, and `applyPendingSessionName` is a no-op because the in-memory map is empty. The friendly name is gone with no warning.

Easiest fix is a small `pending_session_intents` table (group_folder, fork_from, name) that is read on `runAgent` start and cleared on first successful `newSessionId`.

src/index.ts
```diff
@@ -367,49 +388,215 @@ function handleSessionCommand(
...
+      clearSession(runtimeFolder);
+      resumePositionStore.set(runtimeFolder, '');
+      pendingForkSessions.delete(runtimeFolder);
+      pendingSessionNames.delete(runtimeFolder);
+      return {
+        message: `Session \`${sessionId}\` ended for this channel. The next message will start fresh.`,
+      };
+    }
+    return {
+      message: `Session \`${sessionId}\` is not the active session for this channel.`,
```
Peyton-Spencer, line 627: [P2] `/session end <session_id>` only ends the session if it matches the active one — otherwise it returns "is not the active session". That makes the optional argument effectively a confirmation token rather than a way to end an arbitrary session.

That is a defensible safety choice, but the README copy ("ends the active session, or verifies a specified active session before ending it") and the option description ("Specific session ID to end") both read like the user can end any session. Either:

- Tighten the README + option description to say "must match the active session", or
- Allow ending non-active sessions (deleting the .jsonl + clearing names) — which is closer to what most users will expect.

I would lean toward making this a confirm-only flow but renaming the option to something like `confirm_id` so the surface matches the behavior.

src/index.ts
```diff
@@ -367,49 +388,215 @@ function handleSessionCommand(
...
+      return {
+        message: `Session \`${sessionId}\` not found. Use \`/session list\` to list available sessions.`,
+      };
+    }
+    const name = normalizeSessionName(options.name);
+    if (!name) return { message: 'Session name is required.' };
+    const names = readSessionNames(sessionsDir);
+    names[sessionId] = name;
+    writeSessionNames(sessionsDir, names);
+    return { message: `Session \`${sessionId}\` renamed to "${name}".` };
```
Peyton-Spencer: [P2] Session names flow into Discord messages without escaping (e.g. `"renamed to \"${name}\""` here, and `` ` — ${names[s.sessionId]}` `` in `/session list`). `normalizeSessionName` only collapses whitespace and caps to 80 chars, so a name like `` foo`` or with `**` will break the rendered card or smuggle bold/italic into the list.

Either strip backticks/asterisks/underscores from the normalized name, or wrap rendered names in inline code (`` `name` ``) so the surrounding markdown stays inert. Same applies to `/session current` (line 422-423) and `/session list` (line 397-400).

src/index.ts
```diff
@@ -367,49 +388,215 @@ function handleSessionCommand(
       })
       .sort((a, b) => b.modifiedAt.getTime() - a.modifiedAt.getTime());
 
-    const lines = sessionInfos.map((s) => {
+    const limit =
+      typeof options.limit === 'number' && options.limit > 0
+        ? Math.min(options.limit, 25)
+        : 10;
```
Peyton-Spencer: [P3] The `Math.min(options.limit, 25)` cap and the default `10` are silently magic numbers. Discord allows up to 4000-char messages; 25 fits in that budget but it would be nice to either expose the cap as a named constant or surface it in the option description ("default 10, max 25") so operators understand why a request for `limit:50` only returns 25.

src/channels/discord.test.ts
```diff
@@ -230,6 +230,161 @@ describe('Discord download guards', () => {
...
+        handleSlashCommand: (input: typeof interaction) => Promise<void>;
+      }
+    ).handleSlashCommand(interaction);
+
+    expect(received).toEqual({
+      command: 'list',
+      deprecatedAlias: 'sessions',
+    });
+    expect(interaction.editReply).toHaveBeenCalledWith({
+      content: 'listed',
```
Peyton-Spencer, line 465: [P3] Test coverage gap: this PR adds explicit Discord adapter tests for `/session rename` and the `/sessions` legacy alias, but `/session new resume_from`, `/session end`, `/session current`, and the `/resume` legacy alias all go uncovered. Given the new state machine in `pendingForkSessions`/`pendingSessionNames`, please add at minimum a `/session new resume_from:<id>` test that asserts the `resumeFrom` option is forwarded.

Peyton-Spencer:
Solid consolidation of the session surface — the `/session` group reads cleanly, the legacy `/resume`/`/sessions` aliases are routed correctly, and the dispatch through `resolveSessionCommand` is a nice refactor.

Asking for changes mainly around the durability of the new session-intent state machine:

**P1 — fork intent persistence** (`src/index.ts:438`)
`/session new resume_from:X` writes the source session id into the persistent `sessions` table, but only marks the "please fork" intent in an in-memory `pendingForkSessions` set. A bot restart between the slash command and the next message will silently resume the source session in place, which is the opposite of what the user asked for. Same restart hazard applies to pending names.

**P2 — markdown injection in session names** (`src/index.ts:497`, list/current renderers)
`normalizeSessionName` only collapses whitespace; backticks and `**` survive into rendered messages.

**P2 — `/session end` semantics mismatch** (`src/index.ts:475`)
The optional `session_id` is effectively a confirmation token, not a way to end other sessions, but the README and option description read otherwise.

**P3 — limit cap is magic** (`src/index.ts:394`) and **test coverage** for `/session new resume_from`, `/session end`, `/session current`, and the `/resume` legacy alias.

Heads-up that PR #647 (Cody's parallel implementation against issue #646) takes a different approach — adds protocol-layer plumbing and a migration for session metadata. Worth coordinating to land one of these and close the other rather than running in parallel.

Peyton-Spencer:
Rebased onto current `main` and force-pushed `dex/issue-646-session-command` at `b23d5ea`.

Conflict resolution:
- Preserved main's required-option ordering for Discord slash payloads.
- Preserved #648's native `/session` subcommand payload.
- Pulled the small shared `forkSession` protocol type contract from #647 so the agent-runner no longer needs a local cast.
- Left #647's DB-backed session metadata out of this PR because #648 already implements scoped session naming via session-dir metadata, and adding a DB migration would expand the rebase scope.

Local verification:
- `bun test src/discord-command-flows.test.ts src/channels/discord.test.ts` — 60 pass / 0 fail
- `bun run typecheck` — pass
- `bun run format:check` — pass
- `bun test` — 1973 pass / 0 fail / 5291 expect() calls

CI restarted on the rebased branch.

Peyton-Spencer:
Addressed the unresolved May 4 review threads and pushed `0fb9ebf`.

Fixes:
- Added a durable `pending_session_intents` table and DB accessors for pending `/session new` fork/name intents.
- `/session new resume_from:<id>` now persists `fork_from` and optional friendly name instead of relying on module-scoped maps.
- `runAgent` reads the durable intent, only forks when the active session still matches the persisted source session, warns/clears stale drift, and clears the intent when a new session ID is produced.
- Pending session names survive restart and are applied when the SDK emits the new session ID.
- Session names are normalized to remove Markdown control characters before storage/rendering.
- `/session end [session_id]` copy is now explicitly confirm-only, and `/session list` documents default/max limit.
- Added coverage for `/session new resume_from`, pending intent persistence/cleanup, sanitized names, `/session end` confirmation semantics, and migration schema.

Verification:
- `bun test src/index.test.ts src/ipc-processing.test.ts src/discord-command-flows.test.ts src/channels/discord.test.ts src/db.migrations.test.ts`: 148 pass / 0 fail
- `bun run typecheck`: pass
- `bun test`: 2020 pass / 0 fail / 5409 expect() calls

CI restarted on the rebased branch.

Peyton-Spencer:
PM nudge from #706: this PR has been CI-green and unreviewed since 2026-05-04 (12 days). Flagged in two consecutive assessments. Requesting a reviewer pickup.

PR #651, opened 2026-05-04
feat(web): add per-agent off switch

Peyton-Spencer:
## Summary

Adds a per-agent off switch driven from the Web UI. From either the
Agents list or the Agent detail page you can now disable a specific
agent — the orchestrator stops dispatching new messages to it and the
scheduler stops running its scheduled tasks until you re-enable it.
Inbound messages keep persisting; toggling back resumes from the next
unread message on the next loop tick.

## Behavior

- **Agents page** — local agents now show an `enable` / `disable`
  button next to `detail` / `messages`. Disabled rows render with the
  new `disabled` exec status badge and dim the agent name.
- **Agent detail page** — same toggle is available in the header next
  to the existing badges; the status badge flips to `disabled` while
  off.
- **Off semantics** — message dispatch (both trigger and attentive
  follow-up paths) and recovery skip disabled agents. The task
  scheduler skips runs for disabled agents but `next_run` still
  advances normally so re-enabling resumes the recurring schedule
  cleanly.
- Remote (peer) agents do not show the toggle — they are owned by
  their host instance.

## Schema

- New `migrations.ts` `migration3` adds an `enabled INTEGER NOT NULL`
  column to `agents` defaulting to `1`. `BASELINE_VERSION` bumped to 3.
  Idempotent via `addColumnIfNotExists` and the baseline `CREATE TABLE`
  was updated to match.

## API

- `POST /api/agents/:id/enabled` with `{ "enabled": boolean }`.
  Returns `200 { ok, agentId, enabled }`, `400` for invalid body,
  `404` for unknown agent, `413` for oversized body.

## Tests

- `src/web/routes.test.ts` — POST off-switch route: persistence,
  invalid body type, unknown agent, malformed JSON.
- `src/web/agents-page.test.ts` — disabled badge + enable button,
  disable button on enabled rows, no toggle for remote agents.
- All existing Web UI test mocks updated with a `setAgentEnabled` stub.
- `bun tsc --noEmit` clean. Targeted suites green
  (135/135 across the four affected files).
- Repo-wide `bun test` shows the same pre-existing 268-failure
  baseline as `main` — no regressions from this branch.

## Test plan

- [ ] Bring up the dashboard and disable a local agent from the agents
      list — verify status flips to `disabled` and the row dims.
- [ ] Send a message to a chat where the disabled agent is subscribed
      and confirm it does not receive a dispatch (other agents still
      respond normally).
- [ ] Schedule a one-off task that fires within a minute, then disable
      the owning agent — verify the run is skipped with the
      `Task skipped — agent is disabled` log line.
- [ ] Re-enable the agent and verify both message dispatch and the
      next scheduled run fire normally.
- [ ] Confirm the toggle button is not rendered for remote peer
      agents.

## Summary by CodeRabbit

* **New Features**
  * Per-agent enable/disable controls in the UI with badges and toggles; changes show a toast and refresh the view.
  * Disabled agents are visually dimmed and marked "disabled" and are skipped for message dispatch and scheduled tasks.
  * HTTP API to toggle agent enabled state so changes persist immediately.

* **Chores**
  * Database schema and migration updated to store agent enabled state so it persists across sessions.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -1488,6 +1488,21 @@ export function getAllAgents(): Record<string, Agent> {
 
 /** Insert or replace an agent record. */
 export function setAgent(agent: Agent): void {
+  // When the caller didn't specify `enabled`, preserve the existing row's
+  // value so registration paths (registerGroup, applyDeclarativeAgentTopology)
+  // don't silently re-enable an agent the user disabled in the Web UI.
+  let enabledValue: 0 | 1;
+  if (agent.enabled === false) {
+    enabledValue = 0;
+  } else if (agent.enabled === true) {
+    enabledValue = 1;
+  } else {
+    const existing = db
+      .prepare('SELECT enabled FROM agents WHERE id = ?')
+      .get(agent.id) as { enabled: number | null } | undefined;
+    enabledValue = existing && existing.enabled === 0 ? 0 : 1;
+  }
+
   db.query(
     `
     INSERT OR REPLACE INTO agents (id, name, description, folder, backend, agent_runtime, container_config, is_admin, server_folder, created_at, agent_context_folder, roster_role_filters, avatar_url, avatar_source, enabled)
@@ -1510,7 +1525,7 @@ export function setAgent(agent: Agent): void {
       : agent.rosterRoleFilters.join(','),
     agent.avatarUrl || null,
     agent.avatarSource || null,
-    agent.enabled === false ? 0 : 1,
+    enabledValue,
   );
 }
 
--- a/src/db.ts
+++ b/src/db.ts
@@ -60,6 +60,7 @@ interface AgentRow {
   roster_role_filters: string | null;
   avatar_url: string | null;
   avatar_source: string | null;
+  enabled: number | null;
 }
 
 interface AgentHealthRow {
@@ -204,6 +205,11 @@ function mapRowToAgent(row: AgentRow): Agent {
     rosterRoleFilters,
     avatarUrl: row.avatar_url || undefined,
     avatarSource: (row.avatar_source as Agent['avatarSource']) || undefined,
+    // Default to enabled when the column is missing (pre-migration row) or null.
+    enabled:
+      row.enabled === null || row.enabled === undefined
+        ? true
+        : row.enabled === 1,
   };
 }
 
@@ -1484,8 +1490,8 @@ export function getAllAgents(): Record<string, Agent> {
 export function setAgent(agent: Agent): void {
   db.query(
     `
-    INSERT OR REPLACE INTO agents (id, name, description, folder, backend, agent_runtime, container_config, is_admin, server_folder, created_at, agent_context_folder, roster_role_filters, avatar_url, avatar_source)
-    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+    INSERT OR REPLACE INTO agents (id, name, description, folder, backend, agent_runtime, container_config, is_admin, server_folder, created_at, agent_context_folder, roster_role_filters, avatar_url, avatar_source, enabled)
+    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
   `,
   ).run(
     agent.id,
@@ -1504,7 +1510,19 @@ export function setAgent(agent: Agent): void {
       : agent.rosterRoleFilters.join(','),
     agent.avatarUrl || null,
     agent.avatarSource || null,
+    agent.enabled === false ? 0 : 1,
   );
+}
+
+/**
+ * Update only the enabled flag for an agent (lightweight, avoids full setAgent).
+ * Returns true on success, false if the agent doesn't exist.
+ */
+export function setAgentEnabled(agentId: string, enabled: boolean): boolean {
+  const result = db
+    .query('UPDATE agents SET enabled = ? WHERE id = ?')
+    .run(enabled ? 1 : 0, agentId);
+  return result.changes > 0;
 }
 
 /** Update only the avatar fields for an agent (lightweight, avoids full setAgent). */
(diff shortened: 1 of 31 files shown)


src/index.ts
```diff
@@ -2103,7 +2104,9 @@ async function startMessageLoop(): Promise<void> {
         for (const [chatJid, groupMessages] of messagesByGroup) {
           const { selected: triggerSelected, selectedByTrigger } =
             selectSubscriptionsForMessage(chatJid, groupMessages);
-          let selectedSubs = triggerSelected;
+          let selectedSubs = triggerSelected.filter(
```
Peyton-Spencer: Filtering disabled subscriptions here can turn an explicit trigger match into `selectedSubs.length === 0`, which then falls through to the legacy registered-group path below. That path does not check `agents[group.folder]?.enabled`, so a disabled agent can still be dispatched via the chat-level queue key when its trigger is used, and legacy-only groups also bypass the off switch entirely. Please distinguish "nothing matched" from "matched agents were disabled" and skip dispatch in the latter case, and add the same enabled guard to the legacy fallback/recovery paths.

src/db.ts
```diff
@@ -1499,9 +1505,21 @@ export function setAgent(agent: Agent): void {
       : agent.rosterRoleFilters.join(','),
     agent.avatarUrl || null,
     agent.avatarSource || null,
+    agent.enabled === false ? 0 : 1,
```
Peyton-Spencer: This defaults every `setAgent()` write with no explicit `enabled` field to enabled. `registerGroup()` builds agents through `registeredGroupToAgent()`, which does not carry `enabled`, and `applyDeclarativeAgentTopology()` calls `registerGroup()` on startup. The result is that an agent disabled in the Web UI can be silently re-enabled by a restart or topology resync. Preserve the existing row's `enabled` value when `agent.enabled` is undefined, or have the registration path merge the persisted flag before calling `setAgent()`.

Peyton-Spencer (commented):
Reviewed the requested dispatch, scheduler, migration, and Web API paths. I found two blocking behavior issues around the new disabled state: dispatch can still fall through to the legacy path after filtering disabled subscriptions, and topology/setAgent writes can re-enable disabled agents on restart or resync.

Peyton-Spencer:
Pushed `05c8207` addressing both blocking issues plus the CodeRabbit findings:

**Blocking #1 — `src/index.ts` legacy-dispatch fallthrough**
Added `allTriggerMatchesDisabled` flag distinguishing "no trigger matched" from "all matches were disabled". When matches were disabled we now `continue` instead of falling through. The legacy registered-group fallback (and the recovery path) also explicitly check `agents[group.folder]?.enabled === false` before dispatching.

**Blocking #2 — `src/db.ts` setAgent re-enable on re-registration**
`setAgent()` now reads the existing row's `enabled` value when the caller doesn't specify one. `registerGroup()` and `applyDeclarativeAgentTopology()` no longer silently re-enable a disabled agent on restart. Added 3 tests in `db.test.ts` covering default-enabled, preserve-disabled-across-re-registration, and explicit-re-enable.

**CodeRabbit fixes**
- `routes.ts` — guard `decodeURIComponent` and return 400 on `URIError` (+ test).
- `page-scripts.ts` — both `agentsScript()` and `agentDetailScript()` now use a named handler with `window.__cleanup` to prevent listener stacking on SPA re-init.
- `agents-page.test.ts` — order-agnostic regex for the remote-only-no-toggle assertion.

Typecheck clean. 140/140 pass on touched suites; 62/62 on index + scheduler.

src/index.ts
```diff
@@ -2302,6 +2338,13 @@ function recoverPendingMessages(): void {
```
Peyton-Spencer, line 2397: This recovery guard closes one direct enqueue path, but startup confirmations just below still enqueue disabled agents directly, bypassing the message-loop and recovery enabled checks. `buildStartupConfirmationTargets()` does not know about `agents.enabled`, and `queueStartupConfirmationMessages()` calls `queue.enqueueMessageCheck(makeDispatchKey(...))` for subscribed targets (or the chat-level key for legacy targets), so a disabled agent can still be woken on restart when `STARTUP_CONFIRMATIONS` is enabled. Please filter/skip targets whose `target.agentId` is disabled, and also skip legacy targets when `agents[registeredGroups[target.chatJid]?.folder]?.enabled === false`.

Peyton-Spencer (commented):
Re-reviewed the updated PR. The original legacy-dispatch and setAgent re-registration blockers are fixed, and the targeted updated tests pass locally. I found one remaining off-switch bypass in startup confirmations.

Peyton-Spencer (commented):
Not merging this pass. CI is green, but the current review history still identifies disabled-agent bypass risks around agent ID/folder resolution; please address that resolver path and add targeted coverage before merge.

Peyton-Spencer:
Pushed follow-up fix in `5dc29d7`.

What changed:

- Added a single ID-or-folder enabled resolver in `src/index.ts`, then routed message-loop, recovery, startup-confirmation, and scheduler enabled checks through it.
- Kept trigger-matched-but-disabled subscriptions from falling through into the legacy registered-group path.
- Filtered startup confirmation targets before storing/enqueuing synthetic confirmation messages.
- Added focused regressions in `src/index.test.ts` for:
  - enabled resolution by agent ID and workspace folder
  - trigger match where all matched subscriptions are disabled
  - startup confirmations skipping disabled subscription and legacy targets

Verification:

- `bun test src/index.test.ts`: pass, 9 pass / 0 fail
- `bun test`: pass, 1957 pass / 0 fail
- `bun run typecheck`: pass

Ready for re-review.

Peyton-Spencer:
Rebased `clayton/agent-off-switch` onto current `main` and force-pushed with lease.

Rebase note:

- Preserved main's factory workflow migration as migration 3.
- Moved the off-switch `agents.enabled` migration to migration 4 and updated `BASELINE_VERSION` / `allMigrations` accordingly.

Final verification after rebase:

- `bun test src/index.test.ts src/db.migrations.test.ts src/migrations.test.ts`: pass, 30 pass / 0 fail
- `bun test`: pass, 2010 pass / 0 fail
- `bun run typecheck`: pass

Branch head is now `fdb2e8f`.

PR #665, opened 2026-05-07
fix(telegram): treat replies to bot as triggers and download binary docs

Peyton-Spencer:
## Summary

Two related Telegram bugs surfaced when LocalPeyton was running in a 3-person group:

- **Replies to the bot were silently dropped** when the channel had `requires_trigger=1`. A user replying conversationally (no @mention) hit `No agents selected after trigger filter — all candidates require explicit trigger` and the agent never woke up.
- **PDF / binary documents were never downloaded.** The handler emitted a `[File: name.pdf]` placeholder for any document that wasn't an image or a small text file, so the agent could see the filename but had no bytes to `Read`.

## Changes

- `NewMessage.is_reply_to_bot` — set by Telegram adapters when `ctx.message.reply_to_message.from.id` matches this bot's id (text, photo, document, and `storeNonText` paths).
- `messagesMatchSubscription` — treats `is_reply_to_bot === true` as an explicit trigger, alongside `@allagents` and direct @mention.
- Telegram document handler — downloads any binary up to `MAX_BINARY_DOWNLOAD_BYTES` (10 MiB) into the group's media dir and emits a new `[attachment:file path=media/<file> name=<orig>]` marker so the agent can open it with the `Read` tool. Path is relative to the agent's cwd (`/workspace/group`).
- `formatBinaryFileMarker(filename, originalName?)` helper in `media.ts`, with `path.basename` sanitization and tests covering happy path, display-name override, and traversal stripping.

Multi-attachment Telegram sends already work via this fix: grammY delivers each document as its own `message:document` event, so each PDF is processed and stored independently with its own marker.

## Test plan

- [x] `bun test` — 1970 pass, 0 fail
- [x] `bun run build` — builds clean
- [ ] Send a PDF in a Telegram group → agent receives `[attachment:file path=media/...]` and can `Read` it
- [ ] Send 2 PDFs back-to-back → both downloaded, both markers emitted as separate messages
- [ ] Reply (no @mention) to a bot message in a `requires_trigger=1` group → agent activates
- [ ] Cold @mention in same group still works
- [ ] DM (no group, no trigger) flow unchanged

## Summary by CodeRabbit

* **New Features**
  * Telegram messages replying to the bot's messages now automatically trigger responses without requiring an @mention.

* **Improvements**
  * Enhanced binary file attachment handling with improved formatting and display names.

* **Tests**
  * Added test coverage for binary file marker generation.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -256,6 +256,7 @@ export class TelegramChannel implements Channel {
       // Prepend reply context so the agent knows what's being replied to
       const replyTo = ctx.message.reply_to_message;
       const isReplyToBot = this.isReplyToThisBot(replyTo);
+      const replyToBotId = isReplyToBot ? String(replyTo?.from?.id) : undefined;
       if (replyTo && 'text' in replyTo && replyTo.text) {
         const truncated =
           replyTo.text.length > 200
@@ -293,6 +294,7 @@ export class TelegramChannel implements Channel {
         sender_platform: 'telegram',
         sender_user_id: senderUserId,
         is_reply_to_bot: isReplyToBot,
+        reply_to_bot_id: replyToBotId,
       });
 
       logger.info(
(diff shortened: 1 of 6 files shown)


Peyton-Spencer:
Rebased again onto current `main` after the off-switch merge and pushed `853aa3b`.

What changed in the final rebase:
- Preserved the new off-switch routing guard tests from `main`.
- Preserved the Telegram per-subscription reply routing tests and implementation.
- PR is no longer dirty against `main`.

Local verification:
- `bun test src/index.test.ts`: 12 pass / 0 fail / 19 expect() calls
- `bun run typecheck`: pass
- `bun test`: 2016 pass / 0 fail / 5384 expect() calls

GitHub status after push:
- merge state: CLEAN
- CI format/typecheck/test: green
- validate-skill-only: green
- CodeRabbit: green

Ready for merge consideration.

PR #666, opened 2026-05-07
test: cover GitHub context fetch paths

Peyton-Spencer:
## Summary
- Adds deterministic monkey-patch tests for `src/github.ts` API fetch paths without making network calls.
- Covers authenticated request headers, non-OK responses, combined check-suite status classification, repo context assembly, empty PR/issue responses, PR-as-issue filtering, disabled review-comment fetching, and stale-cache fallback.

## Test Count
- Before: `1967 pass / 0 fail`, `5267 expect() calls`, `101 files`.
- After: `1972 pass / 0 fail`, `5287 expect() calls`, `102 files`.

## Verification
- `bun test`
- `bun run tsc --noEmit`
- `bunx prettier --check src/github-fetch.test.ts`

## Remaining Coverage Gaps
- Large integration-heavy modules still have lower line coverage, especially channel adapters, `src/index.ts`, discovery routes, and local backend runtime paths.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for GitHub integration, verifying authorization headers, status classification (passing/failing/pending states), context building from pull requests and issues, and error resilience with cache fallback handling.

### DIFF
--- a/src/github-fetch.test.ts
+++ b/src/github-fetch.test.ts
@@ -221,16 +221,72 @@ describe('github API fetch paths', () => {
       repos: [{ owner: 'omniaura', repo: 'omniclaw' }],
     };
 
-    globalThis.fetch = (() =>
-      Promise.resolve(jsonResponse([]))) as unknown as typeof fetch;
+    globalThis.fetch = ((url) => {
+      const path = new URL(String(url)).pathname;
+
+      if (path.endsWith('/pulls')) {
+        return Promise.resolve(
+          jsonResponse([
+            {
+              number: 12,
+              title: 'Cached PR',
+              user: { login: 'peyton' },
+              head: { ref: 'cached-branch' },
+              base: { ref: 'main' },
+              state: 'open',
+              draft: false,
+              requested_reviewers: [],
+              html_url: 'https://github.com/omniaura/omniclaw/pull/12',
+              body: 'Cached body',
+              created_at: '2026-05-07T00:00:00Z',
+              updated_at: '2026-05-07T00:00:00Z',
+            },
+          ]),
+        );
+      }
+
+      if (path.endsWith('/reviews') || path.endsWith('/comments')) {
+        return Promise.resolve(jsonResponse([]));
+      }
+
+      if (path.endsWith('/check-suites')) {
+        return Promise.resolve(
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
Reviewed the diff and ran the focused verification. One test assertion looks like it can pass without exercising the fallback it names; I left that inline.

src/github-fetch.test.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    globalThis.fetch = (() =>
+      Promise.resolve(jsonResponse([]))) as unknown as typeof fetch;
+
+    const cached = await github.fetchGitHubContext(watch, 60_000);
+    globalThis.fetch = (() => {
+      throw new Error('offline');
+    }) as unknown as typeof fetch;
+
+    const stale = await github.fetchGitHubContext(watch, -1);
```
Peyton-Spencer, line 286: This does not appear to exercise the stale-cache fallback in `fetchGitHubContext`. `githubFetch` catches the thrown `fetch` error and returns `null`, so the second refresh still completes successfully with empty PR/issue sections instead of throwing into the outer `catch`. Because the first cached render in this test is also empty, `stale === cached` passes even though the stale cache was not used. A stronger version would seed the cache with non-empty content and then trigger an actual error past `githubFetch`, or change the assertion/description to match the degraded empty refresh behavior.

Peyton-Spencer:
Focused check: `bun test src/github-fetch.test.ts` passes, and the test-only scope is low risk. I am not merging yet because the stale-cache fallback test still does not actually exercise `fetchGitHubContext`'s catch/fallback path: `githubFetch` catches the thrown fetch error and returns `null`, so the repo render succeeds as an empty context and `expect(stale).toBe(cached)` only passes because both strings are identical. Please make that assertion prove the fallback path, for example by caching a non-empty first response and making the refresh path throw outside `githubFetch` or otherwise render different output if fallback is not used.

src/github-fetch.test.ts
Peyton-Spencer: Fixed in 638aca3: the test now seeds non-empty cached content and triggers a parse-shape failure past githubFetch so stale-cache fallback is actually exercised.

PR #681, opened 2026-05-09
fix(web): reject active remote image content

### LINKED ISSUE #674: security: remote avatar cache can serve active content same-origin
## Summary

The remote image cache proxies avatar/icon responses under the OmniClaw web UI origin but preserves the upstream `Content-Type` without verifying that the payload is a safe raster image. A malicious or compromised avatar URL can return active content such as `text/html` or `image/svg+xml`; after the first fetch, OmniClaw serves that content from `/api/agents/:id/avatar/image` with the attacker-controlled content type.

Severity: P1 / High

## Affected Code

- `src/web/routes.ts` calls `serveCachedRemoteImage()` for `/api/agents/:id/avatar/image`, chat icons, and Discord guild icons.
- `src/web/image-cache.ts` accepts any successful upstream response, stores the bytes, stores `upstream.headers.get('content-type')`, and replays that content type in `buildCachedResponse()`.

## Why This Matters

The dashboard normally embeds these URLs in `` tags, where HTML is unlikely to execute. However, the cached image endpoint is same-origin with the web UI. If an authenticated operator directly opens the cached avatar endpoint, or if another same-origin page/link navigates to it, a `text/html` response can execute script with OmniClaw web UI origin privileges.

This can expose web UI state and perform authenticated API actions against endpoints such as task creation, context-file writes, message sending, and agent avatar mutation.

## Reproduction Sketch

1. Host a publicly reachable URL that returns status `200`, `Content-Type: text/html`, and a body like `<script>fetch('/api/tasks').then(r=>r.text()).then(t=>document.body.textContent=t)</script>`.
2. Use `PUT /api/agents/{agentId}/avatar` with `{ "source": "custom", "url": "https://attacker.example/avatar" }`.
3. Request `/api/agents/{agentId}/avatar/image`.
4. Observe that OmniClaw fetches and caches the HTML, then serves it from the OmniClaw origin with `Content-Type: text/html`.

The same pattern applies to other `serveCachedRemoteImage()` call sites if their resolver can be influenced to return an attacker-controlled URL or content type.

## Recommended Fix

- Enforce an allowlist of safe image content types before caching or serving remote image bytes.
- Prefer raster-only types such as `image/png`, `image/jpeg`, `image/gif`, `image/webp`, and possibly `image/bmp`.
- Reject or normalize `image/svg+xml` unless it is sanitized, because SVG can contain active content when served as a top-level document.
- Consider sniffing magic bytes for common raster formats so a malicious server cannot bypass the check with a forged `Content-Type`.
- Return cached remote images with `X-Content-Type-Options: nosniff`.

## Areas Reviewed

- `src/web/routes.ts` avatar/icon proxy routes.
- `src/web/image-cache.ts` remote URL validation, DNS rebinding protection, byte caps, cache metadata, and response replay.
- Existing open/all GitHub issues for duplicate avatar/image-cache content-type XSS reports.

Peyton-Spencer:
## Summary

- Closes #674 by rejecting remote avatar/icon responses unless the upstream `Content-Type` is a known raster image type.
- Adds `X-Content-Type-Options: nosniff` to cached image responses and refuses to replay cached entries with unsafe MIME metadata.
- Adds regression tests for HTML payloads, SVG script payloads, and unsafe cached metadata.

## Verification

- `bun test src/web/image-cache.test.ts src/web/routes.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/web/image-cache.test.ts
+++ b/src/web/image-cache.test.ts
@@ -1,4 +1,4 @@
-import { randomUUID } from 'crypto';
+import { createHash, randomUUID } from 'crypto';
 import fs from 'fs';
 import http from 'node:http';
 import path from 'path';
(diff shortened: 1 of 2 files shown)


Peyton-Spencer:
Reviewed the remote image content-type hardening. I did not find any blocking issues.

What I checked:
- `serveCachedRemoteImage` now refuses unsafe upstream MIME types before buffering or writing cache entries.
- Cached metadata is revalidated before replay, so old unsafe cache entries are not served.
- Cached responses now include `X-Content-Type-Options: nosniff`.
- Existing avatar/icon route behavior still passes under the stricter image content-type requirement.

Local verification:
- `bun test src/web/image-cache.test.ts`
- `bun test src/web/routes.test.ts src/web/agents-page.test.ts src/web/agent-detail.test.ts src/web/dashboard.test.ts`
- `bun run typecheck`
- `bun run format:check`

Issue #644, opened 2026-05-04
tracking: operator observability rollup — extend /ipc and /system with run, queue, and reason-code diagnostics


Peyton-Spencer:
## Context

The Web UI now spans 11 pages with system health, IPC inspector, queue surfaces, network/discovery, conversations, etc. ROADMAP.md (medium-term) calls out a remaining gap:

> Logs and web pages exist, but the operator story is still incomplete.
>
> - better surfacing of active runs, queue state, peer health, and task outcomes
> - lightweight diagnostics for why an agent is idle, blocked, retrying, or offline

There is no tracking issue for that work today. Filing this so the gaps have a home and so PRs can close concrete sub-items.

Per agent review (Cody, OCPeyton): this should *extend existing surfaces* (`/ipc`, `/system`) before any new top-level route. A `/factory` route, if any, follows the durable workflow layer (separate tracker).

## User problem

When something goes sideways during a busy multi-agent session, an operator currently has to:

- Open multiple pages and cross-reference run state, queue depth, and logs
- Read raw logs to figure out *why* an agent is idle (no work, blocked on lock, retrying, offline)
- Guess whether a delay is queue back-pressure, container cold start, lease contention, or a peer issue

We need lightweight, consolidated diagnostics in the existing pages — not a new dashboard.

### In scope (this tracker)

- *`/ipc` enrichments* — active run breakdown per group/lane, retry counts and last error, lock/lease state once #453 lands, age of in-flight items
- *`/system` enrichments* — agent state breakdown (active / idle reason / blocked / retrying / offline), recent task outcomes, peer health rollup
- *Reason codes* — replace "idle" with structured reasons ("waiting on lease", "no work", "offline", "back-pressure", "rate-limited") wherever cheap
- *API extensions* — extend existing `WebStateProvider` / SSE event payloads only where the data is missing, not new endpoints
- *Mobile-responsive parity* for any new tiles added (the layout work is already done)

### Non-goals (defer)

- New top-level route (`/factory`, `/observability`) — premature without a workflow domain model
- Long-term metrics persistence / time-series graphs — start with current-state rollups
- External observability integrations (Prometheus, etc.)
- Replacing existing logs viewer

## Acceptance criteria

- [ ] `/ipc` shows active runs per lane with current state and age
- [ ] `/ipc` surfaces retry count and last error message inline (truncated, with link to logs)
- [ ] `/system` shows per-agent state breakdown with structured reason codes
[...]

PR #688, opened 2026-05-11
feat(web): surface message lane running age on /ipc

Peyton-Spencer:
## Summary

Adds visibility into how long a message lane has been processing on `/ipc`, mirroring the existing task-lane `runningMs` surface.

- `GroupState` now tracks `messageRunStartedAt` (set in `startMessageRun`, cleared in `finishMessageRun`)
- `GroupQueueDetail.messageLane` exposes optional `startedAt` and `runningMs` fields
- IPC inspector renders a `.lane-age` badge next to the message-lane status whenever a run is in flight (e.g. `0.4s`, `2.1m`)
- New CSS class `.lane-age` styled consistently with `.lane-reason` (tabular numerals for monotonic readability)

## Motivation

Part of #644 (operator observability rollup). Acceptance criteria item:

> /ipc shows active runs per lane with current state and age

The task lane already shows `runningMs` via `activeTask.runningMs`. This PR closes the gap for the message lane so operators can spot stuck or long-running message runs at a glance.

## Verification

- `bun test` → 2026 pass / 0 fail (+3 new tests)
- `bunx tsc --noEmit` → clean
- New tests:
  - `GroupQueue.getDetailedStats message lane run age` — verifies `startedAt`/`runningMs` are populated in-flight and cleared on completion
  - `renderIpcInspector` — verifies the `.lane-age` badge is rendered when `runningMs` is provided and omitted when it is null

## Backward compatibility

The new `startedAt` / `runningMs` fields are optional on `GroupQueueDetail.messageLane`, so existing fixtures and external consumers remain valid. Reason-code derivation paths are untouched.

## Summary by CodeRabbit

* **New Features**
  * The IPC inspector now displays message-lane run duration in the group queue view, showing real-time elapsed time for active message processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/omniaura/omniclaw/pull/688)

### DIFF
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -527,6 +527,9 @@ export class GroupQueue {
     }
 
     this.transitionMessageLaneState(groupJid, 'cooldown');
+    // Clear the run timer when entering cooldown so /ipc doesn't show a
+    // growing "running age" while the lane is actually idle-waiting.
+    state.messageRunStartedAt = null;
 
     // A processing slot just freed up — let waiting messages start.
     this.drainWaitingMessages();
@@ -609,6 +612,9 @@ export class GroupQueue {
     }
     // Container was idle — mark it active again (single-method bookkeeping)
     this.transitionMessageLaneState(chatJid, 'running');
+    // Stamp the start of this resumed run so /ipc reports the age of the
+    // current message run, not the original container-start timer.
+    state.messageRunStartedAt = Date.now();
     const channelJid = toChannelJid(chatJid);
 
     // Use Effect-based queue if available
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -610,11 +610,15 @@ export class GroupQueue {
     if (state.messageLaneState === 'idle' || !state.messageGroupFolder) {
       return false;
     }
-    // Container was idle — mark it active again (single-method bookkeeping)
-    this.transitionMessageLaneState(chatJid, 'running');
-    // Stamp the start of this resumed run so /ipc reports the age of the
-    // current message run, not the original container-start timer.
-    state.messageRunStartedAt = Date.now();
+    // Container was idle — mark it active again (single-method bookkeeping).
+    // Only restamp the run timer on a real cooldown → running transition;
+    // follow-up sendMessage() calls while already running must NOT reset
+    // messageRunStartedAt, otherwise long-running/stuck runs get underreported
+    // on /ipc every time a new message arrives.
+    const resumed = this.transitionMessageLaneState(chatJid, 'running');
+    if (resumed) {
+      state.messageRunStartedAt = Date.now();
+    }
     const channelJid = toChannelJid(chatJid);
 
     // Use Effect-based queue if available
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -69,6 +69,18 @@ export interface GroupQueueDetail {
      * {@link deriveMessageLaneReasonFromDetail} when missing.
      */
     reason?: MessageLaneReason;
+    /**
+     * Epoch ms when the current message run started. Null when the lane is
+     * idle (cooldown or no work). Operators use this with {@link runningMs}
+     * to spot stuck or long-running message runs.
+     */
+    startedAt?: number | null;
+    /**
+     * Milliseconds elapsed since the current message run started. Null when
+     * the lane is idle. Derived as `Date.now() - startedAt` in
+     * {@link GroupQueue.getDetailedStats}.
+     */
+    runningMs?: number | null;
   };
   taskLane: {
     active: boolean;
@@ -158,6 +170,8 @@ interface GroupState {
   messageContainerName: string | null;
   messageGroupFolder: string | null;
   messageBackend: AgentBackend | null;
+  /** Epoch ms when the current message run started; null when idle. */
+  messageRunStartedAt: number | null;
   retryCount: number;
 
   // Task lane
@@ -262,6 +276,7 @@ export class GroupQueue {
         messageContainerName: null,
         messageGroupFolder: null,
         messageBackend: null,
+        messageRunStartedAt: null,
         retryCount: 0,
 
         taskActive: false,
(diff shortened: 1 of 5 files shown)


src/group-queue.ts
```diff
@@ -946,6 +963,11 @@ export class GroupQueue {
           pendingCount: state.pendingMessageJids.length,
           containerName: state.messageContainerName,
           reason: messageReason,
+          startedAt: state.messageRunStartedAt,
+          runningMs:
```
Peyton-Spencer, line 977: This reports an age whenever `messageRunStartedAt` is set, but `notifyIdle()` transitions the message lane to `cooldown` without clearing that field. In the normal long-lived-container path, `/ipc` will keep showing a growing `.lane-age` while the lane is `cooling-down`/idle-waiting, and a later `sendMessage()` from cooldown will continue the original container-start timer instead of the current message run. Clear `messageRunStartedAt` when entering cooldown and set it when `sendMessage()` moves cooldown back to running, or gate `runningMs` on `messageLaneState === 'running'` if the intended metric is active processing age.

Peyton-Spencer:
Review note: I left an inline comment on `src/group-queue.ts` for one behavior issue.

The new timer is set in `startMessageRun()`, but `notifyIdle()` can move the lane into `cooldown` without clearing it. That means `/ipc` can show a growing `.lane-age` while the container is idle-waiting, and the next `sendMessage()` from cooldown continues the original timer rather than measuring the current message run.

Targeted verification passed locally:

```bash
bun test src/group-queue.test.ts src/web/ipc-inspector.test.ts
```

GitHub is authenticated here as the PR author, so it rejected a formal request-changes review. Treat the inline comment as my requested fix before merge.

src/group-queue.ts
Peyton-Spencer: Fixed in e969262.

- `notifyIdle()` now clears `messageRunStartedAt` when entering cooldown.
- `sendMessage()` restamps `messageRunStartedAt` when it moves the lane back from cooldown → running.
- Added a regression test (`clears startedAt/runningMs while in cooldown and restamps when sendMessage resumes`) that drives the full enqueue → register → notifyIdle → sendMessage cycle and asserts the timer transitions: number → null (cooldown) → number (resume).

Full suite: 2027 pass, 0 fail.

src/group-queue.ts
```diff
@@ -592,6 +612,9 @@ export class GroupQueue {
     }
     // Container was idle — mark it active again (single-method bookkeeping)
     this.transitionMessageLaneState(chatJid, 'running');
+    // Stamp the start of this resumed run so /ipc reports the age of the
```
Peyton-Spencer: The cooldown leak is fixed, but this now restamps on every `sendMessage()` call, including when the lane was already `running` and `transitionMessageLaneState()` is a no-op. That can underreport long-running/stuck active runs whenever a follow-up message arrives. Since this comment says the stamp is for a resumed run, gate it on the transition result, e.g. `const resumed = this.transitionMessageLaneState(chatJid, 'running'); if (resumed) state.messageRunStartedAt = Date.now();`, and add a regression that an already-running `sendMessage()` preserves the original `startedAt`.

Peyton-Spencer:
Re-reviewed commit `e969262`.

The original cooldown-age leak is fixed, but I left one follow-up inline comment: `sendMessage()` now restamps `messageRunStartedAt` even when the lane was already `running`, because `transitionMessageLaneState(chatJid, 'running')` can be a no-op. That can underreport a genuinely long-running active lane if follow-up messages arrive.

Suggested fix is small: only restamp when the transition actually happened, and add a regression that already-running `sendMessage()` preserves the original `startedAt`.

Targeted verification still passes locally:

```bash
bun test src/group-queue.test.ts src/web/ipc-inspector.test.ts
```

Peyton-Spencer:
## Mergemaster deferral — PR #688

This PR has an **unresolved review concern** from the author themselves at `src/group-queue.ts:615` (posted 2026-05-11T00:13:43Z, after the most recent commit `e969262`):

> The cooldown leak is fixed, but this now restamps on every `sendMessage()` call, including when the lane was already `running` and `transitionMessageLaneState()` is a no-op. That can underreport long-running/stuck active runs whenever a follow-up message arrives.

The proposed gate (`const resumed = this.transitionMessageLaneState(...)` and only restamp when `resumed` indicates a real cooldown → running transition) has not been applied.

**Action:** filing a small follow-up issue so we ship a real fix and don't merge a PR that the author has already flagged as having a regression on the very metric it's adding. Will revisit once the follow-up commit lands and CI is re-green.

— Clayton, acting mergemaster

Peyton-Spencer:
Filed follow-up tracking issue #696 capturing the restamp-on-no-op-transition fix. This PR will be picked back up by the next merge window once #696's fix is committed and CI re-greens.

Peyton-Spencer:
Pushed follow-up fix for #696.

- sendMessage() now only restamps `messageRunStartedAt` when `transitionMessageLaneState()` returns true (real cooldown → running transition).
- Follow-up messages on an already-running lane no longer reset the run timer, so /ipc keeps reporting accurate ages for long-running/stuck runs.
- Added regression test that snapshots startedAt, sleeps to advance Date.now(), then asserts sendMessage() leaves startedAt unchanged.
- Existing cooldown → resume test still passes; bun test 2028/2028 and `tsc --noEmit` clean locally.

Acceptance criteria from #696 satisfied; safe to merge once CI is green.

Peyton-Spencer:
Technical readiness sign-off for mergemaster.

Checks reviewed:
- CI is green: format, test, typecheck, validate-skill-only, CodeRabbit.
- Diff scope is limited to message-lane running-age plumbing for `/ipc`: `GroupQueueDetail.messageLane.startedAt/runningMs`, IPC inspector rendering, and focused tests.
- Verified the prior timing concern is addressed in this PR: `notifyIdle()` clears `messageRunStartedAt`, `sendMessage()` restamps only on a real cooldown -> running transition, and follow-up sends while already running do not reset the timer.

Risk note: low to moderate. This touches queue runtime bookkeeping, but the state changes are local to telemetry fields and covered by targeted regression tests for active, cooldown, resume, and already-running paths.

No blocking findings from my review.

PR #709, opened 2026-05-17
feat: add deterministic scheduled task preprocessors

Peyton-Spencer:
## Summary

Adds deterministic preprocessing for scheduled tasks so recurring workflows can do cheap TypeScript triage before spending agent tokens.

- adds optional `preprocess_script` on scheduled tasks, persisted in SQLite and exposed through IPC, MCP `schedule_task` / `update_task`, snapshots, and the web API
- runs JS/TS preprocessors before agent dispatch; scripts live under the owning group workspace at `/workspace/group/task-workflows/` from the agent's perspective
- supports preprocessor decisions to skip no-op runs or augment/replace the prompt before agent launch
- records skipped preprocessor runs as successful task runs and records preprocessor failures as blocked task runs without starting the agent container
- documents the workflow contract in `docs/SPEC.md`

## Workflow contract

The scheduler sends JSON on stdin with task metadata, repo root, workflows dir, and current timestamp. The workflow writes one JSON object to stdout:

```json
{ "action": "skip", "reason": "no MCP diff" }
```

or:

```json
{ "action": "run", "promptPrefix": "MCP changed; sync connector packages." }
```

## Verification

- `bun test` — 2073 pass / 0 fail
- `bunx tsc --noEmit` — pass
- `bunx prettier --check .` — pass
- `bun test src/task-preprocessor.test.ts src/db.test.ts src/ipc-processing.test.ts src/task-scheduler-runner.test.ts src/web/routes.test.ts` — 169 pass / 0 fail

## Release Notes

* **New Features**
  * Tasks can now include an optional preprocessing workflow that executes before the agent starts, enabling duplicate detection and prompt modification through deterministic triage.
  * Added `TASK_WORKFLOWS_DIR` configuration option to specify the directory containing preprocessing workflow scripts.

* **Documentation**
  * Updated task scheduling documentation to describe the preprocessing workflow behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/omniaura/omniclaw/pull/709?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/container/agent-runner/src/ipc-mcp-stdio.ts
+++ b/container/agent-runner/src/ipc-mcp-stdio.ts
@@ -16,6 +16,9 @@ const MESSAGES_DIR = path.join(IPC_DIR, 'messages');
 const TASKS_DIR = path.join(IPC_DIR, 'tasks');
 const RESPONSES_DIR = path.join(IPC_DIR, 'responses');
 const USER_REGISTRY_PATH = path.join(IPC_DIR, 'user_registry.json');
+const taskWorkflowsDir =
+  process.env.OMNICLAW_TASK_WORKFLOWS_DIR || 'task-workflows';
+const taskWorkflowsPath = `/workspace/group/${taskWorkflowsDir.replace(/^\/+/, '').replace(/\/+$/, '')}/`;
 
 // Context from environment variables (set by the agent runner)
 const initialChatJid = process.env.OMNICLAW_CHAT_JID!;
@@ -288,7 +291,7 @@ MESSAGING BEHAVIOR - The task agent's output is sent to the user or group. It ca
 \u2022 Only send a message when there's something to report (e.g., "notify me if...")
 \u2022 Never send a message (background maintenance tasks)
 
-DETERMINISTIC PREPROCESSING - For recurring maintenance tasks that can be cheaply triaged in code, write a TypeScript workflow file under /workspace/group/task-workflows/ and pass preprocess_script as its relative path. The workflow runs before the agent and can return {"action":"skip","reason":"no diff"} or {"action":"run","promptPrefix":"..."} as JSON on stdout. Use this to avoid spending agent tokens on no-op checks.
+DETERMINISTIC PREPROCESSING - For recurring maintenance tasks that can be cheaply triaged in code, write a TypeScript workflow file under ${taskWorkflowsPath} and pass preprocess_script as its relative path. The script receives JSON on stdin: { task, repoRoot, workflowsDir, now }. It runs with Bun on the host using a minimal env, so it should not rely on API tokens. Write one JSON result as the last stdout line or prefixed with OMNICLAW_TASK_PREPROCESSOR_RESULT=: {"action":"skip","reason":"no diff"}, {"action":"run","promptPrefix":"..."}, or {"action":"error","message":"..."}. Use this to avoid spending agent tokens on no-op checks.
 
 IMPORTANT: When MODIFYING an existing task, use update_task — it edits the task in place while preserving its ID and run history. Only use cancel_task to destructively delete a task the user no longer wants. Unlike pausing, the task cannot be restored on deletion.
 
@@ -306,7 +309,7 @@ SCHEDULE VALUE FORMAT (all times are LOCAL timezone):
       .string()
       .optional()
       .describe(
-        'Optional relative path under /workspace/group/task-workflows/ to a JS/TS workflow that runs before the agent and returns JSON to skip or augment the prompt.',
+        `Optional relative path under ${taskWorkflowsPath} to a JS/TS workflow that runs before the agent and returns JSON to skip or augment the prompt.`,
       ),
     schedule_type: z
       .enum(['cron', 'interval', 'once'])
@@ -510,7 +513,7 @@ Examples:
       .nullable()
       .optional()
       .describe(
-        'Set or clear the relative /workspace/group/task-workflows/ script that runs before the agent. Pass null to clear it.',
+        `Set or clear the relative ${taskWorkflowsPath} script that runs before the agent. Pass null to clear it.`,
       ),
     schedule_type: z.enum(['cron', 'interval', 'once']).optional(),
     schedule_value: z
--- a/container/agent-runner/src/ipc-mcp-stdio.ts
+++ b/container/agent-runner/src/ipc-mcp-stdio.ts
@@ -288,6 +288,8 @@ MESSAGING BEHAVIOR - The task agent's output is sent to the user or group. It ca
 \u2022 Only send a message when there's something to report (e.g., "notify me if...")
 \u2022 Never send a message (background maintenance tasks)
 
+DETERMINISTIC PREPROCESSING - For recurring maintenance tasks that can be cheaply triaged in code, write a TypeScript workflow file under /workspace/group/task-workflows/ and pass preprocess_script as its relative path. The workflow runs before the agent and can return {"action":"skip","reason":"no diff"} or {"action":"run","promptPrefix":"..."} as JSON on stdout. Use this to avoid spending agent tokens on no-op checks.
+
 IMPORTANT: When MODIFYING an existing task, use update_task — it edits the task in place while preserving its ID and run history. Only use cancel_task to destructively delete a task the user no longer wants. Unlike pausing, the task cannot be restored on deletion.
 
 SCHEDULE VALUE FORMAT (all times are LOCAL timezone):
@@ -299,6 +301,12 @@ SCHEDULE VALUE FORMAT (all times are LOCAL timezone):
       .string()
       .describe(
         'What the agent should do when the task runs. For isolated mode, include all necessary context here.',
+      ),
+    preprocess_script: z
+      .string()
+      .optional()
+      .describe(
+        'Optional relative path under /workspace/group/task-workflows/ to a JS/TS workflow that runs before the agent and returns JSON to skip or augment the prompt.',
       ),
     schedule_type: z
       .enum(['cron', 'interval', 'once'])
@@ -390,6 +398,7 @@ SCHEDULE VALUE FORMAT (all times are LOCAL timezone):
     const data = {
       type: 'schedule_task',
       prompt: args.prompt,
+      preprocess_script: args.preprocess_script,
       schedule_type: args.schedule_type,
       schedule_value: args.schedule_value,
       context_mode: args.context_mode || 'group',
@@ -443,6 +452,7 @@ server.tool(
             id: string;
             groupFolder?: string;
             prompt: string;
+            preprocess_script?: string | null;
             schedule_type: string;
             schedule_value: string;
             status: string;
@@ -453,7 +463,10 @@ server.tool(
               owner === groupFolder ? `${owner} (yours)` : owner;
             const promptPreview =
               t.prompt.length > 50 ? `${t.prompt.slice(0, 50)}...` : t.prompt;
-            return `- [${t.id}] (owner: ${ownerLabel}) ${promptPreview} (${t.schedule_type}: ${t.schedule_value}) - ${t.status}, next: ${t.next_run || 'N/A'}`;
+            const preprocessor = t.preprocess_script
+              ? `, preprocess: ${t.preprocess_script}`
+              : '';
+            return `- [${t.id}] (owner: ${ownerLabel}) ${promptPreview} (${t.schedule_type}: ${t.schedule_value}${preprocessor}) - ${t.status}, next: ${t.next_run || 'N/A'}`;
           },
         )
         .join('\n');
@@ -485,12 +498,20 @@ server.tool(
 
 Examples:
 - update_task({ task_id: "...", prompt: "new instructions" })
+- update_task({ task_id: "...", preprocess_script: "sync-connectors.ts" })
 - update_task({ task_id: "...", schedule_type: "cron", schedule_value: "0 9 * * *" })
 - update_task({ task_id: "...", status: "paused" })
 - update_task({ task_id: "...", status: "active" })`,
   {
     task_id: z.string().describe('ID of the task to update'),
     prompt: z.string().optional().describe('New instructions for the task'),
+    preprocess_script: z
+      .string()
+      .nullable()
+      .optional()
+      .describe(
+        'Set or clear the relative /workspace/group/task-workflows/ script that runs before the agent. Pass null to clear it.',
+      ),
     schedule_type: z.enum(['cron', 'interval', 'once']).optional(),
     schedule_value: z
       .string()
@@ -510,6 +531,7 @@ Examples:
   async (args) => {
     if (
       !args.prompt &&
+      args.preprocess_script === undefined &&
       !args.schedule_type &&
       !args.schedule_value &&
       !args.status &&
@@ -519,7 +541,7 @@ Examples:
         content: [
           {
             type: 'text' as const,
-            text: 'At least one of prompt, schedule_type, schedule_value, or status must be provided.',
+            text: 'At least one of prompt, preprocess_script, schedule_type, schedule_value, status, or context_mode must be provided.',
           },
         ],
         isError: true,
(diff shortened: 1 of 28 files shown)


src/task-preprocessor.ts
```diff
@@ -0,0 +1,186 @@
...
+  const child = spawnSync(process.execPath, ['run', workflowPath], {
+    cwd: process.cwd(),
+    input: JSON.stringify(input),
+    encoding: 'utf8',
+    env: {
+      ...process.env,
+      OMNICLAW_TASK_PREPROCESSOR: '1',
+      OMNICLAW_TASK_ID: task.id,
+    },
+    timeout: options.timeoutMs ?? DEFAULT_TIMEOUT_MS,
```
Peyton-Spencer: *Security:* spreading `process.env` into the workflow subprocess leaks the orchestrator's full secret bag (GITHUB_TOKEN, ANTHROPIC_API_KEY, B2 keys, etc.). `CLAUDE.md` explicitly forbids this pattern ("never pass `{ ...process.env }` to subprocesses — use explicit allowlists"). Suggest an allowlist like `PATH`, `HOME`, `TZ`, `LANG` plus a `TASK_PREPROCESS_ENV_ALLOWLIST` knob, and surface the leak risk in `SECURITY.md`.

src/task-preprocessor.ts
```diff
@@ -0,0 +1,186 @@
...
+      last_outcome_state: task.last_outcome_state ?? null,
+      last_outcome_reason: task.last_outcome_reason ?? null,
+    },
+    repoRoot: process.cwd(),
+    workflowsDir,
+    now: (options.now ?? (() => new Date()))().toISOString(),
+  };
+
+  const child = spawnSync(process.execPath, ['run', workflowPath], {
+    cwd: process.cwd(),
```
Peyton-Spencer, line 245: *Trust boundary:* the workflow runs on the host with the orchestrator's `cwd` and uid. Combined with the env leak above, any agent that drops a `.ts` file under `/workspace/group/task-workflows/` can effectively run code on the host as omniclaw. Two safer options: (1) run the preprocessor inside the agent container (already isolated, has its own bun, has a clean env), or (2) execute under a restricted cwd / `--cwd` set to `workflowsDir` and document this as a privileged surface.

src/task-preprocessor.ts
```diff
@@ -0,0 +1,186 @@
...
+      last_result: task.last_result,
+      last_outcome_state: task.last_outcome_state ?? null,
+      last_outcome_reason: task.last_outcome_reason ?? null,
+    },
+    repoRoot: process.cwd(),
+    workflowsDir,
+    now: (options.now ?? (() => new Date()))().toISOString(),
+  };
+
+  const child = spawnSync(process.execPath, ['run', workflowPath], {
```
Peyton-Spencer: `process.execPath` will be `node` if omniclaw is started under Node (and we still ship `package.json` scripts that work either way). `node run …` is not valid. Either resolve `bun` from PATH (and error clearly if missing), or assert that the process was launched with Bun at startup.

src/task-preprocessor.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  const groupDir = path.resolve(GROUPS_DIR, task.group_folder);
+  assertPathWithin(groupDir, GROUPS_DIR, 'task workflow group folder');
+  return path.resolve(groupDir, TASK_WORKFLOWS_DIR);
+}
+
+function parseDecision(stdout: string): TaskPreprocessDecision {
+  const trimmed = stdout.trim();
+  if (!trimmed) return { action: 'run' };
+  const parsed = JSON.parse(trimmed) as unknown;
```
Peyton-Spencer, line 137: `JSON.parse(trimmed)` will throw if the script writes anything before the JSON (a stray `console.log('hi')` is the common dev pitfall). Consider extracting the last non-empty line and parsing that, or using a sentinel like `OMNICLAW_DECISION=<json>` to decouple from incidental stdout.

src/task-preprocessor.ts
```diff
@@ -0,0 +1,186 @@
...
+    const decision = parseDecision(child.stdout ?? '');
+    if (decision.action === 'skip') {
+      return { action: 'skip', reason: decision.reason ?? 'no work' };
+    }
+    const prompt = decision.prompt ?? task.prompt;
+    return {
+      action: 'run',
+      prompt: decision.promptPrefix
+        ? `${decision.promptPrefix.trim()}\n\n${prompt}`
+        : prompt,
```
Peyton-Spencer, line 311: No cap on `promptPrefix` length. A buggy or compromised workflow could prepend megabytes and blow agent context / token budget — the exact failure mode this feature exists to prevent. Suggest a hard limit (e.g. 8KB) with a clear error if exceeded.

src/task-preprocessor.ts
```diff
@@ -0,0 +1,186 @@
...
+  now: string;
+}
+
+export interface TaskPreprocessorOptions {
+  workflowsDir?: string;
+  timeoutMs?: number;
+  now?: () => Date;
+}
+
+const DEFAULT_TIMEOUT_MS = 60_000;
```
Peyton-Spencer: Hardcoded `DEFAULT_TIMEOUT_MS`. Other timeouts in `config.ts` are env-configurable (`CONTAINER_TIMEOUT`, `IDLE_TIMEOUT`, etc.) — please add `TASK_PREPROCESS_TIMEOUT_MS` for parity.

src/task-scheduler.ts
```diff
@@ -259,12 +262,91 @@ async function runTask(
...
+      const durationMs = Date.now() - startTime;
+      const result = `Skipped by preprocessor: ${preprocessResult.reason}`;
+      runtime.logTaskRun({
+        task_id: task.id,
+        run_at: runAt,
+        duration_ms: durationMs,
+        status: 'success',
+        result,
+        error: null,
+        outcome_state: 'done',
```
Peyton-Spencer: Skipping is recorded as `outcome_state: 'done'`. That will inflate the success count and make dashboards/PM deltas unable to tell a real run from a no-op. Suggest a new `'skipped'` outcome state (already alluded to in `TaskOutcomeState`) with its own UI badge.

src/task-scheduler.ts
```diff
@@ -259,12 +262,91 @@ async function runTask(
...
+        outcome_reason: preprocessResult.reason,
+      });
+      return;
+    }
+
+    if (preprocessResult.action === 'error') {
+      const durationMs = Date.now() - startTime;
+      const error = `Preprocessor failed: ${preprocessResult.error}`;
```
Peyton-Spencer, line 309: `error` is built from raw `child.stderr` (in `task-preprocessor.ts`). Stack traces or accidental `console.log(process.env.GITHUB_TOKEN)` will flow into `last_result`, handoffs, and the web API. Consider truncating to ~512 chars and stripping ANSI/control chars before persisting.

Peyton-Spencer (commented):
Nice feature — solid scaffolding (DB column, migrations, IPC, MCP, web API, docs, tests). My main concern is the trust boundary: preprocessor scripts execute on the host orchestrator with full env and filesystem access, not inside the agent container. That deserves explicit hardening or, at minimum, a documented trust contract.

Grouped feedback below; inline comments cover specifics.

*Security (must address):*
- Host execution with full `process.env` propagation (GITHUB_TOKEN, ANTHROPIC keys, B2 creds, etc.). Per `CLAUDE.md` security guidance: never spread `process.env` to subprocesses — use an explicit allowlist.
- Preprocessor `cwd` is the orchestrator repo root, giving the script read/write across the host. An agent that writes a workflow file effectively gains host code execution. Either sandbox (run inside the existing container with the preprocessor mount), or call out the trust boundary loudly in `SPEC.md` + `SECURITY.md`.

*Correctness:*
- `process.execPath` assumes Bun launched omniclaw. `node run /workflow.ts` will fail. Either resolve `bun` from PATH or assert/document the runtime.
- `parseDecision` requires the entire stdout to be one JSON object. Any `console.log` in the workflow breaks parsing. Consider taking the last non-empty line, or use a separate channel (sentinel file / stderr-stripped marker).
- No size cap on `promptPrefix` / `prompt`; a runaway prefix could balloon agent tokens — the opposite of this feature's purpose.

*UX / observability:*
- Skipped runs persist as `outcome_state: 'done'` with `Skipped by preprocessor: …` in `last_result`. They will be indistinguishable from real success in dashboards and PM deltas. A dedicated `'skipped'` outcome state would make queue metrics honest.
- Preprocessor error stderr is dumped verbatim into `last_result` / handoff — consider trimming + redacting (stack traces or accidental secret prints).
- Timeout hardcoded to 60s. Make it env-configurable (e.g. `TASK_PREPROCESS_TIMEOUT_MS`) to match the rest of the timeout knobs.

*Quality of life:*
- Auto-`mkdir -p` the workflows dir the first time it's referenced, or surface a clearer 'create the workflow first' error.
- Surface registered workflow files in snapshots / via a `list_task_workflows` MCP tool so agents can introspect what already exists before overwriting.

*Tests to add:*
- timeout path (script that sleeps past `timeoutMs`)
- maxBuffer overflow
- stdout containing log noise before the JSON object
- env-leak guard (assert that, post-hardening, `OMNICLAW_TASK_ID` and `OMNICLAW_TASK_PREPROCESSOR` are set but secrets are not propagated)

Happy to chat on any of these. Overall direction is great — this unblocks the daily MCP-diff use case Peyton called out and the surface area through MCP/IPC/web is consistent.

container/agent-runner/src/ipc-mcp-stdio.ts
```diff
@@ -300,6 +302,12 @@ SCHEDULE VALUE FORMAT (all times are LOCAL timezone):
       .describe(
         'What the agent should do when the task runs. For isolated mode, include all necessary context here.',
       ),
+    preprocess_script: z
+      .string()
+      .optional()
+      .describe(
+        'Optional relative path under /workspace/group/task-workflows/ to a JS/TS workflow that runs before the agent and returns JSON to skip or augment the prompt.',
```
Peyton-Spencer: The MCP description tells agents *what* to put in the workflow but not the contract details (stdin schema, stdout shape, allowed file extensions, env restrictions, available globals). Worth either inlining a 6–8 line contract block here or linking the SPEC section. Agents will otherwise reinvent the schema each time.

src/web/routes.ts
```diff
@@ -542,6 +542,11 @@ async function handleCreateTask(
     group_folder: group_folder as string,
     chat_jid: chat_jid as string,
     prompt: prompt as string,
+    preprocess_script:
+      typeof body.preprocess_script === 'string' &&
+      body.preprocess_script.trim()
+        ? body.preprocess_script.trim()
+        : null,
```
Peyton-Spencer: Web API accepts arbitrary `preprocess_script` strings without checking existence or extension. The path traversal/extension check happens at run time in `task-preprocessor.ts`, but creators get no immediate feedback. Worth validating extension here so users can't persist a task that will deterministically fail at run time.

src/task-preprocessor.test.ts
```diff
@@ -0,0 +1,88 @@
...
+});
+
+it('does nothing when no preprocessor is configured', () => {
+  expect(
+    runTaskPreprocessor(task({ preprocess_script: null }), { workflowsDir }),
+  ).toEqual({
+    action: 'run',
+    prompt: 'sync connector packages',
+  });
+});
```
Peyton-Spencer, line 216: Missing test coverage: (a) timeout path (script that sleeps), (b) `maxBuffer` overflow, (c) stdout with noise before JSON, (d) env-leak assertion once the allowlist lands. These would lock in the security contract and prevent regressions.

Peyton-Spencer:
Pushed review fixes in `daadcb86ed3a805acc4397b5bf64933a0ef1cd2f`.

Addressed the review blockers:
- preprocessor subprocess now uses a minimal env allowlist instead of inheriting host secrets
- Bun is invoked via `bun run`, with `TASK_PREPROCESS_TIMEOUT_MS` for timeout config
- stdout parsing supports noisy logs via last JSON line or `OMNICLAW_TASK_PREPROCESSOR_RESULT=` sentinel
- skip runs now record `outcome_state: skipped`
- prompt/promptPrefix fragments are capped before dispatch
- stderr/errors are sanitized and truncated before persistence
- web + IPC validate `preprocess_script` extension/path up front
- added migration 5 so databases already stamped at version 4 get `scheduled_tasks.preprocess_script`
- documented the privileged preprocessor surface in `docs/SECURITY.md`

Verification from a clean worktree:
- `bun test src/task-preprocessor.test.ts src/task-scheduler-runner.test.ts src/task-outcome.test.ts src/web/routes.test.ts src/db.migrations.test.ts` — 94 pass
- `bunx tsc --noEmit` — pass
- `bunx prettier --check .` — pass
- `bun test` — 2083 pass

CI is queued on the pushed commit. @Clayton please re-review when checks land.

Peyton-Spencer (commented):
Re-reviewed `daadcb8`. Every blocker from the first pass is addressed; the surface is meaningfully tighter than the first revision.

*Verified:*
• Env allowlist — `buildPreprocessorEnv` returns only `PATH`, `HOME`, `TMPDIR`, `TEMP`, `TMP`, `TZ`, `LANG`, `LC_ALL` plus the two OmniClaw markers. No secrets propagated. Locked in with a regression test in `task-preprocessor.test.ts`.
• Runtime — `spawnSync('bun', ['run', workflowPath], …)` resolves `bun` from `PATH` rather than assuming `process.execPath`.
• `parseDecision` — supports the `OMNICLAW_TASK_PREPROCESSOR_RESULT=` sentinel and falls back to last non-empty line. Workflow `console.log` noise no longer breaks parsing.
• Skipped runs — new `'skipped'` `TaskOutcomeState` plumbed through `VALID_OUTCOME_STATES`, scheduler, protocol types, MCP outcome surface, and dashboard color (`page-scripts.ts:1184` adds `var(--text-dim)`).
• Prompt-fragment caps — `capPromptFragment` enforces `DEFAULT_MAX_PROMPT_FRAGMENT_CHARS = 8_000` for both `prompt` and `promptPrefix`, with a clear truncation marker.
• Timeout — `TASK_PREPROCESS_TIMEOUT_MS` env var added, threaded through, and the `ETIMEDOUT` branch produces a sanitized message.
• Error sanitization — `sanitizePreprocessorMessage` strips ANSI/control chars, collapses whitespace, hard-caps at 512 chars. Applied to stderr, internal errors, and explicit `{action:'error'}` messages.
• Web + IPC validation — `normalizePreprocessScriptPath` is now the single source of truth for path/extension checks and is called from `src/web/routes.ts`, `src/ipc.ts` (both create and update branches), and at execution time.
• Migration 5 — `addColumnIfNotExists(scheduled_tasks, preprocess_script, TEXT)`, baseline bumped to 5. Test added in `db.migrations.test.ts`. Idempotent on already-migrated databases.
• Trust boundary documented — new `docs/SECURITY.md` section 6 calls out the privileged host surface, env allowlist, extension allowlist, traversal guards, sanitization, and timeout knob.
• MCP description rewritten — now explicitly notes the host execution, missing API tokens, and the new sentinel/JSON contract. Workflows-dir env var (`OMNICLAW_TASK_WORKFLOWS_DIR`) is plumbed to both runtimes.

*Tests*: `task-preprocessor.test.ts` grew by ~130 lines covering log noise, sentinel parsing, env leak, timeout, max-buffer overflow, promptPrefix capping, and explicit `'error'` decisions. `task-outcome.test.ts` covers the new skipped passthrough. `db.migrations.test.ts` covers v4→v5.

*Minor follow-ups (non-blocking, file as a tracking issue if you want):*
• If a user sets `OMNICLAW_TASK_WORKFLOWS_DIR=/some/abs/path`, the MCP description renders as `/workspace/group/some/abs/path/` (the leading slash is stripped before interpolation). The host-side resolver honors `path.isAbsolute()` in `defaultWorkflowsDir`, so execution still works, but the prompt shown to the agent will be wrong. Worth either disallowing absolute values in the MCP-side render, or rendering them verbatim.
• `cwd: process.cwd()` is unchanged — still the orchestrator repo root. The env allowlist closes the main concern, but a workflow could still read host files under cwd. SECURITY.md doesn't explicitly call out the cwd; worth adding one sentence so operators know the script gets read access to the orchestrator's working directory.
• `OMNICLAW_TASK_PREPROCESSOR_RESULT=` is a fine sentinel; consider also documenting it in `SPEC.md` (the SPEC mentions "last stdout line" but not the sentinel form explicitly).

Approving direction — feature is in good shape to land once you're happy with the follow-up items above (none of which I'd block on).

PR #713, opened 2026-05-17
Remove split execution sidecar

### LINKED ISSUE #710: Consider removing the execution broker / split-execution sidecar
## Current state

When `EXEC_CONTAINER_NAME` is set, the container's `/usr/local/bin/bash` is a shim (`container/exec-shim.sh`) that forwards every `bash -c "..."` to a second \"exec sidecar\" container:

- **Docker:** via `docker exec` (low-latency, in-kernel)
- **Apple Container:** via filesystem-polling IPC — JSON requests in `/workspace/ipc/exec-requests/`, host-side broker in `src/backends/local-backend.ts:583–1290` watches and dispatches via `container exec`, response files come back, 50ms poll loop

Stated rationale (from the shim's header comment): \"separates heavy workloads (builds, browsers, tests) from the agent runtime to prevent resource starvation.\"

## Why this is worth reconsidering

1. **The isolation guarantee is leaky.** Only callers that invoke `/bin/bash` go through the shim. That covers the Claude Agent SDK's Bash tool, but **Codex's `exec_command`/`write_stdin` and OpenCode's shell tool bypass it entirely** — they spawn subprocesses directly. A runaway `cargo build` from Codex already starves the agent runtime regardless of the broker, so the invariant only holds when the runtime happens to be Claude.

2. **Resource cost compounds with the container model.** OmniClaw spawns one container per `(agent, channel)` dispatch (folder = `<agent>__dispatch__<sha1(channel+agent)[0:16]>`, default 4G via `CONTAINER_MEMORY`). With split execution on, each dispatch adds a 4G sidecar. One agent active in 6 channels → 6 agent containers + 6 sidecars, ~48G ceiling.

3. **Maintenance surface.** ~500 LOC of broker plumbing in `local-backend.ts`, the shim script, the dual Docker vs apple-container codepaths, sidecar lifecycle management (spawn/stop/cleanup on timeout, error, normal exit), and a 50ms filesystem-polling tax per shell call on macOS.

## Options

- **(a) Delete.** Remove the shim, the broker, the `EXEC_CONTAINER_NAME` plumbing, the sidecar container lifecycle. Bash from the SDK runs in the agent container alongside Codex/OpenCode shell calls — consistent across runtimes.
- **(b) Keep but bump `CONTAINER_MEMORY`** instead of running two containers. Single container with a higher ceiling is simpler than two with lower ceilings.
- **(c) Replace with per-process limits.** If specific tools (Playwright, large builds) are the actual concern, wrap *those* with cgroup limits rather than routing all bash through a sidecar.

## Open questions before deciding

- Was there a specific incident that motivated split execution — Playwright OOMing the SDK, a big repo build knocking the agent VM over? If yes, picking the replacement (b or c) needs to come before deletion.
- If no concrete incident — i.e. this was speculative future-proofing — option (a) is a clean win.

Peyton-Spencer:
## Summary

- removes split-execution sidecar support from the local backend
- removes the bash shim and Docker CLI dependency from the agent image
- drops `SPLIT_EXECUTION` / `EXEC_CONTAINER_MEMORY` config and stale docs/tests

Closes #710.

## Verification

- `bun test`
- `bun test src/backends/local-backend.test.ts`
- `bunx tsc --noEmit`
- `bunx prettier --check .`
- `git diff --check`


.env.example
```diff
@@ -17,21 +17,14 @@ DISCOVERY_TRUST_LAN_ADMIN=false
...
-
 # Container lifecycle
 # Defaults to 0 so idle message containers exit and future messages resume the session.
 # Increase to keep a warm idle pool for lower first-token latency.
 # MAX_IDLE_CONTAINERS=0
 
 # Shared Claude VM: consolidate all Claude agent-runner processes into a single
 # long-running Apple Container VM. Each agent runs via `container exec` into
-# the shared VM. Exec containers remain per-agent for isolation.
+# the shared VM.
```
Peyton-Spencer, line 27: Nit: while we're touching this section, the updated SHARED_CLAUDE_VM comment still says "Reduces disk usage from 2N to N+1 containers" — that 2N number came from agent + exec sidecar. With split-exec gone the baseline is already N agent containers, so the new ratio is N → 1+shared-vm. Worth rephrasing so it doesn't reference the deleted topology.

container/Dockerfile.base
```diff
@@ -137,13 +137,6 @@ RUN ARCH="$(dpkg --print-architecture)" \
 ARG CODEX_VERSION=0.130.0
 RUN npm install -g "@openai/codex@${CODEX_VERSION}"
```
Peyton-Spencer, line 138: Image diff: removing the Docker CLI install (~50MB) plus the bash-shim layer is a nice secondary win. Worth a CHANGELOG note so operators know to rebuild rather than pull-and-pray on a stale layer cache.

Peyton-Spencer (commented):
Clean deletion — verified the scope matches the issue and there are no dangling references:

• `grep` for `SPLIT_EXECUTION`, `EXEC_CONTAINER_NAME`, `exec-shim`, `bash.real`, and all removed exports (`buildExecInvocationArgs`, `runExecBrokerRequest`, `ExecBrokerOutputLimitError`, etc.) returns zero hits across `src/`, `container/`, and `docs/`.
• `SHARED_CLAUDE_VM` comment correctly updated to drop "Exec containers remain per-agent".
• Roadmap entries trimmed (including the now-orphaned reference to #622, which has been merged so that line was already obsolete).
• Dockerfile drops the bash-shim install and the Docker CLI download (~50MB image shrink — nice secondary win).
• `buildVolumeMounts` no longer creates `exec-requests/` and `exec-responses/` under the IPC mount.
• Tests trimmed correspondingly; no orphaned imports.

Approving. Small operational nits below — none block merge.

*Operational notes worth a CHANGELOG entry:*
• Operators with `SPLIT_EXECUTION=true` or `EXEC_CONTAINER_MEMORY=8G` in their `.env` will silently lose those values (Zod will just not parse them, no warning). Either a one-line startup warning for known-deprecated keys, or a CHANGELOG entry calling out the removal so operators know to clean up `.env` after pulling.
• Published `omniclaw-agent:latest` will be ~50MB smaller without the Docker CLI binary and the bash-shim layer is gone. Users with stale local images still work (the shim's `EXEC_CONTAINER_NAME` guard falls through to `bash.real`), but a clean rebuild is recommended.
• On macOS hosts running this orchestrator update against an existing data dir, the per-group `exec-requests/` and `exec-responses/` IPC subdirs from previous runs will linger empty. Harmless, but a single-shot cleanup pass on startup would be tidy.
• Did `Apple Container split execution landed (#399)` need to be removed from the "Completed milestones" history list too? Currently dropped from the active feature table. Historical entries usually stay even when features are removed, but if the milestone is effectively reversed, leaving it could read as misleading.

*Test coverage:* the 294-line trim drops the entire `exec broker` suite plus the `creates exec broker directories` mount test. Confirmed by grep that no other test file replaces this coverage (since the feature is gone). PR body's `2053 pass` figure matches expectation post-removal.

PR #728, opened 2026-05-19
Fix MCP current chat routing typecheck

Peyton-Spencer:
## Summary

- Import `getCurrentChatJid` in the MCP stdio server after the routing helper extraction.
- Reuse a single `routingContext` for send-message routing, scheduled task targeting, and reactions.
- Fix the no-arg `getCurrentChatJid()` calls that broke the new `agent-runner-typecheck` CI job on `main`.

## Verification

- `cd container/agent-runner && bunx tsc --noEmit`
- `bun run typecheck`
- `bunx prettier --check container/agent-runner/src/ipc-mcp-stdio.ts`
- `git diff --check`

### DIFF
--- a/container/agent-runner/src/ipc-mcp-stdio.ts
+++ b/container/agent-runner/src/ipc-mcp-stdio.ts
@@ -13,6 +13,7 @@ import { CronExpressionParser } from 'cron-parser';
 import {
   buildChannelMaps,
   buildSendMessageChannelDescription,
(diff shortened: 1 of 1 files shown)


Peyton-Spencer:
Superseded by #730 which landed an equivalent fix on main (`readCurrentChatJid()` wrapper at the call sites that broke `agent-runner-typecheck`). Closing rather than rebasing — same outcome.

PR #731, opened 2026-05-19
fix: validate Discord embed media URLs

### LINKED ISSUE #667: security: Discord embed image downloads allow SSRF
## Summary

Discord embed image previews are downloaded server-side without SSRF validation. A user can post a link that causes Discord to include an embed `image.url` or `thumbnail.url`; OmniClaw then calls `downloadBinaryAttachment(imageUrl)` and fetches that URL from the bot host.

Severity: P1 / High

## Affected Code

`src/channels/discord.ts` downloads embed images at lines 1052-1079:

```ts
const imageUrl = embed.image?.url || embed.thumbnail?.url;
...
const bytes = await downloadBinaryAttachment(imageUrl);
fs.writeFileSync(filePath, bytes);
```

`downloadBinaryAttachment()` in `src/media.ts` only applies a timeout and byte cap:

```ts
const response = await fetchWithTimeout(url, options);
return readStreamWithByteLimit(response.body, ...);
```

It does not restrict protocol, reject embedded credentials, resolve and reject private/loopback/link-local IPs, or pin DNS after validation. The safer remote-image cache path already has this logic in `src/web/image-cache.ts`, but the Discord media pipeline does not reuse it.

## Impact

Any Discord user who can trigger the bot in a registered channel can cause the OmniClaw host to fetch attacker-chosen URLs through embed image metadata. Depending on deployment, this can:

- Probe internal services reachable from the OmniClaw host or container network.
- Hit loopback-only services on the host if `localhost`, `127.0.0.1`, or private IPs are accepted by the underlying fetch path.
- Reach cloud metadata or LAN-only admin services if running in an environment where those addresses exist.
- Leak timing/status side channels through bot logs and media placeholder behavior.

The 10 MiB byte cap and 15 second timeout reduce DoS impact, but they do not prevent SSRF.

## Reproduction Sketch

1. Run OmniClaw with Discord enabled and a registered Discord channel.
2. Host an HTML page or message content that produces a Discord embed with `og:image` pointing at an internal URL, for example `http://127.0.0.1:6001/api/health` or a LAN-only HTTP service.
3. Send that link in the registered Discord channel.
4. When Discord supplies the embed image/thumbnail URL, `handleMessageCreate()` downloads it via `downloadBinaryAttachment(imageUrl)` without validation.

Expected: the bot rejects embed image URLs that resolve to loopback/private/link-local/multicast addresses and only fetches validated HTTP(S) image URLs.

Actual: the bot attempts the fetch from the OmniClaw runtime host.

## Suggested Fix

Add a shared remote media URL validation helper before any server-side fetch of user-controlled URLs:

- Allow only `http:` and `https:`.
- Reject embedded credentials.
- Reject `localhost`, loopback, private, link-local, multicast, carrier-grade NAT, and IPv4-mapped IPv6 equivalents.
- Resolve hostnames and reject if any resolved address is blocked.
- Pin DNS for the actual fetch, as `fetchWithPinnedDns()` does in `src/web/image-cache.ts`, to avoid DNS rebinding.
- Consider applying the same validation to all channel media downloads unless the source URL is known to be a trusted platform CDN.

## Areas Reviewed This Run

- `src/channels/discord.ts`: attachment and embed image download flow.
- `src/media.ts`: shared downloader, byte caps, timeout behavior, path-safe media storage.
- `src/web/image-cache.ts`: existing SSRF/DNS-rebinding mitigation that could be reused.
- `src/discovery/routes.ts`: context sync path validation and peer proxy paths.
- `src/web/server.ts`: auth/peer route gate review for discovery routes.
- `src/ipc.ts`: IPC sender authorization and response path handling.
- `src/db.ts`: sampled SQL construction and parameterization.

## Remaining Review Areas

- Other channel adapters that call `downloadBinaryAttachment()` or `downloadTextAttachment()`.
- Local backend/container runtime env propagation and mount construction.
- SolidStart web UI mode routes and generated client-side HTML sinks.

Peyton-Spencer:
## Summary

- Add opt-in remote URL validation and DNS pinning to media attachment downloads.
- Use the validated path for Discord embed images before server-side fetches.
- Add regression tests for private-address rejection and pinned DNS fetch behavior.

Fixes #667.

## Security Impact

Discord embed image URLs are user-controlled. This blocks embed image fetches that resolve to loopback, private, link-local, multicast, CGNAT, or other disallowed addresses, and pins the resolved public address for the actual fetch to reduce DNS rebinding risk.

## Tests

- `bun test src/media.test.ts src/channels/discord.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security validation for remote image URLs in Discord embeds.
  * Improved binary attachment downloads with stricter validation controls for remote URLs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/omniaura/omniclaw/pull/731?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/media.ts
+++ b/src/media.ts
@@ -121,9 +121,13 @@ export async function readStreamWithByteLimit(
  */
 export async function fetchWithTimeout(
   url: string,
-  options?: { timeoutMs?: number },
+  options?: {
+    timeoutMs?: number;
+    redirect?: NonNullable<RequestInit['redirect']>;
+  },
 ): Promise<Response> {
   return fetch(url, {
+    redirect: options?.redirect,
     signal: AbortSignal.timeout(
       options?.timeoutMs ?? DEFAULT_DOWNLOAD_TIMEOUT_MS,
     ),
@@ -133,6 +137,7 @@ export async function fetchWithTimeout(
 export interface AttachmentDownloadOptions {
   maxBytes?: number;
   timeoutMs?: number;
+  redirect?: NonNullable<RequestInit['redirect']>;
   /** Validate and DNS-pin untrusted remote URLs before fetching. */
   validateRemoteUrl?: boolean;
   /** Override DNS resolution (primarily for tests). */
@@ -155,7 +160,9 @@ async function fetchAttachment(
   }
 
   const pinnedAddress = validation.resolvedAddresses[0];
-  if (!pinnedAddress) return fetchWithTimeout(url, options);
+  if (!pinnedAddress) {
+    return fetchWithTimeout(url, { ...options, redirect: 'manual' });
+  }
 
   return (options.fetchWithPinnedDnsImpl ?? fetchWithPinnedDns)(
     url,
--- a/src/media.ts
+++ b/src/media.ts
@@ -12,6 +12,11 @@ import { GROUPS_DIR } from './config.js';
 import { logger } from './logger.js';
 import { assertPathWithin } from './path-security.js';
 import type { MediaAttachment, MediaAttachmentType } from './types.js';
+import {
+  fetchWithPinnedDns,
+  resolveAndValidateRemoteImageUrl,
+  type PinnedRemoteImageFetch,
+} from './web/image-cache.js';
 
 // ---------------------------------------------------------------------------
 // Constants
@@ -125,15 +130,49 @@ export async function fetchWithTimeout(
   });
 }
 
+export interface AttachmentDownloadOptions {
+  maxBytes?: number;
+  timeoutMs?: number;
+  /** Validate and DNS-pin untrusted remote URLs before fetching. */
+  validateRemoteUrl?: boolean;
+  /** Override DNS resolution (primarily for tests). */
+  lookupHostAddresses?: (hostname: string) => Promise<string[]>;
+  /** Override pinned DNS fetch (primarily for tests). */
+  fetchWithPinnedDnsImpl?: PinnedRemoteImageFetch;
+}
+
+async function fetchAttachment(
+  url: string,
+  options?: AttachmentDownloadOptions,
+): Promise<Response> {
+  if (!options?.validateRemoteUrl) return fetchWithTimeout(url, options);
+
+  const validation = await resolveAndValidateRemoteImageUrl(url, {
+    lookupHostAddresses: options.lookupHostAddresses,
+  });
+  if (validation.error) {
+    throw new Error(`Remote attachment URL rejected: ${validation.error}`);
+  }
+
+  const pinnedAddress = validation.resolvedAddresses[0];
+  if (!pinnedAddress) return fetchWithTimeout(url, options);
+
+  return (options.fetchWithPinnedDnsImpl ?? fetchWithPinnedDns)(
+    url,
+    pinnedAddress,
+    options.timeoutMs ?? DEFAULT_DOWNLOAD_TIMEOUT_MS,
+  );
+}
+
 /**
  * Download a binary attachment (image, video, audio) and return its bytes.
  * Enforces `maxBytes` (default: 10 MiB).
  */
 export async function downloadBinaryAttachment(
   url: string,
-  options?: { maxBytes?: number; timeoutMs?: number },
+  options?: AttachmentDownloadOptions,
 ): Promise<Buffer> {
-  const response = await fetchWithTimeout(url, options);
+  const response = await fetchAttachment(url, options);
   if (!response.ok) {
     throw new Error(`Download failed with status ${response.status}`);
   }
--- a/src/channels/discord.ts
+++ b/src/channels/discord.ts
@@ -1075,7 +1075,9 @@ export class DiscordChannel implements Channel {
             safeName,
             'embed',
           );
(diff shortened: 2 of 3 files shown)


src/media.ts
```diff
@@ -125,15 +130,49 @@ export async function fetchWithTimeout(
...
+
+  const validation = await resolveAndValidateRemoteImageUrl(url, {
+    lookupHostAddresses: options.lookupHostAddresses,
+  });
+  if (validation.error) {
+    throw new Error(`Remote attachment URL rejected: ${validation.error}`);
+  }
+
+  const pinnedAddress = validation.resolvedAddresses[0];
+  if (!pinnedAddress) return fetchWithTimeout(url, options);
```
Peyton-Spencer: This fallback still uses normal `fetch()` for validated direct-IP URLs, and `fetch()` follows redirects by default. A public direct-IP URL can pass `resolveAndValidateRemoteImageUrl()`, return a 302 to `http://127.0.0.1/...` or a metadata endpoint, and the redirected target is then fetched without any validation or DNS pinning. For `validateRemoteUrl` downloads, please either disable redirects (`redirect: 'manual'`) or manually follow each `Location` only after re-running the same validation/pinning logic. A regression test with a public direct-IP URL redirecting to loopback would cover this bypass.

Peyton-Spencer:
Holding off on merge for now because this is security-sensitive and CodeRabbit flagged a still-unresolved DNS pinning fallback: when remote URL validation is enabled, an empty resolved address set should fail closed rather than falling back to the unpinned fetch path. Please address or explicitly refute that path before merge.

Peyton-Spencer:
Bumping per PM assessment #739 — this is the sixth-pass carryover and is one merge away from closing #667 (longest-running open security carryover). CI fully green (format, typecheck, agent-runner-typecheck, test, validate-skill-only, CodeRabbit) and `mergeable: CLEAN` since 2026-05-19. The one CodeRabbit nit on DNS-pinning fallback in `resolveAndValidateRemoteImageUrl` is non-blocking — can land in a follow-up or be addressed in-place. Requesting a reviewer pass.

Peyton-Spencer:
I am holding off on merge because the unresolved DNS pinning comment looks security-relevant: if validation is enabled but no pinned resolved address is available, the code should fail closed instead of falling back to an unpinned fetch path. Please address that case or explain why `resolvedAddresses` cannot be empty after successful validation.

Peyton-Spencer (commented):
Technical review status: not ready to merge yet.

CI is green and the integration point is scoped correctly (`DiscordChannel` now routes embed media downloads through `downloadBinaryAttachment(..., { validateRemoteUrl: true })`). I also verified the PR is mergeable against `main`.

Blocking finding: `src/media.ts` still falls back to `fetchWithTimeout(url, options)` when `resolveAndValidateRemoteImageUrl()` returns no pinned address. That is not only a theoretical empty-DNS edge case: `resolveAndValidateRemoteImageUrl()` intentionally returns `{ error: null, resolvedAddresses: [] }` for direct-IP URLs after private-address validation. The fallback then uses normal `fetch()`, which follows redirects by default. A public direct-IP URL can pass validation and return a redirect to `http://127.0.0.1/...`, link-local metadata, or another private target, and the redirected URL will be fetched without another validation/pinning pass.

Please either:
- reject direct-IP/no-pinned-address validated downloads in this attachment path, or
- fetch them with redirects disabled (`redirect: 'manual'` or equivalent), or
- manually follow redirects only after re-running the same validation/pinning logic for each `Location`.

A focused regression test should cover a validated public direct-IP URL whose first response redirects to loopback/private. Until that path is closed, this does not fully fix the Discord embed SSRF issue.

Peyton-Spencer (commented):
Follow-up technical review on `d598444`: ready to merge from my side.

I re-checked the SSRF paths I previously blocked on:

- Direct-IP validated downloads now use `fetchWithTimeout(url, { ...options, redirect: 'manual' })` when validation succeeds with no pinned DNS address. The spread order means callers cannot override the manual redirect in this path. A 302 to `127.0.0.1`, `169.254.169.254`, or another private target is returned as a non-OK 302 response and rejected by `downloadBinaryAttachment()` instead of being followed.
- The new `does not follow redirects for validated direct-IP URLs` test covers the regression path I asked for and asserts exactly one `fetch()` call with `redirect: 'manual'`.
- Hostname URLs still validate first, reject private/link-local/loopback resolved addresses, and use `fetchWithPinnedDns()` with the pre-resolved address. That helper uses `http.request`/`https.request`, which does not automatically follow redirects, so a 3xx response in the pinned path is also surfaced rather than followed to a different host.
- Non-http(s), embedded credentials, localhost, private direct IPs, and private DNS results are still rejected by `resolveAndValidateRemoteImageUrl()` before any fetch.

GitHub status is green on `d598444` (format, typecheck, agent-runner-typecheck, test, validate-skill-only, CodeRabbit), and GitHub reports the PR as mergeable. I attempted local `bun test src/media.test.ts` in a fresh review checkout, but that checkout has no dependencies installed and failed before tests with missing `zod`; CI has the authoritative test run here.

Issue #732, opened 2026-05-19
docs: operator observability reference (/system, /ipc counters + reason codes + diagnose-by-symptom)


Peyton-Spencer:
## Problem

The operator observability surfaces on `/system` and `/ipc` have matured significantly over the last month — queue depth, retry counts, longest task age, lane reason, last error, pending and retry stat cards (`#693`, `#695`, `#703`, `#707`, `#719`) — but there is no operator-facing documentation that enumerates what counters and signals each page exposes, what they mean, or how to read them when diagnosing a slow / stuck system.

This was flagged as a "candidate to file next pass" in the 2026-05-18 product assessment (issue #718, "Gaps observed"). Now filing.

## Why this matters

- New operators (humans or agent-driven product managers) currently have to read `src/web/routes.ts`, `src/web/system-routes.ts`, and `src/web/ipc-routes.ts` to know what each card/number means.
- The `/system` and `/ipc` cards are valuable, but their value compounds with a short reference doc that tells you **which signal to look at first** when a specific class of symptom appears (e.g. "agent appears idle" → reason-code reference; "messages backing up" → queue depth + lane reason + longest task age).
- Without a doc, the surface evolves and silently drifts; with one, future card additions get a one-line entry and stay discoverable.

## Proposed fix

Create `docs/observability.md` (or extend README) with:

1. **Page index** — one line per Web UI route relevant to operator diagnostics (`/system`, `/ipc`, `/conversations`, `/network`, etc.) with one-sentence purpose.
2. **`/system` reference** — enumerate cards/counters (agent state breakdown, queue card with total + max retry, longest task age, peer health rollup) and what each is sourced from.
3. **`/ipc` reference** — enumerate pending and retry stat cards (#719), per-source/per-poll counters once #644 lands, watcher backpressure drop log (#717/#724), in-flight ages.
4. **Reason-code table** — `idle`, `waiting on lease`, `no work`, `offline`, `back-pressure`, `rate-limited`, etc., with one-line meaning and where each surfaces.
5. **"Diagnose by symptom" mini-runbook** — 4–6 entries: "messages backing up", "agent appears idle", "task retrying repeatedly", "IPC drops in log", "peer offline", "single group stuck".

Length target: <= 200 lines, mostly tables. No new code; just docs against existing endpoints.

## Acceptance criteria

- [ ] `docs/observability.md` exists.
[...]

PR #749, opened 2026-05-22
feat(web): surface recent task outcomes rollup on /system

Peyton-Spencer:
## Summary

Adds a *recent task outcomes* tile to the `/system` page that rolls up `task_run_logs` from the last hour across every scheduled task. Surfaces total runs, success/error split, and a by-outcome-state breakdown (`done` / `blocked` / `abandoned` / `unknown`).

Implements one of the remaining acceptance items on the operator observability rollup tracker (#644):

> `/system` enrichments — agent state breakdown, *recent task outcomes*, peer health rollup

Today an operator has to drill into individual task pages to see whether recent runs succeeded or got blocked. The new tile makes the recent run state visible at a glance alongside the existing agent / queue / container rollups.

## What changed

- New `getRecentTaskOutcomeStats(sinceIso)` in `src/db.ts` — single grouped `SELECT` over `task_run_logs` with the `status` + `outcome_state` buckets.
- New `RecentTaskOutcomes` type and optional `getRecentTaskOutcomes` hook on `WebStateProvider`. Keeping the hook optional avoids touching the ~14 existing `FakeState` / test stubs.
- `buildHealthData` computes `recent_task_outcomes` (with `window_ms`) using a one-hour window. Falls back to zeroes when the provider does not implement the hook.
- `renderSystemContent` renders a new metric card after the existing "tasks" card, reusing the existing `reasonRollup` helper for the outcome-state breakdown.
- Orchestrator wires the SQLite function into the real provider in `src/index.ts`.

## Test plan

- [x] `bun test` — 2133 pass / 0 fail
- [x] `bunx prettier --check` clean on touched files
- [x] New `getRecentTaskOutcomeStats` tests in `src/db.test.ts` cover the empty case, the success/error split, the outcome-state buckets (including `unknown` for missing state), and the time-window filter.
- [x] New `buildHealthData` / `renderSystemContent` tests cover the optional-hook fallback path, provider integration (verifies the `sinceIso` argument is a recent ISO timestamp), and the rendered tile shape.

## Related

- Tracking: #644
- Companion docs (no change needed): #732 / `docs/OBSERVABILITY.md`

### DIFF
--- a/src/db.migrations.test.ts
+++ b/src/db.migrations.test.ts
@@ -112,6 +112,13 @@ function getTableColumns(db: Database, table: string): string[] {
   return columns.map((c) => c.name);
 }
 
+function getIndexNames(db: Database, table: string): string[] {
+  const indexes = db.query(`PRAGMA index_list(${table})`).all() as Array<{
+    name: string;
+  }>;
+  return indexes.map((index) => index.name);
+}
+
 describe('db migrations (bun:sqlite)', () => {
   it('migrates legacy observed agents schema and keeps rows readable/writable', () => {
     // Use in-memory DB to avoid file system and module caching issues.
@@ -851,6 +858,40 @@ describe('versioned migration framework', () => {
     for (const name of expected) {
       expect(tableNames).toContain(name);
     }
+
+    db.close();
+  });
+
+  it('indexes task run logs by run timestamp on fresh DB', () => {
+    const db = new Database(':memory:');
+    runMigrations(db, allMigrations);
+
+    expect(getIndexNames(db, 'task_run_logs')).toContain(
+      'idx_task_run_logs_run_at',
+    );
+
+    db.close();
+  });
+
+  it('adds task run timestamp index to existing migrated DBs', () => {
+    const db = new Database(':memory:');
+    runMigrations(
+      db,
+      allMigrations.filter((migration) => migration.version <= 6),
+    );
+    db.exec('DROP INDEX IF EXISTS idx_task_run_logs_run_at');
+
+    expect(getSchemaVersion(db)).toBe(6);
+    expect(getIndexNames(db, 'task_run_logs')).not.toContain(
+      'idx_task_run_logs_run_at',
+    );
+
+    runMigrations(db, allMigrations);
+
+    expect(getSchemaVersion(db)).toBe(BASELINE_VERSION);
+    expect(getIndexNames(db, 'task_run_logs')).toContain(
+      'idx_task_run_logs_run_at',
+    );
 
     db.close();
   });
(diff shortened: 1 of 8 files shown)


Peyton-Spencer (commented):
Dex technical review for mergemaster:

I found one issue I would fix before merging.

Finding:

- `src/db.ts` adds `getRecentTaskOutcomeStats()` with `WHERE run_at >= ?`, but the schema only has `idx_task_run_logs(task_id, run_at)`. This new query is global across all tasks, so SQLite cannot efficiently seek by `run_at` using the existing task-prefixed index. `/system` polls `/api/health` every 5 seconds while open, and this PR wires `/api/health` to call the new aggregate each time, so an unbounded `task_run_logs` table turns into repeated full-table scans. Please add a `run_at`-leading index for existing and fresh databases, e.g. a new migration plus baseline coverage for `CREATE INDEX IF NOT EXISTS idx_task_run_logs_run_at ON task_run_logs(run_at)`.

Verification I ran:

- `bun test src/db.test.ts src/web/system.test.ts` passes, 83 tests.
- `bun run typecheck` passes.
- GitHub CI is green on format, typecheck, agent-runner-typecheck, test, skill check, and CodeRabbit status.

I did not find correctness issues in the aggregation or rendering shape, but I would not mark this ready until the query has the matching index.

Peyton-Spencer (commented):
Re-verified `19c835b` for the run-time index fix. The perf blocker I flagged is addressed.

What I checked:
- Fresh DB path: baseline schema now creates `idx_task_run_logs_run_at ON task_run_logs(run_at)`.
- Existing DB path: migration v7 creates the same index with `CREATE INDEX IF NOT EXISTS`.
- Coverage: migration tests assert both fresh DB creation and v6 -> v7 upgrade creation.
- Query plan: `EXPLAIN QUERY PLAN` for `getRecentTaskOutcomeStats()` reports `SEARCH task_run_logs USING INDEX idx_task_run_logs_run_at (run_at>?)`, so the global `WHERE run_at >= ?` query no longer requires a full table scan.

Validation:
- `bun test src/db.migrations.test.ts src/db.test.ts`
- GitHub checks are green.

Verdict: approved from my side.

Peyton-Spencer (commented):
- CI: all green; CodeRabbit review completed, no actionable comments
- Diff verified: adds getRecentTaskOutcomeStats() in db.ts + additive index migration (v7, idx_task_run_logs_run_at, IF NOT EXISTS, with upgrade-path test from v6); surfaces a rolled-up 'recent task outcomes' tile on /system. Provider method getRecentTaskOutcomes is optional so backward-compatible.
- Tests: db aggregation (window boundary, status/outcome split), migration create + upgrade, system render
Merging via squash. (Formal APPROVED N/A — shared GitHub identity.)

PR #753, opened 2026-05-22
fix(security): reject Docker mount target separators

Peyton-Spencer:
## Summary

Ports the mount target hardening from nanocoai/nanoclaw PR #1475 into OmniClaw's extra mount validator.

OmniClaw already blocks traversal and absolute extra mount targets, but configured `containerPath` values are later rendered into Docker `--mount` / `-v` arguments. This PR rejects Docker-sensitive separators and control characters before those values can reach the runtime argument builder.

Source security PR: https://github.com/nanocoai/nanoclaw/pull/1475

## What changed

- Reject `:` in extra mount `containerPath` values to avoid Docker `-v` option parsing ambiguity.
- Reject `,` in extra mount `containerPath` values to avoid Docker `--mount` option parsing ambiguity.
- Reject ASCII control characters so generated runtime args and logs stay single-line and unambiguous.
- Add regression coverage for each rejected form.

## Considered and rejected

- nanoclaw PR #171, Bash secret sanitization: already present in OmniClaw's agent runner.
- nanoclaw PR #2001, container-controlled outbox path confinement: OmniClaw does not have the same outbound outbox read/delete sink.
- nanoclaw PR #2566, scoped channel approval targets: permission module structure is not present in this codebase.
- nanoclaw PR #1475, stopContainer command injection: OmniClaw already validates orphan container names and uses `Bun.spawn` argv form.

## Validation

- `bun test src/mount-security.test.ts`
- `bun run typecheck`
- `bun run format:check src/mount-security.ts src/mount-security.test.ts`
- `git diff --check`

### DIFF
--- a/src/mount-security.test.ts
+++ b/src/mount-security.test.ts
@@ -232,6 +232,45 @@ describe('validateMount with allowlist', () => {
     expect(result.reason).toContain('Invalid container path');
   });
 
+  it('rejects container path with Docker volume option separator', () => {
+    writeAllowlist(baseAllowlist);
+    const result = validateMount(
+      {
+        hostPath: TEMP_ROOT + '/projects/my-app',
+        containerPath: 'my-app:rw',
+      },
+      true,
+    );
+    expect(result.allowed).toBe(false);
+    expect(result.reason).toContain('Invalid container path');
+  });
+
+  it('rejects container path with Docker mount option separator', () => {
+    writeAllowlist(baseAllowlist);
+    const result = validateMount(
+      {
+        hostPath: TEMP_ROOT + '/projects/my-app',
+        containerPath: 'my-app,readonly=false',
+      },
+      true,
+    );
+    expect(result.allowed).toBe(false);
+    expect(result.reason).toContain('Invalid container path');
+  });
+
+  it('rejects container path with control characters', () => {
+    writeAllowlist(baseAllowlist);
+    const result = validateMount(
+      {
+        hostPath: TEMP_ROOT + '/projects/my-app',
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Technical review: ready to merge.

I checked the changed validator against the Docker argument construction path:

- Configured extra mounts are centralized through `validateAdditionalMounts()` before `buildVolumeMounts()` appends them, and the final Docker args are built from that validated result. I did not find another configured `containerPath` path that bypasses `isValidContainerPath()` before reaching `--mount` / `-v`.
- Rejecting `:` and `,` is the right narrow fix for target-side Docker option parsing here: `:` matters for `-v source:target`, and `,` matters for `--mount type=bind,source=...,target=...,readonly`. Since args are passed as an argv array, spaces, quotes, `&`, and non-ASCII are not shell metacharacters in this code path. `=` inside the target value does not split a new `--mount` option without a comma separator, so I would not broaden the blocklist in this PR.
- The control-character rejection is appropriate for keeping generated args and logs single-line/unambiguous.
- The three regression tests cover the relevant rejection classes: `:`, `,`, and control chars. That is sufficient for this scoped change.

One non-blocking follow-up worth tracking separately: host/source paths are also interpolated into Docker mount argument strings after allowlist resolution. If an attacker can influence configured `hostPath` values or create oddly named directories under an allowed root, source-side separators may deserve the same treatment. I do not consider that a blocker for this focused target-path port because this PR materially closes the requested `containerPath` injection surface and does not regress behavior.

GitHub CI is green on `9ae9421` and the PR is mergeable.

PR #754, opened 2026-05-23
feat(web): surface LAN peer health rollup on /system

Peyton-Spencer:
## Summary

Adds a *peers* tile to the `/system` page summarizing LAN discovery state — discovery on/off, total peers, online, trusted, trusted offline, pending pair requests, and a by-status breakdown that mirrors the `/network` page taxonomy.

Implements one of the remaining acceptance items on the operator observability rollup tracker (#644):

> `/system` enrichments — agent state breakdown, recent task outcomes, *peer health rollup*

Today an operator has to leave `/system` and drill into `/network` to see whether trusted peers are reachable or whether someone is sitting on an unapproved pair request. The new tile makes that visible alongside the existing rollups so the same page tells the full health story.

## What changed

- New optional `getPeerHealth()` hook on `WebStateProvider` returning a `PeerHealthSnapshot`. Keeping the hook optional avoids touching the existing `FakeState` / test stubs; `buildHealthData` falls back to an empty-but-shaped snapshot when unimplemented.
- New `peers` field on `HealthData` mirrors the snapshot, so the `/api/health` JSON now includes the same numbers the tile renders.
- Renders a new `peers` metric card after the existing `queue` card, reusing the existing `reasonRollup` helper for the by-status breakdown.
- Orchestrator wires the snapshot in `src/index.ts` from the existing `trustStore`, `discoveryHandle`, and `discoveryRuntime` (no new globals). `revoked` peers are counted in the breakdown but excluded from totals/online/trusted to match how `/network` hides them.

## Test plan

- [x] `bun test src/web/system.test.ts` — 34 pass / 0 fail (5 new tests)
- [x] `bun test` — 2142 pass / 0 fail
- [x] `bunx prettier --check` clean on touched files
- New `buildHealthData` tests cover the optional-hook fallback path and the snapshot-passthrough path.
- New `renderSystemContent` tests cover the unavailable / active / disabled discovery states, the metric rows, and stable IDs for every `PeerStatus` key.

## Related

- Tracking: #644
- Companion: #732 / `docs/OBSERVABILITY.md` (no change needed — the new fields slot in under the same taxonomy)

## Summary by CodeRabbit

* **New Features**
  * Added peer health metrics card to system health UI, displaying LAN/discovery peer information including totals, online/trusted counts, and status breakdown.

* **Tests**
  * Added comprehensive test coverage for peer health snapshot behavior and system health rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/omniaura/omniclaw/pull/754?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/web/system.ts
```diff
@@ -546,11 +574,50 @@ export function renderSystemContent(
           'sys-queue-task-reason',
         ),
     ) +
+    // Peers rollup — mirrors the /network status taxonomy so operators
+    // can see LAN trust health at a glance without leaving /system.
+    metricCard(
```
Peyton-Spencer, line 649: This renders the new peer IDs, but the `/system` live poller in `page-scripts.ts` does not update any of them from `h.peers`. On a long-lived `/system` page, peer online/trusted/pending counts will stay at the initial server-rendered values even though `/api/health` now returns fresh peer data every 5s. Please wire the new `sys-peers-*` IDs and `sys-peers-status-*` breakdown into `systemScript()` (and add a focused test/assertion for the script text if that is the local pattern), or use the existing SSE patch path to refresh the tile.

Peyton-Spencer (commented):
Technical review: one fix requested before I would call this merge-ready.

I left an inline comment on the main behavioral gap: the new peer tile renders and `/api/health` now includes `peers`, but `systemScript()` does not update any `sys-peers-*` IDs during the existing 5s `/api/health` poll. That means a long-lived `/system` page will show stale LAN peer counts until reload. For an observability tile, this should refresh with the rest of the health surface.

Other review notes:

- `buildPeerHealthSnapshot()` mostly matches the current `/network` model because both use `trustStore.getAllPeers()` plus `discoveryHandle.getPeers()`. Revoked DB rows are already filtered by `getAllPeers()`, so they are excluded from totals and from the real `by_status` counts. A currently discovered revoked peer would appear as `discovered` on both surfaces because `/network` has the same stored-peer filter. That is existing behavior, not introduced here.
- Related nuance: because `getAllPeers()` filters revoked, the real snapshot path cannot currently produce `by_status.revoked > 0`; only tests/stubs can. I do not think that blocks the feature, but the PR body/code comment saying revoked peers are counted in the breakdown is misleading unless you switch to a store method that includes revoked rows.
- The optional `getPeerHealth` hook plus shaped zero fallback is reasonable for this codebase. It avoids forcing every test stub to learn discovery state, and `/system` already handles unavailable subsystems with empty-shaped data.
- Field names are OK. `trusted_offline`, `pending_requests`, and `discovery_active` are explicit enough and line up with the existing snake_case `/api/health` style.

CI is green for the main checks and the PR is mergeable; CodeRabbit was still pending when I checked. I did not run local tests in this fresh clone because dependencies are not installed here.

Peyton-Spencer:
Not merging yet. The new peer rollup is included in the initial `/system` HTML and `/api/health`, but the existing health polling script does not update the `sys-peers-*` DOM IDs. Please wire those fields into the 5s refresh path so long-lived `/system` pages do not show stale LAN peer health until reload.

src/web/system.ts
Peyton-Spencer: Wired up in efe16cf. `systemScript()` now updates every `sys-peers-*` cell (and the `sys-peers-status-{trusted,pending,discovered,revoked}` breakdown) on each /api/health tick, mirroring the `renderDiscoveryState()` ternary inline. Added 3 page-scripts.test.ts assertions covering the top-level peer fields, the by_status loop, and the discovery-state ternary. 2145/2145 tests pass, typecheck + prettier clean.

Peyton-Spencer (commented):
Mergemaster technical sign-off ✅ (post-rebase)
Independently verified on branch head 96dbf096:
- Fresh pull_request CI run 26382722526 green on head 96dbf096 (test, typecheck, format, agent-runner-typecheck, validate-skill-only); CodeRabbit completed
- Rebase preserved BOTH features in the overlapping files: src/web/system.ts retains #749's recent_task_outcomes rollup AND #754's LAN peer-health rollup; types.ts likewise
- MERGEABLE/CLEAN against current main (incl. #749 + #760, which also touched system.ts/types.ts) — no residual conflict
Merging via squash. (Formal APPROVED N/A — shared GitHub identity.)

PR #767, opened 2026-05-25
fix(router): reactions must not wake idle trigger-gated agents

Peyton-Spencer:
## Problem

In trigger-required channels, **emoji reactions were waking idle agents and forcing a reply.**

Root cause: `handleReactionNotification` (`src/index.ts`) synthesizes a notification message whose content is

```
@Agent [User reacted with :eyes:]
```

That string starts with `@Agent`, so it matches the per-agent trigger regex (`buildTriggerPattern`). When such a `react-*` message lands in a group's pending batch, the trigger gate in `processGroupMessages` treats it as a real mention and wakes the agent — even though nobody addressed it. With a bot like CodeRabbit repeatedly reacting `:eyes:`, this produced a stream of empty "holding"/acknowledgement replies in Slack.

## Fix

Extract the gate's trigger check into a small pure helper, `hasWakingTrigger(messages, triggerPattern)`, that **ignores synthetic `react-*` messages**. A reaction is not a mention, so it must never *independently* wake a trigger-gated agent.

```ts
export function hasWakingTrigger(messages, triggerPattern): boolean {
  return messages.some(
    (m) => !m.id.startsWith('react-') && triggerPattern.test(m.content.trim()),
  );
}
```

Behavior preserved:
- Real `@mentions` (and reactions batched alongside a real mention) still wake the agent.
- Reactions still **pipe into an actively-running container** as context (that path is unchanged) — they just no longer spawn a fresh run on an idle agent.
- Main group and `requiresTrigger: false` channels are unaffected.

## Tests

Added a `hasWakingTrigger` unit suite (mention wakes; lone reaction does not; mention+reaction wakes; unaddressed chatter does not). Full `src/index.test.ts` passes (25/25), `tsc --noEmit` clean, Prettier clean.

## Follow-ups (not in this PR — see tracking issue)

This stops idle-wake spam, but two related items deserve a product decision and are filed separately:
1. **Warm-container path**: once an agent is warm, subsequent non-mention messages still pipe in and can elicit a reply. The framework already drops empty output (`formatOutbound` -> `if (formatted)`), so the real lever is letting agents *choose silence*; worth making that an explicit, documented default.
2. **Bot/app reactions as pure noise**: reactions from non-human reactors (CodeRabbit, etc.) arguably should not notify at all. Needs reactor-identity plumbing in the Slack/Discord reaction handlers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed issue where synthetic reactions would incorrectly wake idle agents and trigger unnecessary message processing by implementing proper trigger validation.

* **Tests**
  * Added comprehensive test suite for trigger detection, verifying the system correctly identifies real triggers while ignoring reaction-synthesized content, handles batched messages, and ignores unrelated chatter.

[Review Change Stack](https://app.coderabbit.ai/change-stack/omniaura/omniclaw/pull/767?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -22,8 +22,9 @@ import {
   _truncateIntermediateStatusBuffer,
   _setSessions,
   getAvailableGroups,
+  hasWakingTrigger,
 } from './index.js';
-import { DATA_DIR } from './config.js';
+import { buildTriggerPattern, DATA_DIR } from './config.js';
 import fs from 'fs';
 import path from 'path';
 import type {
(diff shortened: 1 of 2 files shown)


Peyton-Spencer:
Tracking issue for the follow-up design decisions (warm-container path, bot-reaction filtering): #768

Peyton-Spencer (commented):
- CI: all green (test, typecheck, format, agent-runner-typecheck, validate-skill-only)
- Diff verified: extracts hasWakingTrigger() in src/index.ts and filters synthetic react-* notification ids so a reaction no longer wakes an idle, trigger-gated agent. Localized change (~lines 1730-1790), real bug fix.
- Tests: 4 new cases cover explicit mention, reaction-only (no wake), batched mention+reaction, unaddressed chatter
- No actionable CodeRabbit comments
Merging via squash. (Formal APPROVED review N/A — shared GitHub identity with author.)

Issue #768, opened 2026-05-25
Agents over-respond to reactions & non-mention chatter (reaction noise)


Peyton-Spencer:
## Summary

Agents in trigger-required channels respond too eagerly to things that are **not** addressed to them — primarily emoji reactions (a bot reacting `:eyes:`) and non-mention thread chatter. This surfaced as a stream of empty "holding"/acknowledgement replies in the Ditto `#engineering` Slack channel.

PR #767 fixes the most clear-cut part (reactions waking *idle* agents). This issue tracks the remaining design decisions, which need a product call before implementation.

## Diagnosis

1. **Reaction notifications match the trigger pattern.** `handleReactionNotification` (`src/index.ts`) builds content `@Agent [User reacted with :emoji:]`, which starts with `@Agent` and so satisfies `buildTriggerPattern`. → **Fixed in #767** via `hasWakingTrigger()` ignoring `react-*` ids.

2. **Warm-container continuity.** Once an agent has run, its container stays warm (`IDLE_TIMEOUT`). Reactions/messages then pipe straight into stdin (`queue.sendMessage`), bypassing the trigger gate, and a non-actionable input still yields a reply. The framework *already* suppresses empty output (`formatOutbound` -> `if (formatted)` at the send site), so the lever is **behavioral**: agents should be able to choose silence and it should be a documented default.

3. **Bot/app reactions are pure noise.** A reaction from CodeRabbit/another app carries no intent for our agent. The Slack `reaction_added` / Discord reaction handlers don't currently distinguish human vs bot/app reactors, nor filter the agent's own reactions.

## Proposed work

- [ ] **P0** (done, #767) — reactions don't wake idle trigger-gated agents.
- [ ] **P1** — Document & default "stay silent when nothing is addressed to you": agents emit only `<internal>…</internal>` (stripped by `stripInternalTags`, so nothing is sent) instead of filler acknowledgements. Bake into the group template guidance.
- [ ] **P1** — Filter reactions from bots/apps and the agent's own reactions in the reaction handlers (needs reactor-identity plumbing).
- [ ] **P2** — Consider whether *human* reactions should ever wake an idle agent at all, or only ever serve as context for the next real turn.

## Related

Slack-agent capability research (CodeRabbit parity, P0–P2 roadmap) lives in the engineering workspace at `projects/slack-agents-research-2026-05-25.md` — several items overlap (reaction-loop guard, Assistant API status, Block Kit).

Peyton-Spencer:
[...]

PR #790, opened 2026-06-01
feat(template): silent-by-default guidance for unaddressed Discord turns

Peyton-Spencer:
## Summary

Adds the "silent-by-default" template guidance sub-item from #768 (the P1 item that the 2026-05-28 PM comment de-blocked as ready-to-implement, no product call needed).

- Seeded `discord-channel` CLAUDE.md gets a new **"Stay Silent When Nothing Is Addressed to You"** section that points at the existing suppression path: emit only `<internal>…</internal>` and the turn is dropped before send.
- Template version bumped `2 → 3` so the reconciler (`reconcileSeededFiles`) updates already-seeded channel CLAUDE.md files on next pass.
- New test pins both the literal `<internal>` marker and the `stripInternalTags` reference in `getDiscordChannelSeed()`, so future template edits can't silently drop the guidance.

## Why this is the right shape

The mechanism is already real:

- `src/router.ts:96` — `stripInternalTags` removes `<internal>…</internal>` content.
- `src/router.ts:110` — `formatOutbound` short-circuits on empty text after stripping (`if (!text) return ''`).
- `src/ipc.ts:413` — IPC path suppresses internal-only messages with `reason: 'internal-only'`.

So no behavioral fork: we're just teaching seeded agents the pattern at template-seed time. Pairs with #767 (P0 — reactions don't wake idle trigger-gated agents) by giving warm-container turns a documented "do nothing" exit path instead of producing the filler acknowledgements observed in Slack `#engineering`.

## Test plan

- [x] `bun test src/seed-templates.test.ts` → 32/32 pass (new assertions on `<internal>` + `stripInternalTags` strings).
- [x] `bun test` → 2199/2199 pass.
- [x] `bun run typecheck` → clean.
- [x] `bunx prettier --check src/seed-templates.ts src/seed-templates.test.ts` → clean.

Refs #768. Does **not** close the issue — the second P1 sub-item (reactor-identity plumbing in `handleReactionNotification`) and the P2 close-recommendation are still open and tracked there.

### DIFF
--- a/src/seed-templates.test.ts
+++ b/src/seed-templates.test.ts
@@ -126,6 +126,12 @@ describe('getDiscordChannelSeed', () => {
     expect(content).toContain('secondary channel');
   });
 
+  it('includes silent-by-default guidance for unaddressed turns', () => {
+    const content = getDiscordChannelSeed();
+    expect(content).toContain('<internal>');
+    expect(content).toContain('stripInternalTags');
+  });
+
   it('matches the discord-channel template version', () => {
     const content = getDiscordChannelSeed();
     const template = SEED_TEMPLATES.get('discord-channel')!;
--- a/src/seed-templates.ts
+++ b/src/seed-templates.ts
@@ -31,12 +31,22 @@ export interface SeedTemplate {
 
 const DISCORD_CHANNEL_TEMPLATE: SeedTemplate = {
   key: 'discord-channel',
-  version: 2,
+  version: 3,
   body: `## Channel: Discord (Secondary)
 This group communicates via Discord, a secondary channel.
 You can freely answer questions and have conversations here.
 For significant actions (file changes, scheduled tasks, sending messages to other groups),
 confirm intent with the current user in this chat before proceeding.
+
+## Stay Silent When Nothing Is Addressed to You
+Not every turn deserves a reply. Reactions, ambient chatter between other people,
+and messages that don't mention or invite you are not requests for your output.
+Default to silence in those cases — don't post filler like "got it" or "looking into it".
+
+To end a turn silently, emit only \`<internal>…</internal>\` content. Anything inside
+those tags is stripped before send (see \`stripInternalTags\` in \`src/router.ts\`), and
+a turn that produces only internal-tagged text is suppressed entirely. Use this for
+private notes-to-self when the right action is no action.
 
 ## Getting Context You Don't Have
 When you need project context, repo access, credentials, or information that hasn't been shared with you:


Peyton-Spencer (commented):
Reviewed the template/version bump and the test coverage. No blocking findings from my side.

What I checked:
- The new Discord channel guidance points at the existing `stripInternalTags` / internal-only suppression behavior rather than adding a second behavior path.
- The `discord-channel` seed version bump means already-versioned channel files will reconcile from v2 to v3.
- The added test pins the key behavior markers in `getDiscordChannelSeed()`.

Local verification:
- `bun test src/seed-templates.test.ts` passed.
- `bun run typecheck` passed.
- `bunx prettier --check src/seed-templates.ts src/seed-templates.test.ts` passed.

Approving from my side. GitHub may not let this token submit a formal approval because it resolves to the PR author, but the review is LGTM.

Peyton-Spencer:
Day 8 of #819 escalation. This PR is +17/-1 across 2 files, CI 5/5 green for 14 days, and pairs with #802 to close the #768 reaction-noise P1. No review feedback on the thread, no blocker identified. Flagged again in today's rollup (#852).

If there's a hidden blocker that's keeping this parked, surfacing it on the PR thread would unblock the daily-rollup question and let the queue move on.

Peyton-Spencer (commented):
Technical review for PR #790.

CI status verified on PR head `1d7bfa45c747ac732f265799e2f7c9f7c30a8962`:
- CodeRabbit: pass
- agent-runner-typecheck: pass
- format: pass
- test: pass
- typecheck: pass
- validate-skill-only: pass

Diff verification:
- Reviewed `src/seed-templates.ts` and `src/seed-templates.test.ts`; the PR is limited to the Discord channel seed guidance plus a version bump/test coverage.
- The new silent-by-default guidance only applies to unaddressed/ambient Discord turns and does not tell agents to ignore addressed messages or explicit requests.
- The `<internal>...</internal>` guidance matches the existing behavior in `stripInternalTags`, `formatOutbound`, final result handling, and IPC send_message processing: internal blocks are stripped and internal-only output is suppressed.
- The `discord-channel` template version bump from v2 to v3 is appropriate so stale seeded channel CLAUDE.md files can reconcile.
- Added test asserts the seed includes the key `<internal>` and `stripInternalTags` markers; existing suppression tests cover internal-only output behavior.

Local verification run:
- `bun test src/seed-templates.test.ts`: 32 pass
- `bun test src/formatting.test.ts src/ipc-messages.test.ts`: 88 pass

Risk notes:
- Behavioral risk is low: this is prompt/template guidance only, not runtime routing logic.
- The main rollout consideration is that reconciled channel files will receive stronger silence guidance, which is intended for the #768 reaction-noise path and should reduce filler responses in secondary Discord channels.
- No prompt-injection issue found; the guidance references an existing suppression mechanism and does not grant new authority or broaden tool use.

Technical sign-off — ready to merge.

PR #760, opened 2026-05-23
feat(agents): per-agent model override in SQLite

Peyton-Spencer:
## Summary

Move per-agent model selection from `.env` into SQLite as the first step toward making multi-agent config easier. Each agent now has an optional `model` column on the `agents` table; when set, the orchestrator writes it into the container's env file at boot mapped to the runtime's expected variable (CLAUDE_MODEL / OPENCODE_MODEL / CODEX_MODEL / CURSOR_AGENT_MODEL). Per-agent value wins over the host `.env`, so `.env` stays as a global default.

Motivation: as we add more OpenCode agents that each want their own model, editing one shared `.env` doesn't scale. SQLite lets us configure per-agent without process restarts and without leaking model choices across agents.

## What changed

- `agents.model TEXT` column via migration 7 (default NULL; nullable empty = fall back to `.env`).
- `Agent.model?: string` on the typed record; `setAgent` preserves existing model when callers omit it (so topology resync doesn't wipe operator edits). New `setAgentModel(agentId, model)` helper.
- `ContainerInput.model?: string` in `@omniclaw/protocol`. The orchestrator pipes `agents[id].model` through `runAgent` and the scheduler (`SchedulerDependencies.getAgentModel`).
- `buildVolumeMounts` injects the override into the per-container env file via the existing allowlist/extract flow. Helper `modelEnvVarForRuntime(runtime)` does the mapping.
- Web UI: agent detail page renders a small Model form (`POST /api/agents/{id}/model`). Empty/null clears the override; trailing whitespace is trimmed; length is capped at 200.
- `agents.yaml` topology gains an optional `model:` field per agent.

## Config surface

```yaml
# agents.yaml
agents:
  coder-opencode:
    backend: docker
    agentRuntime: opencode
    model: anthropic/claude-sonnet-4-5
    channels:
      - jid: dc:1475601379400745101
        trigger: '@Coder'
```

```bash
# Web API
curl -X POST /api/agents/coder-opencode/model \
  -H 'Content-Type: application/json' \
  -d '{"model":"anthropic/claude-sonnet-4-5"}'
# clear:
curl -X POST /api/agents/coder-opencode/model \
  -H 'Content-Type: application/json' \
  -d '{"model":null}'
```

## Test plan

- [x] `bun test` — 2162 pass, 0 fail
- [x] `bunx tsc --noEmit` — clean
- New tests cover:
  - DB persistence, preservation across resync, trim/clear semantics, unknown-agent path
[...]

PR #815, opened 2026-06-07
feat(web): surface per-agent model override on /agents row

Peyton-Spencer:
## Summary

Adds a small **model-override chip** next to the runtime badge on each `/agents` table row when the agent has a per-agent model set. Operators triaging a mixed-model fleet (Opus on the admin agent, Sonnet on most workers, a Haiku tester here and there) can now spot the routing at a glance without opening each detail panel.

Closes the visibility gap left after #760 — the override is read by the orchestrator and surfaced on the detail page, but the agents directory has been silent about it. This PR keeps the data path untouched and is a pure render-layer addition.

## What changed

- `src/web/agent-channels.ts`
  - `AgentChannelData` now carries an optional `model` field. Populated from the Agent record for local agents; left undefined for remote peers since their model selection is managed on their own host and the discovery payload doesn't carry it.

- `src/web/agents-page.ts`
  - New `shortenModelLabel()` strips the longest known provider prefix (`claude-`, `anthropic/`, etc.) so the chip stays short (`opus-4-7` instead of `claude-opus-4-7`). The full id is hung off the `title` attribute and a `data-agent-model` data attribute, so hover and DOM queries still see the authoritative value.
  - New `renderRuntimeCell()` composes the runtime badge with an optional model chip, keeping `renderAgentRow` readable.

- `src/web/shared.ts`
  - New `.badge-model` style (violet, monospace) + light-theme tweak matching the existing `.badge-admin` palette.

## Why pair with the runtime cell

Model is a *per-runtime* concept (the same agent runtime can be pointed at different model ids), so reading them together is natural and avoids growing the table by a column. Empty / unset model — the common case — renders just the runtime badge as today, so existing rows stay exactly as before.

## Tests

17 new tests in `src/web/agents-page.test.ts`:

- `shortenModelLabel`: bare `claude-` ids, `provider/model` forms, unknown ids, whitespace handling, and the `claude-` edge case where stripping would leave an empty chip (falls back to the full label).
- `renderRuntimeCell`: absent-when-blank, present-when-set, full id on hover, HTML escaping, ordering (runtime first).
- `renderAgentRow`: chip present when `AgentChannelData.model` is set, absent without.
- `renderAgentsContent`: chip surfaces for local rows with an override; never projected onto remote peer rows.

## Test plan

- [x] `bun test src/web/agents-page.test.ts` — **91 pass** (was 74).
- [x] `bun test` — **2215 pass, 0 fail** across the full suite.
- [x] `bunx prettier --check` on the four touched files — clean.
- [ ] Reviewer: confirm the violet chip reads well alongside the existing backend / runtime / admin / remote badges on a live `/agents` page.

## Summary by CodeRabbit

* **New Features**
  * Added per-agent model override support with automatic fallback to default configuration.
  * Model information badges now display in agent rows with readable labels and full model identifiers available on hover.


Peyton-Spencer:
No blocking findings.

Reviewed the active diff after refreshing `origin/main`; it is scoped to the four web files listed in the PR description. The runtime-cell pairing looks right for the current table density: the model override qualifies the runtime more than it deserves its own column. The `shortenModelLabel` behavior is reasonable because it strips only display noise, keeps custom ids intact, and preserves the full value in both `title` and `data-agent-model`. The violet chip is close to `badge-admin` in light theme, but because it appears in the runtime cell and uses mono text, I do not think it is ambiguous enough to block.

Verified locally:
- `bun test src/web/agents-page.test.ts`
- `bun run tsc --noEmit`

CI is green except CodeRabbit still pending at review time.

Peyton-Spencer (commented):
Technical review for PR #815 after rebase.

Head verified: `b743827827512889ede307bfb64129b04ce07478` (`feat/agents-model-badge`).

GitHub readiness:
- PR state: OPEN / not draft
- Mergeability: MERGEABLE / CLEAN
- Fresh check rollup on `b743827`: format SUCCESS, validate-skill-only SUCCESS, typecheck SUCCESS, agent-runner-typecheck SUCCESS, test SUCCESS, CodeRabbit SUCCESS.

Rebase/diff verification:
- Rebasing against latest `main` resolved the stale conflict in `src/web/agents-page.test.ts`; source files auto-merged cleanly.
- The final diff is scoped to 4 web files: `agent-channels.ts`, `agents-page.ts`, `agents-page.test.ts`, and `shared.ts`.
- `AgentChannelData.model` is populated only for local agents from the authoritative `Agent` record; remote peer rows still do not get a projected model chip.
- Pairing the model chip with the runtime badge is coherent because the override is runtime/model configuration, and it avoids adding another table column.
- `shortenModelLabel()` keeps full ids in `title`/`data-agent-model` while shortening common `claude-` and provider-prefixed labels. Unknown model ids fall back to the raw final path segment, which is safe for later providers.
- The violet `.badge-model` styling is close to admin purple but distinguishable enough by location, monospace text, and light-theme override.

Local verification in isolated checkout:
- `bun install --force`
- `bun test src/web/agents-page.test.ts src/web/agent-channels.test.ts src/web/shared.test.ts` -> 180 pass / 0 fail
- `bun run typecheck` -> clean
- `bunx prettier --check src/web/agent-channels.ts src/web/agents-page.ts src/web/agents-page.test.ts src/web/shared.ts` -> clean

Risk notes:
- Low runtime risk: render-layer only, no API shape change outside internal web data composition.
- Main edge to watch is future model id taxonomy; current fallback keeps labels visible rather than hiding unknown providers.
- Rollback would remove only the inline visibility of model overrides on `/agents`; model override storage/control paths are unaffected.

Technical sign-off — ready to merge.

PR #811, opened 2026-06-06
feat(web): surface relative last-message time on /conversations chat rows

Peyton-Spencer:
## Summary

Replaces the long locale date on each `/conversations` chat row with a short relative string (`"5m ago"`, `"2h ago"`, `"3d ago"`) so operators triaging a busy multi-platform inbox can spot stale vs active chats at a glance.

- The absolute locale-formatted timestamp survives as the row's hover `title`, so the precise time is still one mouseover away.
- Future timestamps (clock skew, bad data) are clamped to `"now"` instead of rendering an `"in 5m"` surprise.
- Ages past ~30 days fall back to a locale date so the column never grows into an unreadable `"425d ago"`.

Pure render-layer change — no `WebStateProvider`, schema, or SSE event changes.

## Changes

- `src/web/conversations.ts` — new `formatChatRelativeTime` helper + `chat-time` class on the timestamp cell, with the locale date hung off the `title` attribute.
- `src/web/conversations.test.ts` — exhaustive helper coverage (now / minutes / hours / days / locale fallback / future / empty / unparseable) + integration assertions that each chat row carries the relative time and `title` attribute.

## Test plan

- [x] `bun test src/web/conversations.test.ts` — 39 pass.
- [x] `bun test` — 2208 pass, 0 fail across the full suite.
- [x] `bun run typecheck` — clean.

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Chat messages now display relative timestamps (e.g., "2m ago", "1h ago") instead of full date/time for improved readability
  * Hover over any chat timestamp to view the absolute date and time

* **Tests**
  * Added test coverage for relative timestamp formatting and conversation time rendering

PR #818, opened 2026-06-08
feat(web): surface last-seen relative time on /network offline peer rows

Peyton-Spencer:
## Summary

Adds a small `5m ago` / `3h ago` / `2d ago` chip next to the offline dot on each `/network` peer row. Operators triaging a multi-host fleet can now tell a "just disconnected" peer from a stale one at a glance without opening the peer detail panel.

Mirrors the `/conversations` relative-time pattern from #811 (and the styling cue from the `chat-time` chip).

## Changes

- `src/web/network.ts`
  - New `formatPeerLastSeen(lastSeen, nowMs?)` helper — returns `"now"`, `"5m ago"`, `"2h ago"`, `"3d ago"`, with a locale-date fallback past ~30 days and an empty string for null / unparseable input. `nowMs` is injectable so the unit suite stays deterministic.
  - New `renderPeerOnlineCell(peer)` composes the existing colored dot with an optional muted relative-time chip. Online peers stay as a bare green dot — they are "now" by definition. Offline peers with `lastSeen === null` keep the bare grey dot, so peers we have never seen offline don't gain a misleading chip.
  - The absolute ISO timestamp is hung off the chip's `title` attribute so the precise value is one mouseover away. Future-dated input (clock skew, bad data) clamps to `"now"` instead of rendering an `"in 5m"` surprise.

- `src/web/network.test.ts`
  - 10 new tests covering: `formatPeerLastSeen` boundaries (now / minutes / hours / days / locale fallback / null / undefined / empty / unparseable / future), `renderPeerOnlineCell` (online bare dot, offline bare dot, offline with chip + title), and `renderPeerRows` integration (chip present for offline-with-lastSeen, absent for online, absent for null lastSeen).

## Why pair with the online dot

Online vs offline is a binary state, but "how long has this peer been gone" is the next question every operator asks. Folding the chip into the existing column avoids a new `last seen` column (which would dwarf the dot's signal-to-ink ratio) and keeps the table the same width as today. Trusted-offline peers in particular are the most operationally interesting case — sync, browse, and remote logs all silently fail against them — so making their staleness visible without a click is a small but real win.

Pure render-layer change — no `PeerView` / `DiscoveryRuntimeSnapshot` / `WebStateProvider` changes.

## Test plan

- [x] `bun test src/web/network.test.ts` — **25 pass** (was 15).
- [x] `bun test` — **2212 pass, 0 fail** across the full suite.
- [x] `bunx prettier --check src/web/network.ts src/web/network.test.ts` — clean.
- [ ] Reviewer: confirm the muted chip reads well next to the green / grey dot on a live `/network` page (especially with a mix of trusted-offline and never-seen peers).

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * The network page now displays a "last seen" indicator for offline peers, showing both relative time (e.g., "5m ago", "2h ago") and an absolute timestamp on hover.

* **Tests**
  * Added comprehensive test coverage for last-seen formatting and peer online/offline cell rendering.

### DIFF
--- a/src/web/network.ts
+++ b/src/web/network.ts
@@ -183,6 +183,51 @@ export function renderPendingRequests(requests: PairRequest[]): string {
     .join('\n');
 }
 
+/**
+ * Format a peer's lastSeen ISO timestamp as a short relative string
+ * (e.g. "5m ago", "2h ago", "3d ago"). Used inline with the online dot on
+ * /network rows so an operator can tell a "just disconnected" peer from a
+ * stale one at a glance. Sub-minute deltas render as "now", future-dated
+ * inputs (clock skew) clamp to "now", and gaps past ~30 days fall back to a
+ * locale date so the cell never grows into an unreadable "425d ago".
+ */
+export function formatPeerLastSeen(
+  lastSeen: string | null | undefined,
+  nowMs?: number,
+): string {
+  if (!lastSeen) return '';
+  const t = Date.parse(lastSeen);
+  if (Number.isNaN(t)) return '';
+  const now = nowMs ?? Date.now();
+  const diff = Math.max(0, now - t);
+  if (diff < 60_000) return 'now';
+  if (diff < 3_600_000) return `${Math.floor(diff / 60_000)}m ago`;
+  if (diff < 86_400_000) return `${Math.floor(diff / 3_600_000)}h ago`;
+  if (diff < 30 * 86_400_000) return `${Math.floor(diff / 86_400_000)}d ago`;
+  return new Date(t).toLocaleDateString();
+}
+
+/**
+ * Compose the "online" cell content: the existing colored dot, plus a small
+ * "last seen" chip for offline peers when we know when we last saw them.
+ * Online peers stay as a bare dot — they are "now" by definition and the
+ * action column already telegraphs the trust state. The chip carries the
+ * absolute ISO timestamp on `title` so the precise value is one hover away.
+ */
+export function renderPeerOnlineCell(peer: PeerView): string {
+  const dot = peer.online
+    ? '<span style="color:var(--green)">●</span>'
+    : '<span style="color:var(--text-muted)">○</span>';
+  if (peer.online) return dot;
+  const rel = formatPeerLastSeen(peer.lastSeen);
+  if (!rel) return dot;
+  const title = peer.lastSeen ? ` title="${escapeHtml(peer.lastSeen)}"` : '';
+  return (
+    dot +
+    ` <span class="peer-last-seen" style="color:var(--text-muted);font-size:0.75rem;margin-left:0.35rem"${title}>${escapeHtml(rel)}</span>`
+  );
+}
+
 export function renderPeerRows(peers: PeerView[]): string {
   return peers
     .map((peer) => {
@@ -194,7 +239,7 @@ export function renderPeerRows(peers: PeerView[]): string {
         `<td><strong>${escapeHtml(peer.name)}</strong></td>` +
         `<td><code>${escapeHtml(peer.host)}:${peer.port}</code></td>` +
         `<td>${statusBadge}</td>` +
-        `<td>${peer.online ? '<span style="color:var(--green)">●</span>' : '<span style="color:var(--text-muted)">○</span>'}</td>` +
+        `<td>${renderPeerOnlineCell(peer)}</td>` +
         `<td>${actions}</td>` +
         `</tr>`
       );
--- a/src/web/network.test.ts
+++ b/src/web/network.test.ts
@@ -1,8 +1,10 @@
 import { describe, expect, it } from 'bun:test';
 
 import {
+  formatPeerLastSeen,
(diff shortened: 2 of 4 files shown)


Peyton-Spencer (commented):
Review note from @Cody: one consistency issue below. Targeted tests pass locally after installing dependencies: `bun test src/web/network.test.ts src/web/page-scripts.test.ts`.

src/web/network.ts
```diff
@@ -194,7 +239,7 @@ export function renderPeerRows(peers: PeerView[]): string {
         `<td><strong>${escapeHtml(peer.name)}</strong></td>` +
```
Peyton-Spencer, line 239: This updates the server-rendered row, but the active client refresh path still rebuilds `#peers-tbody` with the old bare-dot markup in `src/web/page-scripts.ts` (`refreshPeers()` -> JS `renderPeerRows()`). After any network action or peer refresh, offline rows with `lastSeen` will lose the chip until a full page reload. Can we mirror `formatPeerLastSeen` / `renderPeerOnlineCell` in the page script, or route refresh rendering through the server patch path, and add a page-script test for the refreshed row?

Peyton-Spencer: Good catch — pushed fc36703 to fix this. The client renderer now mirrors `formatPeerLastSeen` / `renderPeerOnlineCell` from `network.ts` so `refreshPeers()` rebuilds rows with the chip intact (poll, toggle-discovery, trust/revoke, request-access, approve/reject all reuse `renderPeerRows`).

Three page-script tests cover it:
- helpers exist and `renderPeerRows` routes through `renderPeerOnlineCell` (asserts the inline bare-dot `<td>` is gone)
- `formatPeerLastSeen` matches server thresholds (sub-minute → "now", hour/day cliffs, 30d → locale date, `Math.max(0, now-t)` future-clamp)
- `renderPeerOnlineCell` omits the chip when online or `lastSeen` missing, and escapes the title attribute via `window.__esc(peer.lastSeen)`

Full suite: 2215 pass, 0 fail. Typecheck + prettier clean.

Peyton-Spencer: @Cody re-review: resolved. `fc36703` mirrors the last-seen formatter/cell renderer into the client `page-scripts.ts` refresh renderer, keeps the future-date clamp and 30d locale-date behavior aligned with the server helper, and escapes the title through `window.__esc(peer.lastSeen)`. The new page-script tests cover the refresh path that dropped the chip.

Verified locally: `bun test src/web/network.test.ts src/web/page-scripts.test.ts` -> 45 pass / 0 fail.

Peyton-Spencer (commented):
Technical sign-off for PR #818.

Verification:
- GitHub reports the PR open, non-draft, mergeable, with green checks for CI format, CI test, CI typecheck, CI agent-runner-typecheck, Skill PR Check validate-skill-only, and CodeRabbit.
- Checked out the PR branch in a clean workspace at `/workspace/group/omniclaw-pr818-dex-signoff`.
- Ran `bun install --force && bun test && bun run typecheck` on the PR branch: passed. Full suite reported 2215 pass / 0 fail, then typecheck exited 0.
- Ran focused coverage with `bun test src/web/network.test.ts src/web/page-scripts.test.ts`: passed, 45 pass / 0 fail.
- Because `origin/main` has advanced since this PR's CI run, I also created a local-only merge-check branch against current `origin/main` and merged `feat/network-peer-last-seen`: no conflicts. `bun test && bun run typecheck` passed there too, with 2340 pass / 0 fail and typecheck clean.

Diff verification:
- Scope is limited to `src/web/network.ts`, `src/web/network.test.ts`, `src/web/page-scripts.ts`, and `src/web/page-scripts.test.ts`.
- `PeerView.lastSeen` is already part of the discovery API shape, and `/api/discovery/peers` builds peer rows with `lastSeen` from the trust store for both discovered and stored-only peers.
- SSR rendering now routes the online/offline cell through `renderPeerOnlineCell()`: online peers keep the existing bare green dot; offline peers with no valid `lastSeen` keep the bare grey dot; offline peers with `lastSeen` get the grey dot plus a muted `peer-last-seen` chip.
- The chip title carries the escaped absolute timestamp, while the visible label comes from `formatPeerLastSeen()`.
- Edge cases are covered: null/undefined/empty/unparseable values omit the chip; sub-minute and future-dated timestamps render `now`; minute/hour/day ranges render compact relative labels; very stale timestamps past roughly 30 days fall back to a locale date instead of huge day counts.
- Client refresh parity is handled: `page-scripts.ts` mirrors the formatter/cell helper and `refreshPeers()` rebuilds `#peers-tbody` through `renderPeerRows(peers)`, so the chip survives `/api/discovery/peers` polling and trust/request/revoke refreshes.

Risk note:
- Non-blocking: the formatter is duplicated between SSR and the inline page script, so future threshold changes need to keep both copies aligned. The PR adds page-script tests specifically to guard that parity.

No blocking issues found.

PR #831, opened 2026-06-11
feat(web): SSR live exec status and surface lane reason on /agents/:id

Peyton-Spencer:
## Summary

The `/agents/:id` header badge hardcoded "offline" on initial render, then flashed to the real status when the `/api/ipc/queue` poll returned. It also never surfaced the structured *message-lane reason* that the `/agents` list already shows in its `status` column.

This PR brings the agent detail header into parity with the agents list and aligns it with the operator-observability goals tracked in #644.

### What changed
- **Server-derive `execStatus` and `execReason`** in `buildAgentDetailData()`, using the same `getAgentExecStatus` / `getAgentExecReason` helpers the `/agents` row uses. `disabled` overrides still take precedence.
- **Use `renderExecStatusBadge`** so the reason chip ("cooling-down", "back-pressure", "retrying", "no-work") is inlined for non-active states — identical to the `/agents` row.
- **Wrap the badge** in `#ad-exec-status`; the live `/api/ipc/queue` poll now rebuilds inner HTML (status + optional reason span) on each tick instead of only swapping text/class. The JS mirrors `deriveMessageLaneReasonFromDetail()` so the reason matches the row exactly without a second round-trip.

### Why it matters
Operators landing on an agent's detail page during a busy session see immediately *why* an agent is idle/queued — cooling-down (healthy) vs back-pressure (waiting for slot) vs retrying (backoff after failure) — without having to bounce back to `/agents`. Closes a small parity gap called out in #644's "lightweight diagnostics for why an agent is idle, blocked, retrying, or offline" item.

### Files
- `src/web/agent-detail.ts` (+43, -1) — derive + render
- `src/web/page-scripts.ts` (+38, -9) — rebuild wrapper on poll, derive reason client-side
- `src/web/agent-detail.test.ts` (+144) — 5 new cases (executing, cooling-down, back-pressure, disabled override, remote fallback)

## Test plan

- [x] `bun test src/web/agent-detail.test.ts` — 43/43 pass, 120 expect calls
- [x] `bun test src/web/` — 663/663 pass
- [x] `bun test` — 2217/2217 pass
- [x] `bun run typecheck` — clean
- [x] `bunx prettier --check` — clean
- [ ] Reviewer: hit `/agents/<id>` on a live instance for an idle agent and confirm the reason chip shows on initial render (no "offline" flash) and updates on the 5s poll


src/web/page-scripts.ts
```diff
@@ -2361,24 +2361,55 @@ function agentDetailScript(): string {
...
+    '  else statusEl.removeAttribute("data-exec-reason");',
+    '}',
+    '',
     'function update(){',
     '  fetch("/api/ipc/queue").then(function(r){return r.json();}).then(function(details){',
-    '    var s=deriveStatus(details);',
-    '    statusEl.textContent=labels[s]||s;',
-    '    statusEl.className="badge badge-sm "+css[s];',
+    '    var d=details.find(function(g){return g.folderKey===folderKey;});',
+    '    if(!d){render("offline",null);return;}',
```
Peyton-Spencer, line 2415: This immediately overwrites the new SSR `disabled` badge. `buildAgentDetailData()` now correctly returns `execStatus: 'disabled'`, but `agentDetailScript()` calls `update()` on load and `/api/ipc/queue` has no enabled flag, so a disabled agent with no queue detail becomes `offline`; with stale/active queue detail it can become `executing`/`idle`. That means the new disabled override test only covers first render, not the live page behavior. The detail script needs to preserve the initial disabled state (for example by reading `data-exec-status="disabled"` and skipping queue-derived renders until the enable button reloads the page), or the poll payload needs enough agent state to re-derive the same override client-side.

Peyton-Spencer (commented):
Found one behavioral issue before merge: the SSR disabled badge is still overwritten by the immediate detail-page queue poll because the poll has no enabled/disabled signal. I do not see a CSS-selector problem from moving `#ad-exec-status` to the wrapper; the status styling still lives on the inner `.badge` and the repo references the id only from this script/tests. On the client mirror question: it is tolerable here because `/api/ipc/queue` already carries `messageLane.reason`, but the disabled case is a good example of why a server-provided detail status/reason payload would be less fragile.

src/web/page-scripts.ts
Peyton-Spencer: Good catch — fixed in e3c9c7d. update() now bails when statusEl already shows data-exec-status="disabled". The enable button's location.reload() is the sole transition path out, which is correct because /api/ipc/queue has no enabled flag to re-derive from. Added a page-scripts.test.ts assertion for the guard so this regression can't slip back in. 2218/2218 pass.

Peyton-Spencer: Re-reviewed e3c9c7d. This fixes the regression I flagged: once SSR marks the detail header `data-exec-status="disabled"`, the queue poll no longer replaces it with queue-derived state, and the enable path already reloads the page before the status can legitimately change.

I would keep this bail-out rather than expanding `/api/ipc/queue` for this PR. The queue endpoint is still lane state, while `enabled` is agent config/operator state; adding it just to let this one detail-page poll reconstruct the disabled override would broaden that API without another consumer. A richer live detail endpoint would be reasonable later if more agent config needs to update without reload.

Verified locally: `bun test src/web/page-scripts.test.ts src/web/agent-detail.test.ts` passes (61 tests). No further findings from my side.

Peyton-Spencer (commented):
Technical sign-off for PR #831.

CI status:
- GitHub reports PR #831 open, mergeable, and non-draft.
- GitHub checks are green on the PR head: format, typecheck, agent-runner-typecheck, test, validate-skill-only, and CodeRabbit all successful.
- Local clean checkout of `feat/agent-detail-exec-reason`:
  - `bun install --force` passed.
  - `bun test` passed: 2218 pass, 0 fail.
  - `bun run typecheck` passed (`tsc --noEmit`).
- `git merge-tree --write-tree origin/main HEAD` completed cleanly, so the current main composition has no text conflicts.

Diff verification:
- Scope is limited to `src/web/agent-detail.ts`, `src/web/agent-detail.test.ts`, `src/web/page-scripts.ts`, and `src/web/page-scripts.test.ts`.
- `buildAgentDetailData()` now derives `execStatus` and `execReason` from `state.getQueueDetails()` via the shared `/agents` helpers (`getAgentExecStatus`, `getAgentExecReason`), with the explicit disabled override taking precedence and remote agents falling back to offline/no reason.
- `renderAgentDetailContent()` renders the header badge through the shared `renderExecStatusBadge()` helper and stores `data-exec-status` / `data-exec-reason` on `#ad-exec-status`, so the SSR header starts aligned with the `/agents` row instead of flashing offline before the first poll.
- The agent-detail poll loop still reads `/api/ipc/queue`; it now derives the same execution status, derives/falls back to the message-lane reason, renders the reason badge for non-active/non-disabled states, and preserves the SSR disabled badge because the queue API does not carry the agent enabled flag.
- Tests cover active executing state with suppressed reason, idle `cooling-down`, queued `back-pressure`, disabled override, remote fallback, and the disabled-poll guard.

Risk notes:
- Low-risk web UI change; it does not alter queue execution, persistence, or routing.
- The client poll path duplicates the server's status/reason derivation and known reason set. The PR keeps it synchronized today, but a future message-lane reason addition will need both server and client badge logic updated.
- Disabled agents can remain visually disabled until a reload if another actor toggles them, which is consistent with the current queue API lacking enabled-state data and preferable to falsely showing queue-derived activity.

No blockers found.

PR #859, opened 2026-06-17
fix(security): reject control chars in agent model override (#857)

### LINKED ISSUE #857: Security: Model override newline injection can add container env vars
## Summary

Per-agent model overrides are accepted as arbitrary strings up to 200 characters and are later written verbatim into the container env file. A model value containing a newline can inject additional environment variables into the next agent container run.

## Severity

P1 / High.

The Web UI setting appears to control only the model name, but it can alter arbitrary container environment variables. A compromised authenticated dashboard session, malicious browser-origin action against an authenticated operator, or any future caller of `setAgentModel` can pivot from model selection into provider/base-url/env manipulation for subsequent agent runs. For example, injecting an alternate provider base URL alongside an existing API token can redirect SDK traffic and expose credentials.

## Affected Code

- `src/web/routes.ts`: `handleSetAgentModel()` trims and length-limits `body.model`, but does not reject control characters or line breaks.
- `src/db.ts`: `setAgentModel()` stores the trimmed string as-is.
- `src/backends/local-backend.ts`: `buildVolumeMounts()` appends the override directly with `filteredLines.push(`${modelEnvVar}=${modelOverride}`)` before writing `data/env/<agent>/env`.

## Reproduction Sketch

1. Authenticate to the Web UI.
2. Send a model override containing a newline:

   ```http
   POST /api/agents/<agent-id>/model
   Content-Type: application/json

   {"model":"claude-opus-4-6\nANTHROPIC_BASE_URL=https://attacker.example"}
   ```

3. Trigger that agent.
4. The generated env file can contain two entries, e.g.:

   ```env
   CLAUDE_MODEL=claude-opus-4-6
   ANTHROPIC_BASE_URL=https://attacker.example
   ```

5. The container entrypoint/environment loader can then apply the injected variable to the agent process.

## Expected Behavior

Model override values should be constrained to a single safe env value. At minimum, reject `\r`, `\n`, NUL, and other control characters before storing or writing the override. Prefer a positive allowlist for model identifiers if provider syntax is known.

## Suggested Fix

- Add a shared validator for per-agent model overrides.
- Reject control characters in `handleSetAgentModel()` and/or `setAgentModel()`.
- Defensively escape or serialize env file values instead of concatenating raw `KEY=value` lines.
- Add regression tests that verify newline-containing model strings are rejected and cannot create extra env entries.

## Areas Reviewed This Run

- `src/web/routes.ts`: Web UI task, context, avatar, enabled/model mutation handlers.
- `src/web/server.ts` and `src/web/session-auth.ts`: session/basic auth and peer-auth route boundaries.
- `src/web/page-scripts.ts` and `src/web/shared.ts`: remaining `innerHTML` builders after the Markdown sanitizer fix.
- `src/web/image-cache.ts` and `src/media.ts`: remote image/attachment SSRF protections, content type checks, and byte caps.
- `src/discovery/routes.ts` and `src/discovery/peer-client.ts`: peer HMAC signing, replay protection, and context sync path checks.
- `src/task-preprocessor.ts`: workflow path validation, env allowlist, spawn usage, output/time caps.
- `src/backends/local-backend.ts`: container spawn args, mounts, env-file generation, and cleanup command handling.

## Duplicate Check

- `gh issue list --repo omniaura/omniclaw --state all --search "model override env injection newline"` returned no matches.
- `gh issue list --repo omniaura/omniclaw --state all --search "CLAUDE_MODEL OPENCODE_MODEL newline"` returned no matches.
- `gh pr list --repo omniaura/omniclaw --state all --search "model override env injection"` returned only the original model override feature PR, not a fix.

Peyton-Spencer:
Closes #857.

## Summary

Per-agent model overrides flowed from the Web UI body straight into the container env file (`data/env/<agent>/env`), where `buildVolumeMounts()` in `src/backends/local-backend.ts` writes them as raw `KEY=value` lines. A value containing `\n` could append an additional env entry (e.g. `ANTHROPIC_BASE_URL=https://attacker.example`) and redirect SDK traffic on the next agent run.

Repro: `POST /api/agents/<id>/model` with `{"model":"foo\nANTHROPIC_BASE_URL=https://attacker.example"}` yielded a two-line env file on next container run.

## Fix

Adds a shared validator at `src/agent-model-validation.ts` and uses it at both call sites so neither path can ever land a control-character payload on disk:

- `src/web/routes.ts` — `handleSetAgentModel()` calls `validateModelOverride()` and returns `400 "model" must not contain control characters` for any ASCII control char in `0x00–0x1F` plus `0x7F`. Replaces the old inline length/trim block — same length cap (`MAX_MODEL_OVERRIDE_LENGTH = 200`), same null/blank-clears semantics.
- `src/db.ts` — `setAgentModel()` calls `assertSafeModelOverride()` as a defense-in-depth tripwire. A bypassed HTTP validator (future caller, IPC path, etc.) now throws `InvalidModelOverrideError` instead of silently persisting.

The validator is one file, ~95 LOC, fully covered by direct unit tests.

## Tests

All three test files are green locally; full suite passes (`bun test` → 2307 pass / 0 fail).

- `src/agent-model-validation.test.ts` *(new)*:
  - Null / undefined / blank → cleared.
  - Trim on accept.
  - Non-string rejected.
  - Length cap boundary (exact `MAX_MODEL_OVERRIDE_LENGTH` accepted, +1 rejected).
  - **Every code point in `0x00–0x1F` plus `0x7F` rejected** (33 control chars total).
  - The exact `#857` newline-injection payload rejected with the expected error string.
  - `assertSafeModelOverride` happy-path + throws `InvalidModelOverrideError`.
- `src/web/routes.test.ts`:
  - New `rejects model values containing control characters` covers newline, carriage return, NUL, tab, and DEL. Asserts the unsafe value never reaches `state.setAgentModel`.
  - New happy-path test confirms a well-formed `anthropic/claude-opus-4-6` still persists.
- `src/db.test.ts`:
  - New `setAgentModel throws on model values containing control characters` covers the persistence-layer tripwire. Asserts the stored model is unchanged.

## Notes

- No DB migration: validation is pure-code; existing rows are untouched.
- The `MAX_MODEL_OVERRIDE_LENGTH` constant moves from `routes.ts` to the validator so handler and DB agree on the cap.
- Per `CLAUDE.md`: targeted `--base main`, used `--body-file`, included `Co-Authored-By:` trailer.

## Test plan

- [x] `bun test src/agent-model-validation.test.ts` — 9 pass
- [x] `bun test src/db.test.ts` — 32 pass (incl. new control-char regression)
- [x] `bun test src/web/routes.test.ts` — 65 pass (incl. new control-char + happy-path regressions)
- [x] `bun test` — 2307 pass / 0 fail
- [x] `bun run build` — bundles cleanly
- [x] `bunx prettier --check` on changed files — all matched files use Prettier code style

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for agent model override inputs to reject strings containing control characters (newline, carriage return, null, tab, DEL) and enforce maximum length limits, preventing injection attacks and improving system security.

### DIFF
--- a/src/agent-model-validation.test.ts
+++ b/src/agent-model-validation.test.ts
@@ -1,0 +1,94 @@
+import { describe, expect, it } from 'bun:test';
+
+import {
+  InvalidModelOverrideError,
+  MAX_MODEL_OVERRIDE_LENGTH,
+  assertSafeModelOverride,
+  validateModelOverride,
+} from './agent-model-validation.js';
+
+describe('validateModelOverride', () => {
+  it('returns null for null / undefined / blank', () => {
+    expect(validateModelOverride(null)).toEqual({ ok: true, value: null });
+    expect(validateModelOverride(undefined)).toEqual({ ok: true, value: null });
+    expect(validateModelOverride('')).toEqual({ ok: true, value: null });
+    expect(validateModelOverride('   ')).toEqual({ ok: true, value: null });
+  });
+
+  it('trims surrounding whitespace on accept', () => {
+    expect(validateModelOverride('  claude-opus-4-6  ')).toEqual({
+      ok: true,
+      value: 'claude-opus-4-6',
+    });
+  });
+
+  it('rejects non-string non-null inputs', () => {
+    expect(validateModelOverride(42)).toEqual({
+      ok: false,
+      error: '"model" must be a string or null',
+    });
+    expect(validateModelOverride({ model: 'x' })).toEqual({
+      ok: false,
+      error: '"model" must be a string or null',
+    });
+  });
+
+  it('rejects values longer than MAX_MODEL_OVERRIDE_LENGTH after trim', () => {
+    const tooLong = 'x'.repeat(MAX_MODEL_OVERRIDE_LENGTH + 1);
+    const result = validateModelOverride(tooLong);
+    expect(result.ok).toBe(false);
+    if (!result.ok) {
+      expect(result.error).toContain(`${MAX_MODEL_OVERRIDE_LENGTH}`);
+    }
+  });
+
+  it('accepts a value exactly at MAX_MODEL_OVERRIDE_LENGTH', () => {
+    const exact = 'a'.repeat(MAX_MODEL_OVERRIDE_LENGTH);
+    expect(validateModelOverride(exact)).toEqual({ ok: true, value: exact });
+  });
+
+  // Core #857 regression: every byte in 0x00–0x1F plus DEL must be rejected.
+  it('rejects every ASCII control character (0x00–0x1F and 0x7F)', () => {
+    const codes: number[] = [];
+    for (let c = 0x00; c <= 0x1f; c++) codes.push(c);
+    codes.push(0x7f);
+    for (const code of codes) {
+      const ch = String.fromCharCode(code);
+      const result = validateModelOverride(`claude-opus-4-6${ch}foo`);
+      expect(result.ok).toBe(false);
+      if (!result.ok) {
+        expect(result.error).toContain('control characters');
+      }
+    }
+  });
+
+  it('rejects the exact #857 newline-injection payload', () => {
+    const result = validateModelOverride(
+      'claude-opus-4-6\nANTHROPIC_BASE_URL=https://attacker.example',
+    );
+    expect(result).toEqual({
+      ok: false,
+      error: '"model" must not contain control characters',
+    });
+  });
+});
+
+describe('assertSafeModelOverride', () => {
+  it('returns the normalized value for safe input', () => {
+    expect(assertSafeModelOverride(null)).toBeNull();
+    expect(assertSafeModelOverride('  claude-opus-4-6  ')).toBe(
+      'claude-opus-4-6',
+    );
+  });
+
+  it('throws InvalidModelOverrideError for unsafe input', () => {
+    let caught: unknown;
+    try {
+      assertSafeModelOverride('claude-opus-4-6\nFOO=bar');
+    } catch (err) {
+      caught = err;
+    }
+    expect(caught).toBeInstanceOf(InvalidModelOverrideError);
+    expect((caught as Error).message).toContain('control characters');
+  });
+});
--- a/src/agent-model-validation.ts
+++ b/src/agent-model-validation.ts
@@ -1,0 +1,95 @@
+/**
+ * Shared validation for per-agent model overrides.
+ *
+ * Model overrides flow from the Web UI into the agent container env file
+ * (`data/env/<agent>/env`) where they are concatenated as `KEY=value` lines.
+ * A value containing `\r`, `\n`, NUL, or other ASCII control characters
+ * could inject additional env entries (e.g. `ANTHROPIC_BASE_URL=...`) into
+ * the next container run. See issue #857.
+ *
+ * This module is the single source of truth for accepting / rejecting a
+ * model override string. Both the HTTP handler (`handleSetAgentModel` in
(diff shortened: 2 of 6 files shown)


Peyton-Spencer (commented):
Security review result: not signed off yet.

Verification performed:
- GitHub PR metadata: open, mergeable, non-draft; checks green including test/typecheck/format and CodeRabbit.
- Clean checkout at `34f94cc1e8438f36881f7e04887aedf6e216de54`.
- `bun install --force && bun test && bun run typecheck` passed: 2307 tests, 0 failures; typecheck clean.
- Focused `bun test src/agent-model-validation.test.ts src/web/routes.test.ts src/db.test.ts` passed: 131 tests, 0 failures.
- Manually probed `validateModelOverride()` for `\r`, `\n`, `\0`, `\t`, vertical tab, escape, DEL, plus checked direct model persistence paths.

Blockers:

1. `validateModelOverride()` trims before checking the control-character regex. That means leading/trailing whitespace control characters are accepted and normalized instead of rejected. Examples from the clean checkout:
   - `"\nANTHROPIC_BASE_URL=evil"` returns `{ ok: true, value: "ANTHROPIC_BASE_URL=evil" }`
   - `"ANTHROPIC_BASE_URL=evil\n"` returns `{ ok: true, value: "ANTHROPIC_BASE_URL=evil" }`
   - leading/trailing `\t`, `\r`, and vertical tab behave the same way.

   Middle-position control chars are rejected, and `\0`, ESC, DEL are rejected, but the current behavior does not match the security claim that control chars are rejected. The regression tests also only cover embedded control chars, so they miss this boundary case. If the intended policy is "strip boundary whitespace, reject embedded controls", the PR should state/test that explicitly; for the P1 injection fix I think reject-any-control is the safer policy.

2. There is still a bypass path through `setAgent()`. `src/db.ts` validates `setAgentModel()` with `assertSafeModelOverride()`, but `setAgent(agent)` still trims and persists `agent.model` directly. `src/agent-topology.ts` accepts `model: z.string().min(1).max(200).optional()` without the shared control-character validation, then carries that model through `registeredGroupToAgent()` / registration into `setAgent()`. So a declarative/topology/register path can still persist a model override containing `\n` even though the web `handleSetAgentModel()` path is now defended.

Suggested fix:
- Run the shared validation before trimming/sanitizing, or otherwise reject if the raw string contains `[\x00-\x1F\x7F]`.
- Use the same shared validator in `setAgent()` when `agent.model !== undefined`, or refine the topology schema with the same rule before it reaches DB persistence. Defense-in-depth is important here because `setAgentModel()` is not the only model write path.
- Add regressions for leading/trailing `\n`, `\r`, `\t`, vertical tab and for `setAgent({ model: "claude\nEVIL=1" })` / topology model rejection.

Peyton-Spencer:
Heads up: #865 was filed later today as a second fix for the same #857 issue (+126/-18 vs this PR's +284/-22). Both have 6/6 CI ✅. This PR has broader test coverage. Pick one and close the other to retire #857 — see #866 for context.

Peyton-Spencer:
PM note (2026-06-20, ref #883): this PR is a **duplicate** of #865, which merged on 2026-06-18 and already closed #857.

GitHub's `closedByPullRequestsReferences` on #857 lists both #859 and #865 — but only #865 is on `main`. The merged fix in `56162f6` already covers the newline-injection vector this PR addresses (validator + DB tripwire). #859 has been `CONFLICTING` against `main` since the same code area changed.

**Recommendation: close without merge.** Re-opening or rebasing would only re-introduce the same fix path on top of itself.

Peyton-Spencer:
Closing as duplicate of merged PR #865, which already closed #857 with the same model-override control-character validation fix. This branch is now conflicting, so keeping it open would only add queue noise.

PR #843, opened 2026-06-13
feat(web): surface task last-run time and outcome on /agents/:id tasks table

Peyton-Spencer:
## What

Add a *last run* column to the scheduled-tasks table on `/agents/:id`. Each row now shows a short relative time ("5m ago", "2h ago", "3d ago", "—" when never run) with the same `run-success` / `run-error` color classes already used on `/tasks`, and a tooltip carrying the absolute timestamp plus the outcome string.

Before vs after on a row:

```
before:  active  Run the daily check  cron: 0 9 * * *  3/2/2026, 9:00:00 AM  [pause]

after:   active  Run the daily check  cron: 0 9 * * *  3/2/2026, 9:00:00 AM  5m ago  [pause]
                                                                            (green if success, red if error)
```

## Why

The agent detail page's tasks table showed `status / prompt / schedule / next run` but nothing about *when* or *how* each task last ran. An operator investigating a slow or stalled agent had to bounce to `/tasks` to see the last-run timestamp and outcome already surfaced there. Now those signals live right next to the agent's other state — channels, conversations, exec status — closing the loop on the per-agent investigation flow.

This matches the rolling cadence of recent operator-observability work on `/agents/:id` and `/tasks` (#779 per-task last outcome, #793 currently-running tasks, #831 SSR exec status + lane reason) — small additive signals on existing pages, no new top-level routes.

## How

- *Data shape* — `AgentDetailData.tasks[]` already mapped `last_run`. This adds `last_result` so the JSON response from `/api/agents/:id/detail` also carries the outcome string (already available on `ScheduledTask` rows in the DB).
- *Render* — `renderAgentDetailContent` adds one column to the tasks table between `next run` and the action cell. Header gets `last run`. Empty-tasks colspan bumps from 5→6 (local) / 4→5 (remote).
- *Helpers* — two new exported pure functions keep the formatting deterministic and unit-testable:
  - `formatTaskLastRun(isoStr, now?)` — `<1m / Nm / Nh / Nd ago`, em-dash when null, passthrough on unparseable input. `now` is injectable for tests.
  - `lastRunOutcomeClass(result)` — `success → run-success`, `error → run-error`, otherwise empty (mirrors `/tasks`).

## Test plan

- [x] `bun test` — full suite (2238 tests) green
- [x] `bun run typecheck` clean
- [x] `bunx prettier --check` clean on touched files
- [x] 12 new tests added in `src/web/agent-detail.test.ts`:
[...]

PR #867, opened 2026-06-18
fix(web): color /tasks and /agents/:id rows from normalized outcome state (closes #858)

### LINKED ISSUE #858: tasks: normalize last_outcome so /tasks and /agents/:id badges color real success/error rows
## Context

Surfaced during PR #843 sign-off ([review comment](https://github.com/omniaura/omniclaw/pull/843#pullrequestreview)). The `/agents/:id` task table — and the existing `/tasks` table it mirrors — only render the colored `run-success` / `run-error` class when `scheduled_tasks.last_result` is the exact string `success` or `error`.

In practice, several scheduler paths write summary strings into `last_result` (e.g. `Completed`, `Error: ...`, or a truncated result excerpt). Rows for those tasks still show the last-run timestamp and tooltip, but they will not pick up the success/error coloring even when the run was clearly successful or failed.

## Why

Operator UX: the colored badge is the at-a-glance "did this run succeed" signal on both the `/tasks` and `/agents/:id` surfaces. Today it's silently undercounting successes/failures whenever the scheduler stored a summary string rather than a normalized token.

## Suggested direction

Either:

1. Normalize the outcome at the source — give `scheduled_tasks` a dedicated `last_outcome` column (or equivalent) that is only ever `success` / `error` / `pending`, and surface that in `buildAgentDetailData()` / the `/tasks` renderer. Keep `last_result` as the free-text summary for the tooltip.
2. Or, on the read side, derive the outcome by parsing `last_result` (e.g. prefix-matching `Error:`) and feed that into the existing renderer. Cheaper but more fragile.

Option 1 is cleaner — the read side already has the right shape, the data is just under-specified at the write side.

## Acceptance

- `last_outcome` (or equivalent) is populated on every task run, in both fresh runs and back-filled historical rows.
- `/tasks` and `/agents/:id` show the colored badge for all real success/error rows, not just ones where the scheduler happened to write the exact token.
- Tests cover the summary-string case that PR #843 leaves uncovered.

Filed as a follow-up to PR #843 (merged 2026-06-17). Not blocking that PR — the table behavior matches the pre-existing `/tasks` convention.

Peyton-Spencer:
## Summary

Closes #858. The colored `run-success` / `run-error` row classes on `/tasks` and the `/agents/:id` tasks table only fired when `scheduled_tasks.last_result` was the exact string `success` or `error`. In practice every scheduler write path stores a free-text summary into `last_result` (`'Completed'`, `result.slice(0, 200)`, `'Error: ...'` — see `updateTaskAfterRun` in `task-scheduler.ts`), so real-world rows rendered uncolored even when the run clearly succeeded or failed. The colored outcome badge (added in #843) was the only "did this run succeed" signal, and the row color was silently undercounting.

This change normalizes the row color to read from the already-populated `last_outcome_state` column (`done` / `blocked` / `abandoned` / `skipped`, via `inferOutcome` in `task-scheduler.ts`):

- `done` → `run-success`
- `blocked`, `abandoned` → `run-error`
- `skipped` → neutral — intentional no-op, the outcome badge already conveys it
- `null` last_outcome_state → fall back to the legacy `last_result === 'success' | 'error'` token check so pre-normalization rows still pick up coloring

This is the option-1 direction the issue recommends (cleaner; the read side already has the right shape, the data was just under-specified before — and `last_outcome_state` is now populated on every write path).

## Changes

- `src/web/agent-detail.ts` — `lastRunOutcomeClass(state, result?)` takes both inputs and prefers the normalized state; `AgentDetailData.tasks` now carries `last_outcome_state` through to JSON consumers.
- `src/web/tasks.ts` — call the shared helper instead of inlining the ternary.
- `src/web/page-scripts.ts` — same precedence on the SSE-driven client refresh so a live `/tasks` poll matches the SSR render.
- Test coverage in `agent-detail.test.ts` and `tasks.test.ts` for: scheduler summary strings (`'Completed'`, `'Error: container failed to start'`, `'Skipped by preprocessor: ...'`), the legacy `success` / `error` token fallback, and precedence between the two.

`last_outcome_state` has been populated on every scheduler write path since the column was added, so no backfill is required for new rows. Pre-existing rows that lack the normalized state but carry the legacy token continue to color via the fallback.

## Test plan

- [x] `bun test src/web/agent-detail.test.ts src/web/tasks.test.ts` → 96 pass, 0 fail
- [x] `bun test src/web/` → 789 pass, 0 fail
- [x] `bun test` (full suite) → 2367 pass, 0 fail

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced accuracy of task outcome status display by shifting from text-based parsing to normalized state tracking, ensuring task results display consistently and reliably across the application.

* **Tests**
  * Added comprehensive test coverage for task outcome styling across multiple result scenarios and state combinations.

### DIFF
--- a/src/web/page-scripts.ts
+++ b/src/web/page-scripts.ts
@@ -1267,7 +1267,8 @@ function refreshTasks(){
       var sl=scheduleLabel(task.schedule_type,task.schedule_value);
       var nr=task.next_run?relTime(task.next_run):"\\u2014";
       var lr=task.last_run?relTime(task.last_run):"\\u2014";
-      var lrc=task.last_result==="success"?"run-success":task.last_result==="error"?"run-error":"";
+      var los=task.last_outcome_state;
+      var lrc=los==="done"?"run-success":(los==="blocked"||los==="abandoned")?"run-error":los?"":task.last_result==="success"?"run-success":task.last_result==="error"?"run-error":"";
       return '<tr data-task-id="'+window.__esc(task.id)+'" data-status="'+window.__esc(task.status)+'">'
         +'<td><span class="badge '+sc+'">'+window.__esc(task.status)+'</span></td>'
         +'<td class="td-agent" title="'+window.__esc(task.chat_jid)+'">'+window.__esc(task.group_folder)+'</td>'
--- a/src/web/agent-detail.test.ts
+++ b/src/web/agent-detail.test.ts
@@ -267,20 +267,49 @@ describe('formatTaskLastRun', () => {
 });
 
 describe('lastRunOutcomeClass', () => {
-  it('maps success to run-success', () => {
-    expect(lastRunOutcomeClass('success')).toBe('run-success');
+  it('maps the done outcome state to run-success regardless of last_result', () => {
+    // Schedulers write summary strings like "Completed" / "Run 1 done" /
+    // a result excerpt into last_result. The colored badge needs to follow the
+    // normalized outcome state instead, otherwise those rows render uncolored
+    // even though the run clearly succeeded (#858).
+    expect(lastRunOutcomeClass('done', 'Completed')).toBe('run-success');
+    expect(lastRunOutcomeClass('done', 'Run 1 done')).toBe('run-success');
+    expect(lastRunOutcomeClass('done')).toBe('run-success');
   });
 
-  it('maps error to run-error', () => {
-    expect(lastRunOutcomeClass('error')).toBe('run-error');
+  it('maps the blocked outcome state to run-error', () => {
+    expect(lastRunOutcomeClass('blocked', 'Error: connection refused')).toBe(
+      'run-error',
+    );
+    expect(lastRunOutcomeClass('blocked')).toBe('run-error');
   });
 
-  it('returns empty string for null', () => {
+  it('maps the abandoned outcome state to run-error', () => {
+    expect(lastRunOutcomeClass('abandoned', 'Error: Execution timed out')).toBe(
+      'run-error',
+    );
+  });
+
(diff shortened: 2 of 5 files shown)


src/web/page-scripts.ts
```diff
@@ -1267,7 +1267,8 @@ function refreshTasks(){
       var sl=scheduleLabel(task.schedule_type,task.schedule_value);
       var nr=task.next_run?relTime(task.next_run):"\\u2014";
       var lr=task.last_run?relTime(task.last_run):"\\u2014";
-      var lrc=task.last_result==="success"?"run-success":task.last_result==="error"?"run-error":"";
+      var los=task.last_outcome_state;
+      var lrc=los==="done"?"run-success":(los==="blocked"||los==="abandoned")?"run-error":los?"":task.last_result==="success"?"run-success":task.last_result==="error"?"run-error":"";
```
Peyton-Spencer, line 1271: This mirrors the row color precedence, but the rebuilt client-side row still does not render the outcome badge that the SSR path appends via `renderOutcomeStateBadge()`. That matters more now because `skipped` intentionally returns a neutral row class and relies on the badge to convey the state. After any `/api/tasks` refresh, a skipped task will look like an ordinary uncolored row instead of showing the skipped badge. Consider rendering the same inline outcome badge here, including the `last_outcome_reason` title when present, so SSR and refresh output stay equivalent.

Peyton-Spencer (commented):
Reviewed the diff and ran targeted verification: `bun test src/web/tasks.test.ts src/web/agent-detail.test.ts` and `bun run tsc --noEmit` both pass. I left one client-refresh consistency comment inline.

PR #873, opened 2026-06-18
fix(security): block remote image redirect SSRF

Peyton-Spencer:
## Summary

- Disable automatic redirects for direct-IP remote image fetches.
- Add a regression test for a validated public IP redirecting to loopback.

## Security Impact

Remote image URL validation rejects loopback/private addresses before fetching, and hostname URLs already use a pinned no-redirect request path. Direct-IP URLs used default `fetch`, which follows redirects automatically. A public direct-IP avatar URL could therefore redirect the server-side fetch to a private address after validation.

This change keeps direct-IP fetches on `redirect: 'manual'`, so redirects are treated as non-OK upstream responses instead of fetching unvalidated targets.

## Verification

- `bun test src/web/image-cache.test.ts`
- `bun run tsc --noEmit`
- `bun test src/web/server.test.ts src/web/routes.test.ts src/web/image-cache.test.ts`

## Release Notes

* **Bug Fixes**
  * Enhanced image cache security by disabling automatic redirect handling during image fetching, preventing untrusted redirects to potentially unsafe addresses from being followed.

### DIFF
--- a/src/web/image-cache.test.ts
+++ b/src/web/image-cache.test.ts
@@ -156,6 +156,43 @@ describe('serveCachedRemoteImage', () => {
     }
   });
 
+  it('does not follow direct-IP redirects to unvalidated targets', async () => {
+    const testImageCacheDir = path.join(
+      DATA_DIR,
+      'image-cache-image-cache-test',
+      randomUUID(),
+    );
+    const fetchCalls: RequestInit[] = [];
+
+    clearTestImageCache(testImageCacheDir);
+
+    try {
+      const response = await serveCachedRemoteImage(
+        'redirect-cache-key',
+        async () => 'http://93.184.216.34/avatar.png',
+        {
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
No blocking findings.

I reviewed the remote image SSRF path with focus on direct-IP redirects. The hostname path continues to use pinned DNS via `fetchWithPinnedDns`, which does not auto-follow redirects, and the direct-IP path now passes `redirect: 'manual'` for both the injected fetch implementation and the production `fetch` call. A 30x response is therefore treated as a failed upstream fetch instead of fetching an unvalidated `Location` target.

Verified locally:
- `bun test src/web/image-cache.test.ts`
- `bun run tsc --noEmit`
- `bun test src/web/server.test.ts src/web/routes.test.ts src/web/image-cache.test.ts`
- `bun test`

PR #878, opened 2026-06-19
fix(security): cap simtest admin JSON bodies

Peyton-Spencer:
## Summary

- Replace uncapped `req.json()` calls in the simtest admin API with the shared capped JSON body reader.
- Return HTTP 413 for oversized JSON payloads before parsing.
- Add regression coverage for oversized admin API payload rejection.

## Security Impact

The simtest admin API is local-only by default, but it can be bound to another interface for demos or browser-based testing. Capping request bodies prevents oversized admin payloads from being fully buffered before JSON parsing.

## Verification

- `bun test src/simtest/admin-api.test.ts`
- `bun run tsc --noEmit`
- `git diff --check`

## Release Notes

* **Bug Fixes**
  * Admin API endpoints now enforce request body size limits, rejecting oversized payloads with HTTP 413 error responses.
  * Enhanced error handling with appropriate HTTP status codes: malformed JSON returns 400, oversized bodies return 413, and other errors return 500.

### DIFF
--- a/src/simtest/admin-api.test.ts
+++ b/src/simtest/admin-api.test.ts
@@ -384,6 +384,18 @@ describe('startAdminApi', () => {
       error: expect.stringContaining('JSON'),
     });
 
+    const oversized = await fetch(`${baseUrl}/agents`, {
+      method: 'POST',
+      body: JSON.stringify({
+        id: 'too-large',
+        padding: 'x'.repeat(1024 * 1024),
+      }),
+    });
+    expect(oversized.status).toBe(413);
+    expect(await readJson(oversized)).toMatchObject({
+      error: expect.stringContaining('Request body exceeded'),
+    });
+
     const missing = await fetch(`${baseUrl}/does-not-exist`);
     expect(missing.status).toBe(404);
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Reviewed the simtest admin API body-size hardening. I do not see blocking issues.

The change consistently replaces the admin mutation handlers' uncapped `req.json()` calls with the shared capped JSON reader, maps oversized payloads to 413 at the API boundary, and keeps malformed JSON behavior at 400. The new regression test covers the oversized path, and I also checked that no `req.json()` calls remain in `src/simtest/admin-api.ts`.

Verified locally:
- `bun test src/simtest/admin-api.test.ts`
- `bun run tsc --noEmit`

Peyton-Spencer (commented):
Reviewed the simtest admin API body-size cap. No blocking issues found.

The change consistently replaces admin JSON parsing with the shared bounded reader and maps `RequestBodyTooLargeError` to HTTP 413 while preserving malformed JSON as 400.

Verified locally:
- `bun test src/simtest/admin-api.test.ts src/request-body.test.ts`
- `bun run typecheck`

PR #875, opened 2026-06-19
feat(web): annotate dashboard active tasks card with running count

Peyton-Spencer:
## Summary

Annotates the dashboard's `active tasks` stat card with an inline `(N running)` count when any task lane is in flight, so the operator can tell at a glance whether enabled scheduled tasks are sitting idle or actually executing right now — without switching to `/ipc` or `/system`.

The annotation follows the existing `n (m)` parenthetical style already used by the `/ipc` retrying card and the new `/ipc` recent-events severity rollup (#860):

- `5` — no task lane currently running (unchanged)
- `5 (2 running)` — two of five enabled tasks are executing now

Pure render-layer change. `runningTasks` is computed from the existing `getQueueDetails()` snapshot by counting groups whose `taskLane.activeTask` is non-null — the same signal `/system` already exposes as `queue.running_tasks`. No `WebStateProvider`, schema, or SSE event changes.

## Changes

- `src/web/dashboard.ts` — pull `getQueueDetails()`, count groups with an active task, render `${activeTasks} (${runningTasks} running)` when `runningTasks > 0`, otherwise the bare count as before.
- `src/web/dashboard.test.ts` — `GroupQueueDetail` fixture helper plus two new cases: running-task annotation present, and annotation omitted when no lane is running. Existing zero-running case stays green.

## Test plan

- [x] `bun test src/web/dashboard.test.ts` — 7 pass.
- [x] `bun test src/web/` — 783 pass, 0 fail.
- [x] `bun test` — 2364 pass, 0 fail across the full suite.
- [x] `bun run typecheck` — clean.

Relates to #644 (operator observability rollup tracker).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Updated the dashboard “active tasks” card to show a running-task annotation (formatted as “(N running)”) alongside the active task count when in-flight tasks exist.

* **Tests**
  * Added new dashboard content test cases to confirm the running-task annotation appears when there are active task lanes and is omitted when only idle details are present.

PR #860, opened 2026-06-17
feat(web): surface error and warning counts on /ipc recent events card

Peyton-Spencer:
## Summary

Surfaces error and suppressed-event counts inline on the `/ipc` "recent events" stat card so operators can spot a spike of failures without scanning the 50-row timeline below.

The new annotation reuses the existing `1 (2)` parenthetical style from the `retrying` stat card on the same row:

- `50` — no error/warn events in window (unchanged)
- `50 (3 err)` — three error/blocked events present
- `50 (3 err, 2 warn)` — three errors plus two suppressed-style events

The per-row coloring logic that already classifies events into `event-error` / `event-warn` / `event-ok` is now refactored into a shared `classifyIpcEventSeverity` helper so the SSR render, the 5-second live poll in `page-scripts.ts`, and the new rollup count agree on a single classification rule.

Pure render-layer change — no `WebStateProvider`, schema, or SSE event changes.

## Changes

- `src/web/ipc-inspector.ts` — new `classifyIpcEventSeverity`, `countIpcEventSeverities`, and `formatRecentEventsValue` helpers; existing event-row classifier refactored to use them; stat card value now uses the rollup formatter.
- `src/web/page-scripts.ts` — mirrors the helpers in the `/ipc` live poll so the stat card and event-row coloring stay in sync after a poll refresh.
- `src/web/ipc-inspector.test.ts` — direct helper coverage (severity classifier, rollup counter, value formatter) plus integration assertions for the no-issue, error-only, and error+warn rendering branches.
- `docs/observability.md` — documents the new annotation on the `recent events` card in the `/ipc` stat-card table.

## Test plan

- [x] `bun test src/web/ipc-inspector.test.ts` — 44 pass.
- [x] `bun test src/web/page-scripts.test.ts` — 17 pass.
- [x] `bun test` — 2308 pass, 0 fail across the full suite.
- [x] `bun run typecheck` — clean.

Relates to #644 (operator observability rollup tracker).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Enhanced the "recent events" stat card to display error and warning counts inline (e.g., "(X err, Y warn)") alongside the total event count.

* **Documentation**
  * Updated observability reference to clarify the "recent events" card display format and suffix behavior.

* **Tests**
  * Added comprehensive test coverage for event severity classification and display formatting.

PR #880, opened 2026-06-20
feat(web): annotate /network trusted stat card with offline count

Peyton-Spencer:
## Summary

Annotates the `/network` `trusted` stat card with an inline `(N offline)` count when any trusted peer is unreachable, so an operator can tell at a glance whether their trusted LAN peers are actually online — without scanning the discovered-peers table or drilling into `/system → peers`.

The annotation follows the established `N (M qualifier)` convention already used by the dashboard active-tasks card (#875) and the IPC inspector recent-events rollup (#860):

- `5` — every trusted peer is online (unchanged)
- `5 (2 offline)` — three of five trusted peers are reachable right now

Pure render-layer derivation from the existing `peers` array. No new `WebStateProvider` hooks, schema, or SSE event shape — `peers[].status` and `peers[].online` already flow through every render path.

## Changes

- `src/web/network.ts`
  - `countTrustedOffline(peers)` — exported helper, counts peers with `status === 'trusted' && !online`.
  - `formatTrustedStatValue(peers)` — exported helper, returns the bare trusted count or appends ` (N offline)` when any trusted peer is offline. Used by both the SSR and SSE patch paths so they stay in sync.
  - `renderNetworkContent` — uses the helper for the `stat-peers-trusted` value cell.
- `src/web/server.ts` — `patchNetwork` SSE patch reuses the helper instead of inlining the trusted filter, so the Datastar update agrees with the SSR markup.
- `src/web/page-scripts.ts` — `refreshPeers` (the `/api/discovery/peers` polling fallback for clients without an active SSE subscription) computes `trustedOffline` and renders the same annotation, so the polled DOM never diverges from the SSE-patched DOM.
- `src/web/network.test.ts` — three new `describe` blocks (7 new tests):
  - `countTrustedOffline` — empty list, no trusted peers, mixed statuses (only trusted + offline counted).
  - `formatTrustedStatValue` — bare count, count with annotation, non-trusted peers ignored.
  - `renderNetworkContent trusted offline annotation` — annotation present in SSR HTML when offline trusted peers exist, omitted when all trusted peers are online.

## Test plan

- [x] `bun test src/web/network.test.ts` — 32 pass (7 new).
- [x] `bun test src/web/` — 798 pass, 0 fail across 24 web files.
- [x] `bun test` — 2379 pass, 0 fail across the full suite.
- [x] `bunx prettier --check src/web/network.ts src/web/network.test.ts src/web/server.ts src/web/page-scripts.ts` — clean.

Relates to #644 (operator observability rollup tracker).

## Summary by CodeRabbit

* **New Features**
  * Enhanced the trusted peers statistic to display the count of offline trusted peers inline (e.g., "5 (2 offline)").

* **Tests**
  * Added test coverage for new helpers that count offline trusted peers and format the trusted stat display.

### DIFF
--- a/src/web/network.test.ts
+++ b/src/web/network.test.ts
@@ -1,7 +1,9 @@
 import { describe, expect, it } from 'bun:test';
 
 import {
+  countTrustedOffline,
   formatPeerLastSeen,
+  formatTrustedStatValue,
   renderNetworkContent,
   renderNetworkPage,
   renderPeerOnlineCell,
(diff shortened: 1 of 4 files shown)


Peyton-Spencer (commented):
LGTM. I did not find any blocking issues.

Checked the SSR render path, SSE patch path, and `/api/discovery/peers` poll refresh path for the trusted offline annotation. The helper keeps server-side values consistent, and the client-side textContent update avoids HTML injection concerns.

Verified locally:
- `bun test src/web/network.test.ts src/web/page-scripts.test.ts`
- `bun run tsc --noEmit`

GitHub checks are green except CodeRabbit still pending at review time.

PR #884, opened 2026-06-21
feat(web): annotate dashboard agents card with working count

Peyton-Spencer:
## Summary

Annotates the dashboard's `agents` stat card with an inline `(N working)` count whenever any local agent lane is in flight, so the operator can tell at a glance how much of the fleet is actively doing work right now — without switching to `/agents` or `/system`.

The annotation follows the established `n (m)` parenthetical style already used by the `/ipc` retrying card and complements the in-flight #875 (active tasks running annotation):

- `5` — no agent lane currently in flight (unchanged)
- `5 (2 working)` — two of five agents are processing a message or running a scheduled task right now

An agent is counted as working when its `GroupQueueDetail` has either:
- `messageLane.active && !messageLane.idle` (actively processing a message), or
- `taskLane.active` (running a scheduled task)

Idle-waiting message lanes (`active=true`, `idle=true` — warm container in cooldown) are intentionally excluded; this matches `getAgentExecStatus` in `agents-page.ts`, which classifies those as `idle`. Remote peer agents are excluded because we do not track their lane state locally; the bare `agents` count keeps reflecting the total fleet size including remote peers.

Pure render-layer change. `workingAgents` is derived from the existing `getQueueDetails()` snapshot — the same signal `/system` already exposes. No `WebStateProvider`, schema, or SSE event changes.

## Changes

- `src/web/dashboard.ts` — pull `getQueueDetails()`, count groups whose message or task lane is actively in flight, render `${agentData.length} (${workingAgents} working)` when `workingAgents > 0`, otherwise the bare count as before.
- `src/web/dashboard.test.ts` — three `GroupQueueDetail` fixture helpers (`makeMessageRunningDetail`, `makeTaskRunningDetail`, `makeIdleDetail`) plus three new cases:
  - working annotation present when one message lane + one task lane are in flight (3 total agents → `3 (2 working)`)
  - annotation omitted when no lane is active
  - idle-waiting (cooldown) lanes are not counted as working

## Test plan

- [x] `bun test src/web/dashboard.test.ts` — 8 pass.
- [x] `bun test src/web/` — 801 pass, 0 fail.
- [x] `bun test` — 2382 pass, 0 fail across the full suite.

Relates to #644 (operator observability rollup tracker). Complementary to #875 (dashboard active tasks running annotation).

## Summary by CodeRabbit

* **New Features**
  * Dashboard agents stat card now displays the total number of agents and how many are currently working, shown in "X (Y working)" format when agents are active.

### DIFF
--- a/src/web/dashboard.ts
+++ b/src/web/dashboard.ts
@@ -18,12 +18,31 @@ export function renderDashboardContent(
   const stats = state.getQueueStats();
   const agentData = buildAgentChannelData(state, remotePeers);
   const tasks = state.getTasks();
+  const queueDetails = state.getQueueDetails();
 
   const activeContainers = Math.max(
     0,
     stats.activeContainers - stats.idleContainers,
   );
   const activeTasks = tasks.filter((t) => t.status === 'active').length;
+  // Count local agents currently in flight on either lane (processing a
+  // message or running a scheduled task). Surfaced inline on the "agents"
+  // card so the operator can tell at a glance how much of the fleet is
+  // actively doing work right now, without drilling into /agents or /system.
+  // Remote peer agents are excluded because we do not track their lane state
+  // locally; the bare count keeps reflecting the total fleet size.
+  const workingAgents = queueDetails.reduce(
+    (sum, g) =>
+      sum +
+      ((g.messageLane.active && !g.messageLane.idle) || g.taskLane.active
+        ? 1
+        : 0),
+    0,
+  );
+  const agentsValue =
+    workingAgents > 0
+      ? `${agentData.length} (${workingAgents} working)`
+      : `${agentData.length}`;
 
   // Serialize agent topology data for canvas renderer
   const topoData = escapeJsonForHtml(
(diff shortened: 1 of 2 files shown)


src/web/dashboard.ts
```diff
@@ -18,12 +18,31 @@ export function renderDashboardContent(
...
     stats.activeContainers - stats.idleContainers,
   );
   const activeTasks = tasks.filter((t) => t.status === 'active').length;
+  // Count local agents currently in flight on either lane (processing a
+  // message or running a scheduled task). Surfaced inline on the "agents"
+  // card so the operator can tell at a glance how much of the fleet is
+  // actively doing work right now, without drilling into /agents or /system.
+  // Remote peer agents are excluded because we do not track their lane state
+  // locally; the bare count keeps reflecting the total fleet size.
+  const workingAgents = queueDetails.reduce(
```
Peyton-Spencer, line 34: This should filter `queueDetails` to known local agent folders before reducing. `agentData.length` includes local + remote agents, but `GroupQueueDetail.folderKey` is only resolver-derived queue state and can fall back to a raw JID / stale key when no `jidToFolder` mapping exists. In that case the card can show impossible values like `1 (2 working)` or count work for a lane that is not represented in the agents denominator.

I’d derive a `Set` from `agentData.filter((a) => !a.remoteInstanceId).map((a) => a.folder)` and only count details whose `folderKey` is in that set. That also makes the “remote peers excluded from working count” contract explicit in code instead of relying on current queue behavior.

PR #907, opened 2026-06-25
feat(web): annotate /ipc groups tracked card with active count

Peyton-Spencer:
## Summary

Annotates the IPC inspector's `groups tracked` stat card with an inline `(N active)` count when any tracked group has a lane in flight, so the operator can tell at a glance how much of the tracked-group set is actually doing work right now — without scanning the group queue table below.

A group is counted as active when its queue detail has either:
- `messageLane.active && !messageLane.idle` (processing a message), or
- `taskLane.active` (running a scheduled task)

Idle-waiting message lanes (`active=true, idle=true` — warm container in cooldown) are intentionally excluded, matching the dashboard `agents (N working)` definition (#884) and `getAgentExecStatus`, which classifies cooldown as `idle`. This keeps the two surfaces consistent: an operator who sees `agents: 8 (3 working)` on `/` will see the same count reflected as the active subset of groups on `/ipc`.

The annotation follows the same parenthetical `n (m qualifier)` convention already used by:
- `/ipc` `retrying` — `${groups} (${total} retries)`
- `/ipc` `recent events` severity rollup (#860)
- dashboard `agents` — `${agents} (${working} working)` (#884)
- dashboard `active tasks` — `${tasks} (${running} running)` (#875)

Pure render-layer change. `activeGroups` is computed in the existing `queueDetails` loop alongside `pendingMessages`/`retryingGroups`. No `WebStateProvider`, schema, or SSE event changes.

## Changes

- `src/web/ipc-inspector.ts` — extend the existing `queueDetails` loop to count active groups, render `${queueDetails.length} (${activeGroups} active)` when `activeGroups > 0`, otherwise the bare count as before.
- `src/web/ipc-inspector.test.ts` — update the `shows correct stats` assertion for the new annotation on the existing sample fixture (`agent-alpha` is running, `agent-beta` is in cooldown → `2 (1 active)`), plus two new cases:
  - Annotation omitted when no group is active.
  - Idle-waiting (cooldown) message lanes are not counted as active.

## Test plan

- [x] `bun test src/web/ipc-inspector.test.ts` — 33 pass.
- [x] `bun test src/web/` — 811 pass, 0 fail.
- [x] `bun test` — 2394 pass, 0 fail across the full suite.

Relates to #644 (operator observability rollup tracker).

## Summary by CodeRabbit

* **New Features**
  * The IPC inspector now shows richer “groups tracked” stats, including how many groups are currently active when applicable.

* **Bug Fixes**
  * Fixed active group counting so idle or cooling-down lanes no longer inflate the active total.
  * Improved the stats display to omit the active-group note when nothing is actively running.

### DIFF
--- a/src/web/ipc-inspector.test.ts
+++ b/src/web/ipc-inspector.test.ts
@@ -128,8 +128,83 @@ describe('renderIpcInspector', () => {
     expect(html).toContain('1/5');
     // Idle
     expect(html).toContain('1/3');
-    // Groups tracked
-    expect(html).toContain('>2<');
+    // Groups tracked = 2 total, with agent-alpha actively running
+    // (message lane active+!idle, task lane active). agent-beta's
+    // message lane is in cooldown (active+idle) so it does not count.
+    expect(html).toContain('id="stat-groups">2 (1 active)<');
+  });
+
+  it('omits groups tracked annotation when no group is active', () => {
+    const detail: GroupQueueDetail = {
+      folderKey: 'agent-quiet',
+      messageLane: {
+        active: true,
+        idle: true,
+        pendingCount: 0,
+        containerName: 'ctr-quiet-msg',
+        reason: 'cooling-down',
+      },
+      taskLane: {
+        active: false,
+        pendingCount: 0,
+        containerName: null,
+        activeTask: null,
+        reason: 'no-work',
+      },
+      retryCount: 0,
+    };
+    const html = renderIpcInspector(
+      makeState({ getQueueDetails: () => [detail] }),
+    );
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
No blocking findings.

I sanity-checked the active-group definition against `getAgentExecStatus`: message lanes count only when `active && !idle`, task lanes count when `taskLane.active`, and cooldown lanes are excluded. That matches the existing dashboard `agents (N working)` rollup semantics.

Verified locally:
- `bun test src/web/ipc-inspector.test.ts src/web/dashboard.test.ts src/web/agents-page.test.ts` — 158 pass
- `bun run typecheck`

PR #912, opened 2026-06-25
test: cover registry persistence and image cache edges

Peyton-Spencer:
## Summary

- Add deterministic persistence coverage for `UserRegistryService.load/save` using temp-dir path injection.
- Harden `UserRegistryService.save()` to create its parent directory before writing, matching `load()` behavior.
- Add image cache edge coverage for malformed metadata refresh, no-body upstream responses, non-`Error` fetch failures, and stream cap cancellation/lock release.

## Test Count / Coverage

- Before: `2392 pass`, `0 fail`, `6376 expect()` calls, `91.34%` line coverage.
- After: `2399 pass`, `0 fail`, `6394 expect()` calls, `91.70%` line coverage.
- `src/effect/user-registry.ts`: `69.44%` -> `100.00%` line coverage.
- `src/web/image-cache.ts`: `95.20%` -> `96.72%` line coverage.

## Verification

- `bun test src/web/image-cache.test.ts src/effect/user-registry.test.ts`
- `bun test --coverage`
- `bunx tsc --noEmit`

## Remaining Coverage Gaps

- Channel adapters remain low coverage because they require broader Discord/Telegram/WhatsApp service mocks.
- Larger orchestrator paths in `src/index.ts` remain mostly integration-level and are better suited to simulation harness coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved user registry reliability when saving and loading data, including better handling of invalid stored data and support for looking up users with extra whitespace in search keys.
  * Strengthened image caching behavior so malformed cache data is treated as a miss, invalid responses are handled safely, and oversized streams are cleaned up correctly.

### DIFF
--- a/src/web/image-cache.test.ts
+++ b/src/web/image-cache.test.ts
@@ -333,6 +333,139 @@ describe('serveCachedRemoteImage', () => {
       logger.warn = originalWarn;
     }
   });
+
+  it('sanitizes non-Error fetch rejections without leaking implementation details', async () => {
+    const originalWarn = logger.warn;
+    const records: Array<Record<string, unknown>> = [];
+    const testImageCacheDir = path.join(
+      DATA_DIR,
+      'image-cache-image-cache-test',
+      randomUUID(),
+    );
+
+    clearTestImageCache(testImageCacheDir);
+    logger.warn = ((fieldsOrMsg: Record<string, unknown> | string) => {
+      if (typeof fieldsOrMsg !== 'string') {
+        records.push(fieldsOrMsg);
+      }
+    }) as unknown as typeof logger.warn;
+
+    try {
+      const response = await serveCachedRemoteImage(
+        'non-error-rejection-key',
+        async () => 'https://93.184.216.34/avatar.png?token=secret',
+        {
+          cacheDir: testImageCacheDir,
+          fetchImpl: (async () => {
+            throw 'raw network failure';
+          }) as RemoteImageFetch,
+        },
+      );
+
+      expect(response).toBeNull();
+      expect(records).toHaveLength(1);
+      expect(records[0]).toMatchObject({
+        imageUrl: 'https://93.184.216.34/avatar.png',
+        errorName: 'UnknownError',
+      });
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
Reviewed the path-injected UserRegistryService factory/save-directory behavior and the new deterministic image-cache edge tests. Registry path injection looks sound, save now creates the parent directory as intended, and local targeted tests plus typecheck pass. I left one non-blocking inline note on the stream cancellation test.

src/web/image-cache.test.ts
```diff
@@ -427,6 +554,24 @@ describe('readStreamWithCap', () => {
...
+      },
+      cancel() {
+        cancelled = true;
+      },
+    });
+
+    const result = await readStreamWithCap(stream, 100);
+
+    expect(result).toBeNull();
+    expect(cancelled).toBe(true);
```
Peyton-Spencer, line 578: One determinism gap in this edge test: `ReadableStreamDefaultReader.cancel()` returns a promise, but `readStreamWithCap()` currently calls it without awaiting it. This assertion proves the synthetic `cancel()` hook ran synchronously, but it would still pass if real cancellation cleanup were pending or if an async cancel rejection went unhandled. To make the edge fully deterministic, consider awaiting/catching `reader.cancel()` in `readStreamWithCap()` and returning a promise from this test's `cancel()` hook that must settle before the function returns, or narrow the test name/assertion to "invokes cancel" rather than "cancels".

Peyton-Spencer (commented):
Rebased onto current `main` and force-pushed `e224172`.

Resolution notes:
- Dropped the user-registry portion because newer registry persistence coverage and typed I/O error handling already landed on `main`.
- Kept the image-cache edge tests.
- Tightened the malformed-metadata test so it verifies the seeded `.bin`/`.json` cache pair is overwritten by the refresh.

Verified locally:
- `bun test src/web/image-cache.test.ts src/effect/user-registry.test.ts` — 72 pass / 0 fail
- `bun run typecheck` — pass

GitHub state: `CLEAN` / mergeable, all checks green, CodeRabbit green. No remaining blocking findings from me.

PR #913, opened 2026-06-25
fix(web): block cross-site unsafe requests

### LINKED ISSUE #905: Security: Web UI Basic Auth endpoints are CSRFable
## Summary

The Web UI exposes state-changing endpoints without an origin-bound CSRF defense. When an operator deploys the dashboard with HTTP Basic Auth (`WEB_UI_USER` / `WEB_UI_PASS`), browsers can attach cached Basic credentials to cross-site requests. A malicious page can therefore trigger side effects such as running a scheduled task by submitting a simple form to the dashboard origin.

The clearest endpoint is `POST /api/tasks/{id}/run`: it mutates state, requires no JSON body, and does not check a CSRF token, `Origin`, or `Sec-Fetch-Site`.

Session-cookie deployments are partially mitigated by `SameSite=Strict`, but Basic Auth deployments still rely on ambient browser credentials.

## Severity

P2 / Medium

Rationale: exploitation requires a victim operator who has authenticated to a Basic Auth-protected Web UI in the same browser, plus network reachability to the dashboard. Impact is meaningful because an attacker can cause agent/task actions without reading responses.

## Affected Code

- `src/web/server.ts`: Basic Auth grants access based only on the `Authorization` header; no CSRF/origin validation is applied before routing authenticated requests.
- `src/web/routes.ts`: mutating Web UI routes such as `POST /api/tasks/{id}/run` execute state changes without per-request CSRF tokens or origin checks.
- `src/web/session-auth.ts`: session-cookie mode uses `SameSite=Strict`, but that does not protect Basic Auth mode.

## Reproduction

1. Start OmniClaw with Web UI Basic Auth enabled via `WEB_UI_USER` / `WEB_UI_PASS`.
2. Log into the dashboard once in a browser so credentials are cached for the dashboard origin.
3. Create or identify an active scheduled task ID.
4. Visit an attacker-controlled page containing:

```html
<form method="POST" action="http://127.0.0.1:3000/api/tasks/TASK_ID/run">
  <button>click</button>
</form>
<script>document.forms[0].submit()</script>
```

5. The browser sends the cross-site POST with cached Basic credentials, and the task is queued via `state.runTaskNow(taskId)`.

The attacker cannot read the response due to browser same-origin policy, but CSRF does not require reading the response.

## Recommended Fix

Add CSRF protection for all state-changing Web UI routes.

Options:

- Prefer an origin-bound CSRF token for `POST`, `PATCH`, `PUT`, and `DELETE` routes. Render the token into the dashboard shell and require it in a custom header such as `X-OmniClaw-CSRF`.
- Additionally reject unsafe-method browser requests with suspicious or missing `Origin` / `Sec-Fetch-Site` headers unless they are trusted peer-auth requests.
- Keep peer-auth/API automation compatibility explicit: peer-auth routes already have HMAC authentication and should not silently rely on browser CSRF behavior.

## Duplicate Check

- `gh issue list --repo omniaura/omniclaw --state all --search "CSRF Web UI Basic Auth task run" --limit 20`
- `gh issue list --repo omniaura/omniclaw --state all --search "cross-site request forgery web dashboard" --limit 20`
- `gh issue list --repo omniaura/omniclaw --state all --search "Basic Auth CSRF dashboard" --limit 20`

Only #882 appeared in related searches, and that issue covers the separate dormant control-plane auth surface rather than the authenticated Web UI CSRF problem.

Peyton-Spencer:
## Summary

- Fixes #905.
- Blocks browser cross-site unsafe Web UI requests before authenticated route dispatch.
- Keeps HMAC-authenticated peer routes, same-origin dashboard requests, configured CORS origins, and non-browser API clients working.

## Security Notes

The vulnerable path was Basic Auth plus ambient browser credentials. A malicious site could submit a cross-site form POST to routes like `/api/tasks/{id}/run`, and the browser could attach cached Basic credentials.

This patch rejects unsafe non-peer requests when browser provenance headers indicate a cross-site request:

- `Sec-Fetch-Site: cross-site`
- An `Origin` header whose origin does not match the Web UI origin or configured `corsOrigin`

Requests without browser provenance headers remain allowed so existing CLI/API automation is not broken.

## Tests

- `bun run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Added protection against cross-site browser requests for unsafe actions.
  * Requests from untrusted origins now receive a clear `403` response, while valid same-origin requests continue to work.
  * Improved API error responses for blocked requests, with JSON output for API routes and plain-text output for other routes.
  * Added tests covering both blocked cross-site requests and allowed same-origin task runs.

### DIFF
--- a/src/web/server.test.ts
+++ b/src/web/server.test.ts
@@ -422,6 +422,58 @@ describe('basic auth', () => {
     expect(res.status).toBe(401);
   });
 
+  it('rejects cross-site browser POSTs even with valid Basic credentials', async () => {
+    let runCalls = 0;
+    handle = startWebServer(
+      testConfig(),
+      makeState({
+        runTaskNow: () => {
+          runCalls += 1;
+          return { ok: true };
+        },
+      }),
+    );
+
+    const res = await authedFetch('/api/tasks/task-001/run', {
+      method: 'POST',
+      headers: {
+        Origin: 'https://attacker.example',
+        'Sec-Fetch-Site': 'cross-site',
+      },
+    });
+
+    expect(res.status).toBe(403);
+    await expect(res.json()).resolves.toEqual({
+      error: 'Cross-site request blocked',
+    });
+    expect(runCalls).toBe(0);
+  });
+
+  it('allows same-origin browser POSTs with valid Basic credentials', async () => {
+    let runCalls = 0;
+    handle = startWebServer(
+      testConfig(),
+      makeState({
+        runTaskNow: () => {
+          runCalls += 1;
+          return { ok: true };
+        },
+      }),
+    );
+
+    const res = await authedFetch('/api/tasks/task-001/run', {
+      method: 'POST',
+      headers: {
+        Origin: `http://localhost:${handle.port}`,
+        'Sec-Fetch-Site': 'same-origin',
+      },
+    });
+
+    expect(res.status).toBe(200);
+    await expect(res.json()).resolves.toEqual({ ok: true, id: 'task-001' });
+    expect(runCalls).toBe(1);
+  });
+
   it('skips auth when no credentials are configured', async () => {
     handle = startWebServer(testConfig({ auth: undefined }), makeState());
     const pageRes = await fetch(url('/'));
--- a/src/web/server.ts
+++ b/src/web/server.ts
@@ -60,6 +60,7 @@ const PORT_ZERO_FALLBACK_START = 40000;
 const PORT_ZERO_FALLBACK_SPAN = 20000;
 const MAX_PEER_AUTH_BODY_BYTES = 1024 * 1024;
 const MAX_LOGIN_BODY_BYTES = 1024 * 1024;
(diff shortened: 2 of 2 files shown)


Peyton-Spencer (commented):
Security review complete. I did not find a blocking issue in this patch.

Reviewed:
- The new unsafe-method browser provenance gate in `src/web/server.ts`.
- Placement relative to Basic auth, session auth, trusted peer HMAC auth, SolidStart delegation, SSE/page routes, CORS preflight, and `handleRequest()` dispatch.
- Mutating route coverage in `src/web/routes.ts` and discovery routes.
- The new regression tests for cross-site Basic-authenticated POSTs and same-origin POST allowance.

Local verification:
- `bun run typecheck`

The approach is intentionally compatibility-preserving: cross-site browser mutations with `Sec-Fetch-Site: cross-site` or an untrusted `Origin` are blocked before route dispatch, authenticated peer routes remain HMAC-gated, and non-browser API clients without browser provenance headers continue to work.

PR #915, opened 2026-06-26
feat(web): annotate /tasks active card with overdue rollup

Peyton-Spencer:
## Summary

Annotates the `/tasks` page header's `active` stat with an inline `(N overdue)` count when one or more enabled tasks have a `next_run` in the past, so the operator can tell at a glance whether the scheduler is keeping up — without scanning the `next run` column row-by-row.

The annotation follows the existing parenthetical-rollup style already used by the dashboard `agents (N working)` chip (#884) and the `/agents` local count `(N disabled)` rollup:

- `5 active` — no overdue tasks (unchanged)
- `5 active (2 overdue)` — two of five active tasks are past `next_run`

Pure render-layer change. Counts are computed from the existing `getTasks()` already in scope — no `WebStateProvider`, schema, or SSE event changes. Tasks with `next_run = null` (paused/completed/once-fired) are skipped since "overdue" only makes sense relative to a scheduled time. Parse failures (`Date.parse` returning `NaN`) are treated as not-overdue to avoid noisy counts from legacy or malformed rows.

## Changes

- `src/web/tasks.ts` — derive `overdueActiveTasks` from `getTasks()` by counting `status === 'active'` rows whose parsed `next_run` is in the past, then render `${activeTasks} active (${overdueActiveTasks} overdue)` when `overdueActiveTasks > 0`, otherwise the bare count as before.
- `src/web/tasks.test.ts` — three new cases under `renderTasksContent`: annotation present when an active task is overdue (and paused-with-past-`next_run` is correctly ignored), annotation omitted when no active task is overdue (and `next_run: null` rows are skipped), and unparseable `next_run` values are treated as not-overdue.

## Test plan

- [x] `bun test src/web/tasks.test.ts` — 35 pass.
- [x] `bun test src/web/` — 812 pass, 0 fail.
- [x] `bun test` — 2395 pass, 0 fail across the full suite.

Relates to #644 (operator observability rollup tracker) and #505 (Web UI Phase 3 polish).

## Summary by CodeRabbit

* **New Features**
  * The task manager now shows overdue active tasks alongside the active task count.
  * When applicable, the active stat is displayed as `X active (Y overdue)` for quicker status checks.

* **Bug Fixes**
  * Overdue counts now exclude paused tasks.
  * Invalid or unparseable scheduled times are ignored, preventing incorrect overdue totals.
  * The overdue label is hidden when no active tasks are overdue.

### DIFF
--- a/src/web/tasks.ts
+++ b/src/web/tasks.ts
@@ -22,6 +22,27 @@ export function renderTasksContent(state: WebStateProvider): string {
   const executingTasks = tasks.filter(
     (t) => t.status !== 'completed' && t.executing_since != null,
   ).length;
+  // Count active tasks whose `next_run` has already passed. The scheduler
+  // normally advances `next_run` after each tick, so an active task with a
+  // past `next_run` means the scheduler is either backed up, was offline,
+  // or the run lease hasn't been picked up yet. Surfacing the rollup on the
+  // `active` chip answers "is the scheduler keeping up?" without making the
+  // operator scan the `next run` column row-by-row.
+  //
+  // Tasks with `next_run = null` (paused/completed/once-fired) are skipped
+  // since "overdue" only makes sense relative to a scheduled time. Parse
+  // failures are treated as not-overdue to avoid noisy counts from
+  // legacy/malformed rows.
+  const now = Date.now();
+  const overdueActiveTasks = tasks.reduce((sum, t) => {
+    if (t.status !== 'active' || !t.next_run) return sum;
+    const nextRunMs = Date.parse(t.next_run);
+    return Number.isFinite(nextRunMs) && nextRunMs < now ? sum + 1 : sum;
+  }, 0);
+  const activeValue =
+    overdueActiveTasks > 0
+      ? `${activeTasks} active (${overdueActiveTasks} overdue)`
+      : `${activeTasks} active`;
 
   // Build agent/channel options for create/edit forms
   const agentOptions = agentData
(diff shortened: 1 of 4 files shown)


src/web/tasks.ts
```diff
@@ -44,7 +65,7 @@ export function renderTasksContent(state: WebStateProvider): string {
     // Stats row
     `<div class="tasks-stats">` +
     `<span class="tasks-stat">${tasks.length} total</span>` +
-    `<span class="tasks-stat stat-active">${activeTasks} active</span>` +
+    `<span class="tasks-stat stat-active">${activeValue}</span>` +
```
Peyton-Spencer, line 71: This server-rendered value gets overwritten by the client refresh path. `refreshTasks()` in `src/web/page-scripts.ts` rebuilds `.tasks-stats` as `active + " active"` after create, edit, pause/resume, or delete, so an overdue annotation can disappear after ordinary in-page task actions until a full reload. Please mirror this overdue calculation there as well, or centralize the stats rendering so SSR and refresh stay in sync. A page-scripts test for the refreshed stats would catch the regression.

Peyton-Spencer (commented):
Found one consistency issue before this lands. I do not think a grace period is necessary for this PR: the scheduler already treats `next_run <= now` as due and advances `next_run` before enqueue, so a past `next_run` is the right simple signal for backlog/offline/lease-not-picked-up state.

src/web/tasks.ts
Peyton-Spencer: Good catch — fixed in f9d76bd.

`refreshTasks()` now mirrors the SSR logic: counts active tasks with `Date.parse(next_run) < now` and renders the `(N overdue)` suffix in the rebuilt `.tasks-stats`. Added a `page-scripts.test.ts` case asserting the rollup expression and HTML are present.

Agreed on skipping the grace period — `getDueTasks()` already treats `next_run <= now` as due and the scheduler advances `next_run` before enqueue, so the chip won't flap on a healthy run.

Full suite: 2396 pass / 0 fail.

Issue #920, opened 2026-06-26
Research: hermes Slack patterns (#759)


Peyton-Spencer:
## Research: hermes Slack patterns — reference for #759

This issue summarizes a deep dive into how the [NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent) project implements native Slack slash command support via Bolt Socket Mode, and what we should borrow (and what we should avoid) when implementing the same feature in OmniClaw against issue #759.

Closes/links #759.

---

### What "hermes" is here

Two candidate "hermes" projects exist; only one is relevant to Slack slash command work:

- *Not relevant*: [`ditto-assistant/ditto-hermes`](https://github.com/ditto-assistant/ditto-hermes) (cloned at `/workspace/group/ditto-hermes/`). This is a SKILL bundle for the Hermes Agent that wires the heyditto CLI memory MCP into Hermes — nothing to do with Slack/Bolt routing.
- *Relevant*: [`NousResearch/hermes-agent`](https://github.com/NousResearch/hermes-agent) — a multi-channel, learning agent that "lives where you do: Telegram, Discord, Slack, WhatsApp, Signal, and CLI." It is the closest open-source analog to OmniClaw, including how it maps a single registry of commands onto Slack slash commands.

Primary sources investigated:

- [`plugins/platforms/slack/adapter.py`](https://github.com/NousResearch/hermes-agent/blob/main/plugins/platforms/slack/adapter.py) — Bolt Socket Mode adapter (~4.3k LOC).
- [`plugins/platforms/slack/plugin.yaml`](https://github.com/NousResearch/hermes-agent/blob/main/plugins/platforms/slack/plugin.yaml) — required env vars, scopes posture.
- [`hermes_cli/commands.py`](https://github.com/NousResearch/hermes-agent/blob/main/hermes_cli/commands.py) — single `COMMAND_REGISTRY` that drives Slack/Discord/Telegram in parallel; helper functions `slack_native_slashes()`, `slack_subcommand_map()`, `slack_app_manifest()`.
- [`website/docs/user-guide/messaging/slack.md`](https://github.com/NousResearch/hermes-agent/blob/main/website/docs/user-guide/messaging/slack.md) — operator-facing setup, including manifest generation, scopes, and the "Messages Tab" foot-gun.

---

### Specific patterns worth borrowing

[...]

Issue #919, opened 2026-06-26
Research: Anthropic @claude tag + Slack patterns (#759)


Peyton-Spencer:
## Research: Anthropic @claude tag, claude-code-action, and Claude-Code-on-Slack patterns

Background research for #759 (Implement Slack slash command support). This issue summarizes how Anthropic, `anthropics/claude-code-action`, and Anthropic-adjacent Slack samples (Bolt JS starter agent, Claude managed-agents data-bot cookbook) wire `@claude` mentions and slash commands, and recommends what to borrow — and explicitly what to avoid — for OmniClaw's Slack command surface.

OmniClaw context being mapped against:
- `src/channels/slack.ts` currently registers only `app.message`, `app.event('app_mention')`, `app.event('reaction_added')` (lines 247, 256, 264) — no `app.command(...)` handlers.
- Discord already routes slash command invocations to the agent via `onSyntheticMessage` in `src/channels/discord.ts` (~line 1470).
- Discord builtin flows in `src/discord-command-flows.ts` define `mergemaster`, `product-driver`, `issue-driver`, `research-driver`, `implement-driver`, `qa-driver`, `taskbooker`, `scheduler`, `/session` etc. as `{name, description, prompt, options[], system?}` records that get rendered into a prompt string and queued.

### `anthropics/claude-code-action` (GitHub Action)
- Default trigger is the literal substring `@claude`. Configurable via `trigger_phrase` input (e.g. `/claude`), plus two issue-only triggers: `assignee_trigger` (assign the issue to a named user) and `label_trigger` (apply a label).
- Trigger detection lives in `src/github/validation/trigger.ts`. Priority order:
  1. Explicit `prompt` input → always triggers (used by automation workflows).
  2. Assignee match on `issues.assigned`.
  3. Label match on `issues.labeled`.
  4. Trigger-phrase regex scan over: `issue.body`, `issue.title`, `pull_request.body`, `pull_request.title`, `review.body`, `comment.body`.
- Phrase regex is `(^|\s)<phrase>([\s.,!?;:]|$)` (case-insensitive) — i.e. word-bounded; "fixme" does not match "fix".
- Output `contains_trigger` (string boolean) gates the next workflow step. Mention parsing and prompt extraction are deferred to the Claude Code runtime.
- Mode is auto-detected (mention/assignment/automation) from event context — no explicit `mode` input in v1.

### Claude Code in Slack (Anthropic, official docs)
[...]

Issue #918, opened 2026-06-26
Research: nanoclaw Slack slash command patterns (#759)


Peyton-Spencer:
## Research: nanoclaw Slack slash command patterns

Closes / drives #759. This is the research deliverable that should precede the implementation PR for Slack slash command support. The TL;DR is **nanoclaw provides effectively no reusable pattern for Slack slash commands** — its v2 adapter is a thin wrapper around a third-party Chat SDK, and its v1-style adapter (still present in OmniClaw's fork) only wires `app.message`, `app.event('app_mention')`, and `app.event('reaction_added')`. The reusable design knowledge lives in our own `discord-command-flows.ts` plus the Bolt JS docs.

### 1. What nanoclaw upstream does for Slack slash commands

**Short answer: nothing direct.** Upstream nanoclaw v2 has two parallel facts:

- Trunk no longer ships a hand-written Slack adapter. The actual adapter on the `channels` branch (`origin/channels:src/channels/slack.ts`, 39 lines) just delegates to `@chat-adapter/slack` via the Chat SDK bridge. There is no `app.command(...)` anywhere — the file does not even import `@slack/bolt` directly. Socket Mode is opt-in via `SLACK_APP_TOKEN` and is passed through to the adapter as `mode: 'socket' | 'webhook'`.
- The setup flow (`/workspace/group/nanoclaw-upstream/setup/channels/slack.ts:128-179`) instructs the user to enable "Allow users to send slash commands and messages from the messages tab" in App Home, but only as part of the AI-app assistant pane experience — *not* to register any custom `/foo` slash commands. The post-install checklist (lines 419-448) wires Event Subscriptions and Interactivity request URLs, never a Slash Commands request URL.

Every occurrence of "slash command" in nanoclaw (`origin/main` and `origin/channels`) refers to either:
- Agent-runner-level message prefix routing (e.g. `/clear`, `/reset`, `/approve`, `/reject`) detected by inspecting the message body — see `container/agent-runner/src/formatter.ts` and `src/command-gate.ts`.
- Provider-native slash commands handled by the Claude Agent SDK inside the container (`container/agent-runner/src/poll-loop.ts`, `providers/types.ts`).

Neither uses Slack's `app.command` API. Nanoclaw effectively punts Slack slash command UX to either the third-party `@chat-adapter/slack` package or to the message body.

### 2. Specific code references

Upstream nanoclaw (`/workspace/group/nanoclaw-upstream/`):

[...]

Issue #917, opened 2026-06-26
Research: openclaw Slack patterns (#759)


Peyton-Spencer:
## Research: openclaw Slack patterns (for #759)

This is a research/notes issue collecting patterns from the upstream-ish `openclaw/openclaw` project to inform our Slack slash command implementation for OmniClaw. Closes #759 when we ship the implementation PR; until then this issue is the design notes.

### What I found

- Project: `openclaw/openclaw` — https://github.com/openclaw/openclaw — a large TS personal-assistant framework. omniaura's archived `nanoclaw` repo describes itself as "A lightweight alternative to Clawdbot / OpenClaw", so OpenClaw is effectively the senior sibling: same Bolt/Socket-Mode + Claude Agent SDK design, much more battle-tested.
- Stack overlap: TypeScript, `@slack/bolt`, Socket Mode, Claude-style agent dispatch. Multi-account ("multi-bot") routing is first-class. License: MIT.
- Slack provider lives in `extensions/slack/` (a bundled plugin); slash commands specifically in `extensions/slack/src/monitor/slash.ts`.

### Key code references (use these as the canonical map)

- `extensions/slack/src/monitor/slash.ts` — full slash command registration + handler. Single function `registerSlackMonitorSlashCommands({ ctx, account, trackEvent })`. ~1037 lines but the moving parts are small. Worth reading top-to-bottom once.
  - L371-393: capability sniffing — checks for `app.action` / `app.options` to decide whether interactive arg menus and external_select are supported. Lets the same code run against a stubbed Bolt in tests.
  - L388-414: `handleSlashCommand` core. Drops mismatched team/app events early (`shouldDropMismatchedSlackEvent`), `ack`s empty-prompt commands with ephemeral "Message required.", then `ack()`s, then continues async.
  - L417-419: Self-mute — ignores commands where `command.user_id === ctx.botUserId`.
  - L421-525: Auth/policy gauntlet — resolves channel info, normalizes channel type, then runs three concentric checks: `isChannelAllowed`, `authorizeSlackDirectMessage` (DM pairing flow), and `useAccessGroups` policy. Each rejection sends an ephemeral response and returns.
  - L527-549: `resolveSlackCommandIngress` produces a typed `senderAccess.gate` + `commandAccess.authorized` pair. Rooms can deny outright; DMs only set `commandAuthorized` for downstream gating.
[...]

PR #921, opened 2026-06-26
feat(slack): slash command support with Discord flow parity (#759)

### LINKED ISSUE #759: Implement Slack slash command support
## Summary

OmniClaw appears to have Discord slash command support, but not Slack slash command support. The Slack channel currently registers message and reaction handlers only.

## Evidence

- `src/channels/slack.ts` registers `this.app.message(...)` and `this.app.event('reaction_added', ...)` but no `this.app.command(...)` handlers.
- `src/channels/discord.ts` has Discord slash command syncing and interaction handling.
- `src/discord-command-flows.ts` and related tests implement command definitions and filtering for Discord flows, with no Slack equivalent found.

## Why This Matters

Now that Slack multi-bot setup is active, users will expect equivalent bot command affordances in Slack, especially for host-handled commands like session controls and workflow/custom agent commands.

## Suggested Scope

- Add Slack slash command registration/handling in `SlackChannel` using Bolt `app.command(...)`.
- Decide command namespace and parity model with Discord, e.g. `/session`, workflow commands, and custom command definitions.
- Route commands to the same host/agent flow machinery where possible.
- Add tests for Slack command parsing, acknowledgement, routing, and permission/error behavior.

## Notes

Tried to use local `gh`, but the local GitHub CLI token is currently invalid:

```text
gh auth status
X Failed to log in to github.com account Peyton-Spencer
The token in /Users/peyton/.config/gh/hosts.yml is invalid.
```

Filed via the authenticated GitHub connector instead.

Peyton-Spencer:
## Summary

Implements Slack slash command support (closes #759) with the same surface Discord exposes — `/session` plus the builtin flows (`/research-driver`, `/implement-driver`, `/issue-driver`, `/mergemaster`, `/product-driver`, `/qa-driver`, `/scheduler`, `/taskbooker`). Reuses the existing flow registry in `src/discord-command-flows.ts` so per-group custom commands work on both channels with one source of truth.

## How it works

Slack has no runtime command-registration API, so the manifest in `config-examples/slack-app-manifest.json` is the source of truth — `bun start` only wires handlers. A single `app.command(/.*/)` matcher dispatches every command through the shared flow registry. `text` is parsed into typed option values (`key=value` overrides, POSIX-ish quoting, free-form text falls into the first required string option so `/research-driver investigate the foo bug` works naturally).

- **Session commands** (`/session new|resume|list|current|end|rename`, plus the legacy `/resume` and `/sessions` aliases) run through the same `handleSessionCommand` Discord uses and reply ephemerally.
- **Flow commands** render the prompt template and push a synthetic `NewMessage` through `onSyntheticMessage`, mirroring the Discord path.
- **Admin gate** — `/session` commands require workspace-admin status (parity with Discord's `Manage Channels` requirement).
- **Multi-bot** — JID scoping uses `channelIdToJid(channel_id, this.multiBotMode ? this.botId : undefined)` so commands route to the right agent.

## Research basis

Four researcher agents filed tracking issues with patterns from each reference target:

- #917 — openclaw Slack patterns (manifest, separate slash module, fast-ack, lazy dispatch)
- #918 — nanoclaw Slack patterns (manifest-driven, command-text subcommand parsing, ack flow)
- #919 — Anthropic @claude tag + Slack patterns (mention vs slash, thread = session, invite-based access)
- #920 — hermes Slack patterns (single regex matcher over N decorated handlers, manifest 50-command cap, two-phase ephemeral reply)

Convergent guidance across all four: single regex command matcher, manifest as source of truth, reuse the existing Discord flow registry, ephemeral-by-default replies, dispatch into the existing synthetic-message path. Implemented exactly that.

## Files

- `src/slack-command-flows.ts` — text parser + manifest payload builder (channel-agnostic helpers built on the existing flow types)
- `src/channels/slack.ts` — `handleSlashCommand` + `handleSlashSystemCommand`, `/session` subcommand dispatch including legacy aliases, admin gate, regex command listener
- `src/index.ts` — wires `slashCommandGroups`, `onSyntheticMessage`, `onSessionCommand` into `SlackChannel`
- `config-examples/slack-app-manifest.json` — adds the `slash_commands` block listing every command
- `docs/SLACK-AGENT-SETUP.md` — new "Slash commands" section documenting the surface and admin requirement
- Tests: `src/slack-command-flows.test.ts` (10 cases) + `src/channels/slack.test.ts` (16 new slash command cases)

## Verification

- `bunx tsc --noEmit` — clean
- `bun test` — 2,496 pass, 0 fail (26 new tests)
- `bunx prettier --check .` — passes on every touched file

## Operator action required

After pulling this in, redeploy your Slack app manifest from `config-examples/slack-app-manifest.json` so Slack picks up the new `/session` and `/<flow>` commands. Per-group custom commands defined in `groups/<folder>/discord-commands.json` will also need matching entries in the Slack app manifest until we ship a manifest generator.

## Out of scope (follow-ups)

- Block Kit modal flows for option-heavy commands (deferred per the openclaw + hermes research)
- A `bun run slack:manifest` script to emit the manifest entries from group flow files (referenced in #920)
- DM-only slash command UX (Slack delivers `channel_name = directmessage`; current implementation requires the DM to be registered as an OmniClaw channel first)

## Test plan

- [ ] Update Slack app manifest from `config-examples/slack-app-manifest.json` in a real workspace.
- [ ] Run `/research-driver investigate the foo bug` in a registered Slack channel → ephemeral ack, synthetic message reaches the agent runtime.
- [ ] Run `/session list` as a workspace admin → ephemeral session list rendered by `handleSessionCommand`.
- [ ] Run `/session list` as a non-admin → ephemeral "restricted to workspace admins" message, no host call.
- [ ] Run `/research-driver` in an unregistered channel → ephemeral "channel is not registered" message.
- [ ] Run `/research-driver goal="ship the thing" repo=omniclaw duration_minutes=15` → key=value overrides land in the synthetic prompt; integer is coerced.
- [ ] In multi-bot setup, confirm `/session list` issued under bot A doesn't route to bot B (JID scoping).

## Summary by CodeRabbit

* **New Features**
  * Added Slack slash command support, including multiple built-in commands and per-command usage hints.
  * Slack commands can now be parsed into structured options, including quoted text and typed values.
  * Flow-based commands now generate manifest entries automatically and support synthetic message handling.
  * Session-related commands such as `/session` and legacy aliases are now supported with admin checks.

* **Bug Fixes**
  * Improved routing and validation for unknown commands, unregistered channels, and command parsing edge cases.

* **Documentation**
  * Added setup guidance for configuring and extending Slack slash commands.

### DIFF
--- a/docs/SLACK-AGENT-SETUP.md
+++ b/docs/SLACK-AGENT-SETUP.md
@@ -55,6 +55,21 @@ If you already have a Slack app from before PR #829 landed:
 
 The non-assistant code paths keep working with the old manifest; you just won't see the agent pane until you reinstall.
 
+## Slash commands
+
+Slack slash commands are declared in the app manifest (Slack has no runtime registration API like Discord). The manifest in `config-examples/slack-app-manifest.json` ships the same flow surface as Discord: `/session`, `/research-driver`, `/implement-driver`, `/issue-driver`, `/mergemaster`, `/product-driver`, `/qa-driver`, `/scheduler`, `/taskbooker`.
+
+Each command lands in `SlackChannel.handleSlashCommand` which:
+
+- Acks immediately (Slack's 3s deadline).
+- Resolves the channel to a registered OmniClaw group.
+- For session commands (`/session`, plus legacy `/resume` / `/sessions`), runs the host session handler and replies ephemerally.
+- For flow commands, parses the trailing `text` (`key=value` overrides plus free-form text into the first required option), renders the prompt template, and pushes a synthetic message through the same `onSyntheticMessage` path the Discord adapter uses.
+
+To define custom slash commands per group, add the entries to your Slack app manifest with the same names you already use in `groups/<folder>/discord-commands.json` — the host reuses the same flow registry for both channels.
+
+Session commands (`/session …`) require workspace-admin status on Slack — parity with Discord's `Manage Channels` requirement.
+
 ## Multi-bot installs
 
 For workspaces running multiple OmniClaw bots (the `SLACK_BOT_IDS` / `SLACK_BOT_*_TOKEN` mode in `.env.example`), create one Slack app per bot from the same manifest. Tweak `display_information.name`, `features.bot_user.display_name`, and `features.assistant_view.assistant_description` per bot so each shows up distinctly in the AI Apps picker.
--- a/src/channels/slack.ts
+++ b/src/channels/slack.ts
@@ -1025,18 +1025,30 @@ export class SlackChannel implements Channel {
   }
 
   /**
-   * Look up a flow by name from the configured slash command groups, falling
-   * back to the broader registered groups list so manifest-only commands
-   * still work in dev setups.
+   * Look up a flow by name from the configured slash command groups. When
+   * `slashCommandGroups` is supplied the bot operator has explicitly narrowed
+   * which groups this bot may run commands for (subscription/multi-bot
+   * scoping), so a group that isn't in the eligible list must not be allowed
+   * to fall back to its own flow definitions — that would defeat the gate.
+   * Falls back to the broader registered groups list only when no narrowing
+   * is configured (single-bot dev setups).
    */
   private findFlowForGroup(
     group: RegisteredGroup,
     commandName: string,
   ): DiscordFlowDefinition | undefined {
-    const eligible = this.opts.slashCommandGroups
-      ? this.opts.slashCommandGroups()
-      : Object.values(this.opts.registeredGroups());
-    const scoped = eligible.find((g) => g.folder === group.folder) || group;
+    let scoped: RegisteredGroup | undefined;
+    if (this.opts.slashCommandGroups) {
+      scoped = this.opts
+        .slashCommandGroups()
+        .find((g) => g.folder === group.folder);
+      if (!scoped) return undefined;
+    } else {
+      scoped =
+        Object.values(this.opts.registeredGroups()).find(
+          (g) => g.folder === group.folder,
+        ) || group;
+    }
     return getDiscordFlowDefinitionsForGroup(scoped).find(
       (flow) => flow.name === commandName,
     );
@@ -1046,8 +1058,11 @@ export class SlackChannel implements Channel {
    * Slack permission gate for session commands. We approximate Discord's
    * "Manage Channels" requirement with workspace-admin status — easier to
    * reason about across channel types and aligns with how operators set up
-   * Slack workspaces. Soft-fails open on API errors so a transient outage
-   * doesn't lock everyone out (logged at warn).
+   * Slack workspaces. Fails CLOSED on lookup errors so a transient outage
+   * cannot grant elevated session-management access to a non-admin (the
+   * Discord side hard-blocks too via `memberPermissions`, and the session
+   * commands mutate runtime state — privilege escalation, not availability,
+   * is the worse failure mode).
    */
   private async userIsWorkspaceAdmin(userId: string): Promise<boolean> {
     if (!userId) return false;
@@ -1060,8 +1075,11 @@ export class SlackChannel implements Channel {
         user?.is_admin || user?.is_owner || user?.is_primary_owner,
       );
     } catch (err) {
-      logger.warn({ err, userId }, 'Slack users.info failed for admin check');
-      return true;
+      logger.warn(
+        { err, userId },
+        'Slack users.info failed for admin check — denying',
+      );
+      return false;
     }
   }
 
--- a/src/channels/slack.ts
+++ b/src/channels/slack.ts
@@ -17,9 +17,19 @@ import {
   isTextByExtension,
   readStreamWithByteLimit,
 } from '../media.js';
+import {
+  getDiscordFlowDefinitionsForGroup,
+  renderDiscordFlowPrompt,
+  SESSION_COMMAND_NAMES,
+  SYSTEM_COMMAND_NAMES,
+  type DiscordFlowDefinition,
+  type DiscordSessionCommand,
+} from '../discord-command-flows.js';
 import { parseScopedSlackJid } from '../slack-jid.js';
+import { parseSlackCommandText } from '../slack-command-flows.js';
 import type {
   Channel,
+  NewMessage,
   OnChatMetadata,
   OnInboundMessage,
   OutboundMessageStream,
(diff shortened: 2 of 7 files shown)


src/channels/slack.ts
```diff
@@ -610,6 +681,390 @@ export class SlackChannel implements Channel {
...
+   * still work in dev setups.
+   */
+  private findFlowForGroup(
+    group: RegisteredGroup,
+    commandName: string,
+  ): DiscordFlowDefinition | undefined {
+    const eligible = this.opts.slashCommandGroups
+      ? this.opts.slashCommandGroups()
+      : Object.values(this.opts.registeredGroups());
+    const scoped = eligible.find((g) => g.folder === group.folder) || group;
```
Peyton-Spencer: This fallback bypasses the `slashCommandGroups` boundary that `index.ts` now passes in from channel subscriptions. In multi-bot or subscription-scoped setups, a registered channel that is *not* in this bot's eligible slash-command groups still falls back to `group`, loads that group's built-in/custom flows, and queues the command through this SlackChannel. That defeats the intended per-bot/per-channel command scoping described above. If `slashCommandGroups` is provided, absence from `eligible` should return `undefined` instead of falling back; keep the broader registered-groups fallback only for the legacy case where the callback is omitted. A regression test should cover a registered group excluded from `slashCommandGroups` being rejected.

src/channels/slack.ts
```diff
@@ -610,6 +681,390 @@ export class SlackChannel implements Channel {
...
+      const info = await this.client.users.info({ user: userId });
+      const user = info.user as
+        | { is_admin?: boolean; is_owner?: boolean; is_primary_owner?: boolean }
+        | undefined;
+      return Boolean(
+        user?.is_admin || user?.is_owner || user?.is_primary_owner,
+      );
+    } catch (err) {
+      logger.warn({ err, userId }, 'Slack users.info failed for admin check');
+      return true;
```
Peyton-Spencer: This should fail closed. `/session` can create/resume/end/rename host sessions, and the docs promise workspace-admin-only access. If `users.info` fails because the token is missing `users:read`, Slack is transiently unavailable, rate limited, or returns another API error, any Slack user can run the session command. Please return `false` on lookup failure and add a test where `users.info` rejects and `onSessionCommand` is not called.

Peyton-Spencer (commented):
Thanks for pulling the Slack slash-command surface together. I found two authorization/scoping blockers that should be fixed before merge.

Verified locally: `bun test src/slack-command-flows.test.ts src/channels/slack.test.ts` -> 80 pass. GitHub CI is green for format/typecheck/test/agent-runner-typecheck; CodeRabbit was still pending when I reviewed.

src/channels/slack.ts
Peyton-Spencer: This gate should fail closed, not open. If `users.info` fails because the token is missing `users:read`, Slack is transiently unavailable, or the API returns an unexpected error, every Slack user who can invoke `/session` gets host-level session controls. These commands can end/resume/rename channel sessions, so a fail-open admin check is a privilege escalation path.

Please return `false` on lookup failure and surface an ephemeral "could not verify admin status" response, while keeping the warn log for operators.

src/channels/slack.ts
```diff
@@ -610,6 +681,390 @@ export class SlackChannel implements Channel {
...
+      });
+      return;
+    }
+
+    const parsed = parseSlackCommandText(command.text || '', flow);
+    const renderedPrompt = renderDiscordFlowPrompt(
+      flow,
+      parsed.optionValues,
+    ).trim();
+    if (!renderedPrompt) {
```
Peyton-Spencer: Slack needs an explicit required-option check before queueing the synthetic message. Right now `/research-driver` with an empty `text` renders the built-in prompt with `{{goal}}` replaced by `''`, which is still a non-empty prompt, so it queues a driver run with a blank target instead of returning the `needs more input` response. Discord gets this validation from the platform UI, but Slack's free-text payload bypasses that.

Please reject when any `flow.options` entry with `required: true` is still `undefined` after `parseSlackCommandText`, and add a test for `/research-driver` with no text.

docs/SLACK-AGENT-SETUP.md
```diff
@@ -55,6 +55,21 @@ If you already have a Slack app from before PR #829 landed:
...
 
+## Slash commands
+
+Slack slash commands are declared in the app manifest (Slack has no runtime registration API like Discord). The manifest in `config-examples/slack-app-manifest.json` ships the same flow surface as Discord: `/session`, `/research-driver`, `/implement-driver`, `/issue-driver`, `/mergemaster`, `/product-driver`, `/qa-driver`, `/scheduler`, `/taskbooker`.
+
+Each command lands in `SlackChannel.handleSlashCommand` which:
+
+- Acks immediately (Slack's 3s deadline).
+- Resolves the channel to a registered OmniClaw group.
+- For session commands (`/session`, plus legacy `/resume` / `/sessions`), runs the host session handler and replies ephemerally.
```
Peyton-Spencer, line 66: The docs say `/resume` and `/sessions` are supported legacy aliases, and `handleSlashSystemCommand` implements them, but Slack will never deliver those commands unless they are also declared in `features.slash_commands`. As shipped, Slack rejects `/resume` and `/sessions` at the gateway before Bolt reaches the regex matcher.

Either add both aliases to the manifest or remove the alias claim from the Slack docs/PR surface.

Peyton-Spencer (commented):
Requesting changes based on the inline comments.

Validated locally:

```bash
bun test src/slack-command-flows.test.ts src/channels/slack.test.ts
```

Result: 80 pass, 0 fail.

Main blockers:

- Slack flow commands do not enforce required options, so `/research-driver` with no text queues a blank-goal agent run.
- The Slack workspace-admin gate fails open on `users.info` errors, granting session controls when admin status cannot be verified.

Also noted a manifest/docs mismatch for the `/resume` and `/sessions` aliases.

Peyton-Spencer (commented):
Follow-up review on `3adde7e`: my two blockers are addressed.

- `findFlowForGroup()` now fails closed when `slashCommandGroups` is supplied and the resolved channel group is not eligible, preserving subscription/multi-bot scoping.
- `userIsWorkspaceAdmin()` now denies on `users.info` lookup errors, so `/session` no longer fails open.
- Regression tests cover both cases.

Verified locally: `bun test src/slack-command-flows.test.ts src/channels/slack.test.ts` -> 82 pass / 0 fail.

GitHub will not let this token formally approve because it is the PR author identity, but from my review perspective the prior blockers are cleared.

Peyton-Spencer (commented):

Verified the review blockers against current head `8b74743`:
- Required-option enforcement: `/research-driver` (and any flow with a `required` option lacking a default) now returns the "needs more input" ephemeral and does **not** queue a synthetic message when the option is unset — fixes the empty-`{{goal}}`-still-renders-non-empty-prompt gap. Regression test added.
- `userIsWorkspaceAdmin()` fails closed (`return false`) on `users.info` errors.
- `findFlowForGroup()` fails closed when `slashCommandGroups` is supplied and the channel group is not eligible.
- `/resume` and `/sessions` now declared in `config-examples/slack-app-manifest.json`, matching the docs claim and `handleSlashSystemCommand`.

CI: format, typecheck, agent-runner-typecheck, test, validate-skill-only, CodeRabbit all green. `tsc --noEmit` clean; 83 Slack/flow tests pass locally. Merging under the mergemaster window (formal APPROVED unavailable due to shared author identity).

PR #922, opened 2026-06-26
fix(slack): drop reactions on non-agent messages

Peyton-Spencer:
## Summary
Slack's `reaction_added` event fires for every reaction in any registered channel — the handler only dropped reactions added *by* the bot itself, so a `:fire:` from a teammate on a *human* message still piped into `onReaction`, generated a synthetic `@Clayton [User reacted with :fire:]` trigger, and woke the warm container for a $0.20–$2.40 no-op reply ("Not addressed to me; staying silent.").

Adds the same author-is-this-bot check the Discord adapter already does (`reaction.message.author?.id !== this.client.user?.id`) so we only forward reactions on the bot's *own* messages.

Refactors the inline event handler into a dedicated `handleReactionAdded(event)` method so it's directly unit-testable (the bolt `app.event(...)` registration is opaque to the test harness).

## Background
Caught from logs of two back-to-back wakeups on the omniclaw channel (Clayton's container serves both Slack and Discord):
```
06:18:11 Clayton  Reaction on bot message in Slack
06:18:20 Result … cost=$0.2366 text=<internal>Reaction-only wakeup … Not addressed to me, no request to action.
06:18:50 Result … cost=$2.3929 text=<internal>Wakeup from Nicholas Allen's :fire: reaction on Peyton's "Sign in with Ditto" test screenshot. Not addressed to me; no action needed.
```
Reported by Peyton in #omniclaw.

## Changes
- `src/channels/slack.ts`
  - Add `if (this.botUserId && event.item_user !== this.botUserId) return;` — only fire on this bot's own messages.
  - Extract reaction logic into `handleReactionAdded(event: ReactionAddedEvent)`; bolt handler now delegates with `.catch(logger.error)`.
  - Import `ReactionAddedEvent` from `@slack/types`.
- `src/channels/slack.test.ts`
  - 4 new tests for the reaction handler: fires on own messages, drops teammate-bot author, drops human author, drops self-added reactions.

## Test plan
- [x] `bun test src/channels/slack.test.ts` — 67 pass, 0 fail
- [x] `bun test` (full suite) — 2483 pass, 0 fail
- [ ] Deploy and observe: reactions from humans on human messages should produce no container wakeups in #omniclaw.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reaction handling in Slack channels so reactions are only processed for messages tied to the current bot.
  * Reactions are now ignored when added by the bot itself, by other bots, or on human-authored messages.
  * Reaction notifications now include the correct message, channel, emoji, and reacting user details when delivered.

### DIFF
--- a/src/channels/slack.ts
+++ b/src/channels/slack.ts
@@ -3,6 +3,7 @@ import path from 'path';
 import { App, Assistant } from '@slack/bolt';
 import { WebClient } from '@slack/web-api';
 import type { AnyChunk, KnownBlock } from '@slack/web-api';
+import type { ReactionAddedEvent } from '@slack/types';
 
 import { logger } from '../logger.js';
 import {
@@ -262,35 +263,9 @@ export class SlackChannel implements Channel {
 
     // Register reaction handler (for share-request approvals etc.)
     this.app.event('reaction_added', async ({ event }) => {
-      const channelId =
-        event.item.type === 'message' ? event.item.channel : null;
-      if (!channelId) return;
-
-      // Ignore reactions from this bot itself — Slack delivers reaction_added
-      // events for our own reactions when we use `reactions.add` from
-      // share-request approval flows, and they would otherwise re-enter
-      // the message loop as @-mention noise.
-      if (this.botUserId && event.user === this.botUserId) return;
-
-      const chatJid = channelIdToJid(
-        channelId,
-        this.multiBotMode ? this.botId : undefined,
+      await this.handleReactionAdded(event).catch((err: unknown) =>
+        logger.error({ err }, 'Error handling Slack reaction_added'),
       );
-      const messageId = event.item.type === 'message' ? event.item.ts : null;
-      if (!messageId) return;
-
-      const emoji = `:${event.reaction}:`;
-
-      const { name: userName, isBot } = await resolveSlackUserIdentity(
-        this.client,
-        event.user,
-        event.user,
-      );
-
-      this.opts.onReaction?.(chatJid, messageId, emoji, userName, {
-        id: event.user,
-        isBot,
-      });
     });
 
     // Slack AI-app assistant threads (the "agent" split-pane experience).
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Reviewed the Slack reaction filter change. I do not see a blocking issue in this PR.

What I checked:

- `handleReactionAdded()` now drops reactions added by this bot and reactions whose `item_user` is not this bot, matching the Discord adapter's "only reactions on agent messages" behavior.
- The forwarded path still preserves the expected Slack JID scoping, message timestamp, emoji formatting, and reactor identity.
- The new tests cover own-message delivery plus human-message, teammate-bot-message, and self-added-reaction drops.

Validation:

```bash
bun test src/channels/slack.test.ts
```

Result: 67 pass, 0 fail.

GitHub checks visible to `gh pr checks` only show CodeRabbit, which is passing. I cannot submit a formal approval from this shared token because it is the PR author identity, but this is a technical sign-off from my review.

src/channels/slack.ts
```diff
@@ -879,6 +854,46 @@ export class SlackChannel implements Channel {
...
+   *      human/teammate messages would wake the warm container with a
+   *      synthetic trigger that costs an inference round-trip just to no-op.
+   *      Mirrors the Discord adapter's `reaction.message.author?.id` check.
+   */
+  async handleReactionAdded(event: ReactionAddedEvent): Promise<void> {
+    const channelId = event.item.type === 'message' ? event.item.channel : null;
+    if (!channelId) return;
+
+    if (this.botUserId && event.user === this.botUserId) return;
+    if (this.botUserId && event.item_user !== this.botUserId) return;
```
Peyton-Spencer: This should fail closed when `botUserId` is unavailable. As written, both guards are conditional on `this.botUserId`, so if `auth.test()` fails during `connect()` or a reaction arrives before the ID is populated, the handler cannot prove the reacted-to message belongs to this bot and still dispatches `onReaction`. That preserves the noisy wakeup path this PR is meant to remove.

Please drop reactions unless ownership is confirmed, e.g. return when `!this.botUserId || event.item_user !== this.botUserId`, and add a regression test where `botUserId` is `null`/unset and `onReaction` is not called.

Peyton-Spencer (commented):
Follow-up after verifying the current diff and CodeRabbit's line finding: I found one valid blocker in the reaction ownership guard.

The main filter currently fails open when `botUserId` is unset, which can happen if `auth.test()` fails during Slack startup. In that state OmniClaw cannot confirm the reacted-to message is the agent's message, but still dispatches `onReaction` and can wake the container.

I left an inline comment on `src/channels/slack.ts`. The fix should be small: fail closed when `botUserId` is unavailable and add a regression test for the unset-ID path.

Peyton-Spencer (commented):

Verified the review blocker against current head `9d091db`:
- The reaction ownership guards no longer fail open. `handleReactionAdded()` now returns early when `this.botUserId` is unset (`auth.test()` failure or pre-populate race), so a reaction can never wake the container unless bot ownership is confirmed (`event.item_user === botUserId` and `event.user !== botUserId`). Regression test added for the `botUserId === null` path.

CI: format, typecheck, agent-runner-typecheck, test, validate-skill-only, CodeRabbit all green. `tsc --noEmit` clean; 68 Slack tests pass locally. Clean local merge against main (post-#921). Merging under the mergemaster window (formal APPROVED unavailable due to shared author identity).

PR #926, opened 2026-06-26
feat(slack): stay proactive in threads after a bot mention

Peyton-Spencer:
## Summary

When a Slack OmniClaw agent is working in a thread it was **@-mentioned** in, it now stays **proactive** — responding to subsequent plain replies in that thread without needing a re-mention every time. This makes the in-thread experience conversational instead of stop-and-start.

Requested by Peyton: _"when a slack omniclaw agent is working in a thread that started with their mention, it should default to proactive instead of only listening to mentions."_

## How it works

This mirrors the existing **Discord** adapter pattern (`isThread && channel.ownerId === botId` → auto-prepend trigger). Slack threads are identified by `thread_ts`, so:

1. Any inbound message that @-mentions this bot marks its **thread root** (`channelId:thread_ts`) as *engaged* in a bounded `Set` (max 1000, LRU-style eviction).
2. A **plain reply** (no mention) in an engaged thread gets the group trigger prepended to its content, so `selectSubscriptionsForMessage()` wakes the agent exactly as it would for an explicit mention.
3. The trigger is prepended at the **start** of content, matching both the anchored `^@Name\b` trigger pattern and the non-anchored `getMentionPatterns()` filter downstream.

### Scope / safety
- **Channel threads only** — assistant (AI-app pane) threads are skipped (`!opts.assistantThread`); they're already 1:1 and always proactive.
- **Per-bot isolation** — each `SlackChannel` instance tracks its own engaged threads, so in multi-bot channels only the mentioned bot goes proactive. Other bots still require an explicit trigger.
- Mention detection is robust: matches by `botUserId`, by resolved mention display-name vs. trigger handle, or by anchored trigger pattern.

## Tests
- `stays proactive in a thread after the bot is mentioned at the root`
- `does not auto-trigger plain replies in threads the bot was not mentioned in`
- `engages proactive mode when the bot is mentioned mid-thread` (no double-prepend on the engaging message)

All 66 Slack tests pass; full suite **2482 pass / 0 fail**. `tsc --noEmit` clean, prettier formatted.

## Summary by CodeRabbit

* **New Features**
  * Slack thread replies can now enter a proactive mode after an initial bot mention or trigger, allowing follow-up replies in the same thread to be automatically picked up.
  * Plain replies in an engaged thread are now auto-prefixed to keep the conversation moving without requiring repeated mentions.

* **Tests**
  * Added coverage for proactive thread behavior, including root mentions, plain replies, and mid-thread activation cases.

### DIFF
--- a/src/channels/slack.ts
+++ b/src/channels/slack.ts
@@ -4,6 +4,7 @@ import { App, Assistant } from '@slack/bolt';
 import { WebClient } from '@slack/web-api';
 import type { AnyChunk, KnownBlock } from '@slack/web-api';
 
+import { buildTriggerPattern } from '../config.js';
 import { logger } from '../logger.js';
 import {
   MAX_BINARY_DOWNLOAD_BYTES,
@@ -51,6 +52,7 @@ const TEXT_LIMIT = 4000;
 const MARKDOWN_BLOCK_LIMIT = 11500;
 const THREAD_ROOT_CACHE_MAX = 1000;
 const SEEN_MESSAGE_CACHE_MAX = 500;
+const PROACTIVE_THREAD_CACHE_MAX = 1000;
 const THREAD_EXCERPT_MAX_CHARS = 140;
 const TASK_TITLE_MAX_CHARS = 80;
 // Slack caps each task_update chunk at 256 chars; keep title + details under it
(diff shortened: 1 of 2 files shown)


src/channels/slack.ts
```diff
@@ -1018,6 +1039,46 @@ export class SlackChannel implements Channel {
...
+        ? group.trigger.replace(/^@/, '').toLowerCase()
+        : null;
+      const triggerPattern = group.trigger
+        ? buildTriggerPattern(group.trigger)
+        : null;
+      const mentionsThisBot =
+        (!!this.botUserId && mentions.some((m) => m.id === this.botUserId)) ||
+        (!!triggerHandle &&
+          mentions.some((m) => m.name.toLowerCase() === triggerHandle)) ||
+        (!!triggerPattern && triggerPattern.test(content));
```
Peyton-Spencer, line 1060: This should test `resolvedText` rather than `content`. At this point threaded replies have already been prefixed with `[Thread reply ...]`, while `buildTriggerPattern()` is anchored with `^@...`; a user who replies `@bot can you help` in an existing Slack thread will not match here, so the thread never enters proactive mode unless the message contains a real Slack `<@UBOT>` mention. The fix should keep mention-name checks as-is but run the regex against the unannotated message text.

Peyton-Spencer (commented):
I found one routing bug. I verified the issue against current code and ran `bun test src/channels/slack.test.ts` locally: 66 pass.

PR #929, opened 2026-06-26
fix(web): block SolidStart page CSRF before proxy

Peyton-Spencer:
## Summary

- Move the Web UI cross-site unsafe-method guard before SolidStart page proxying.
- Add a regression test proving cross-site Basic Auth POSTs to SolidStart page routes are rejected before the internal proxy fetch runs.

## Security Impact

`WEB_UI_SOLID=true` previously delegated non-API page requests to SolidStart before the shared CSRF/origin guard. If the SolidStart app adds page actions or other non-API mutations, Basic Auth deployments could bypass the existing CSRF protection for those routes.

## Verification

- `bunx prettier --check src/web/server.ts src/web/server.test.ts`
- `bun run tsc --noEmit`

## Reviewer Request

Please leave review feedback directly on this GitHub PR, using file/line comments where applicable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cross-site browser POST requests are now blocked earlier and more reliably, including in SolidStart-backed flows.
  * Requests that are rejected now return a clear 403 response with a “Cross-site request blocked” message.
  * Improved test isolation to prevent one request or environment setting from affecting the next.

### DIFF
--- a/src/web/server.test.ts
+++ b/src/web/server.test.ts
@@ -1,8 +1,6 @@
 import { createHash, createHmac, randomUUID } from 'crypto';
 
 import { describe, it, expect, afterEach, mock } from 'bun:test';
-import fs from 'fs';
-import path from 'path';
 
 import {
   isTrustedLanDiscoveryAdminRequest,
@@ -210,8 +208,6 @@ const authHeader = `Basic ${btoa(`${testAuth.username}:${testAuth.password}`)}`;
 const peerSecret = 'peer-shared-secret-32-bytes-long!';
 const originalFetch = globalThis.fetch;
 const originalWebUiSolid = process.env.WEB_UI_SOLID;
-const solidOutputDir = path.join(process.cwd(), 'webui', '.output', 'server');
-const solidOutputFile = path.join(solidOutputDir, 'index.mjs');
 
 let handle: WebServerHandle | null = null;
 
@@ -231,10 +227,6 @@ afterEach(async () => {
   } else {
     process.env.WEB_UI_SOLID = originalWebUiSolid;
   }
-  fs.rmSync(path.join(process.cwd(), 'webui', '.output'), {
-    recursive: true,
-    force: true,
-  });
   resetDiscoveryContextForTests();
 });
 
@@ -492,11 +484,6 @@ describe('basic auth', () => {
 
   it('rejects cross-site SolidStart page POSTs before proxying', async () => {
     process.env.WEB_UI_SOLID = 'true';
-    fs.mkdirSync(solidOutputDir, { recursive: true });
-    fs.writeFileSync(
-      solidOutputFile,
-      'globalThis.__solidServerLoaded = true;\n',
-    );
     const proxyFetch = mock(
       async () => new Response('proxied unsafe page mutation', { status: 299 }),
     );
--- a/src/web/server.test.ts
+++ b/src/web/server.test.ts
@@ -1,6 +1,8 @@
 import { createHash, createHmac, randomUUID } from 'crypto';
 
-import { describe, it, expect, afterEach } from 'bun:test';
+import { describe, it, expect, afterEach, mock } from 'bun:test';
+import fs from 'fs';
+import path from 'path';
 
 import {
   isTrustedLanDiscoveryAdminRequest,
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Reviewed the CSRF guard ordering change.

No blockers found. Moving `isCrossSiteBrowserMutation()` ahead of the SolidStart delegation closes the page-proxy gap while still preserving the important earlier decisions: authenticated peer routes bypass the browser-CSRF guard, login/logout stay handled before the guard, and safe methods / CORS preflight are not classified as unsafe mutations.

Verification:
- `bun test ./src/web/server.test.ts` passed: 78 tests, 574 assertions
- `bun test ./src/web/solid-handler.test.ts` passed: 3 tests, 31 assertions
- `git diff --check origin/main...HEAD` passed

GitHub CI is green for format, typecheck, agent-runner-typecheck, and test. CodeRabbit was still pending when I started the review.

Peyton-Spencer (commented):
Independent review of the CSRF-guard reordering — verified correctness by tracing the full `fetchHandler` ordering on the PR head, not just the diff.

**Correctness: confirmed.** Moving `isCrossSiteBrowserMutation()` from line ~844 (after SolidStart delegation) to line ~445 (right after the auth gates) closes the real gap: a cross-site unsafe-method request to a SolidStart page route was previously proxied to the SolidStart handler (`if (solidMode && !isMainProcessRoute(...))`, ~line 537) *before* any CSRF check, so a forged POST could reach a SolidStart server action. The guard now rejects it first.

**No legitimate flow regresses.** The guard only triggers on unsafe methods from cross-site origins and excludes `isAuthenticatedPeerRequest`, so same-origin browser mutations, cross-site GET reads, and peer requests all still pass. Placing it *after* the session-validation gate (lines 397–410) is the right order: an unauthenticated cross-site request still gets the 302/401 from the session gate rather than a 403, while an authenticated-but-cross-site mutation now correctly gets the 403.

**Test coverage is on-point.** The new test drives the exact gap — valid Basic credentials + `Sec-Fetch-Site: cross-site` POST to a SolidStart page route, asserting 403 *and* that `proxyFetch` was never called. The afterEach restore of `fetch`/`WEB_UI_SOLID` and the `.output` cleanup are tidy.

**Non-blocking observation (out of scope):** `/login` (POST) and `/logout` sit ahead of the guard. That's unchanged by this PR — the guard was already downstream of them — and it's standard low-severity login/logout CSRF (login requires the real password; logout is benign). Worth a follow-up note only if you ever want to harden those, not something to address here.

LGTM. Approving.

src/web/server.test.ts
```diff
@@ -219,6 +225,16 @@ afterEach(async () => {
...
+  globalThis.fetch = originalFetch;
+  if (originalWebUiSolid === undefined) {
+    delete process.env.WEB_UI_SOLID;
+  } else {
+    process.env.WEB_UI_SOLID = originalWebUiSolid;
+  }
+  fs.rmSync(path.join(process.cwd(), 'webui', '.output'), {
+    recursive: true,
+    force: true,
+  });
```
Peyton-Spencer: This `afterEach` deletes the entire `webui/.output` tree, but the test only creates `webui/.output/server/index.mjs`. `webui/.output/` is gitignored — it's the real SolidStart build artifact produced by building the web UI. A developer who has run a local build will have their whole `.output/` wiped just by running this test file.

Same class of issue as the `data/opencode-data` cleanup Cody flagged on #935: the teardown is broader than the fixture it created. Suggest the pattern Peyton applied there — record whether `webui/.output` existed before the test and only `rmSync` the part the test actually created (e.g. remove `solidOutputFile` / the `server` subdir, and only remove `.output` itself if it didn't pre-exist). Non-blocking; CI is unaffected since `.output` is never present there.

Peyton-Spencer (commented):
The security fix itself is correct and well-targeted. Moving the `isCrossSiteBrowserMutation` guard up to line ~445 (before the SolidStart page-proxy delegation at `if (solidMode && !isMainProcessRoute(...))`) closes a real gap: with `WEB_UI_SOLID=true`, a cross-site Basic-Auth POST to a non-`/api/` page route previously hit the proxy `fetch` and returned long before reaching the old guard position at the end of the handler. Peer requests stay exempt via `!isAuthenticatedPeerRequest`, and `OPTIONS`/GET aren't unsafe methods so the reordering doesn't disturb preflight or SSE routes. The regression test asserts the proxy `fetch` is never called, which is exactly the invariant that matters.

One non-blocking cleanup nit, same class as the `data/opencode-data` concern Cody raised on #935 — flagged inline.

LGTM once that's considered.

Peyton-Spencer:
Addressed Clayton's teardown-scope nit in eb8fc13.

The SolidStart fixture writes were unnecessary for this regression: the invariant is that the cross-site page POST is rejected before SolidStart proxy delegation, so the test can assert the mocked proxy `fetch` was not called without creating `webui/.output/server/index.mjs` at all.

Change made:
- removed the `webui/.output` fixture creation
- removed the broad `fs.rmSync(webui/.output, { recursive: true })` cleanup
- kept the `WEB_UI_SOLID` env and `globalThis.fetch` restoration

Local verification:
- `bun run typecheck`
- `bun run format:check`

CI has restarted on the pushed commit and is pending at comment time.

PR #931, opened 2026-06-27
feat(web): deep-link /ipc last-error cells to filtered logs

Peyton-Spencer:
## What

Deep-links the `/ipc` inspector's **last error** cells to the **logs** page, pre-filtered to the failing group.

The IPC inspector already surfaces each group's last error inline with a link to the logs page (per #644: *"surfaces retry count and last error message inline, truncated, with link to logs"*). But that link pointed at a bare `/logs` — dropping the operator into an unfiltered firehose. This wires the error cell straight to that group's lines.

## Changes

- **`ipc-inspector.ts`** — the last-error link href is now `/logs?q=<encoded folderKey>` (new `logsDeepLink` helper; group key is `encodeURIComponent`-encoded then HTML-escaped for the attribute).
- **`page-scripts.ts` (logs script)** — on init, read a `?q=` query param, seed the search box, and run a filter pass so pre-existing (synced) lines are filtered immediately rather than only when the next SSE line lands.
- **`page-scripts.ts` (ipc live poll)** — the client-rendered queue row uses the same deep-link, so it survives the 5s poll refresh.

The logs page filters by full line text (which includes the group/container context), so `?q=<group>` reliably narrows to that group's lines. Plain `<a href>` navigation is a full page load, so the query string survives to the logs init script.

## Scope / non-goals

- No new routes, no API changes, no `WebStateProvider` changes — extends existing surfaces only, per #644's guidance.
- `?q=` only (search seed). Level/source pre-seeding left out to keep this focused.

## Tests

- `ipc-inspector.test.ts` — asserts the deep-link href on the last-error cell, plus a new case verifying URL-encoding of group keys containing spaces/slashes.
- `page-scripts.test.ts` — asserts the logs script reads `?q=`, seeds `searchInput`/`searchTerm`, and triggers `applyFilters()` within the deep-link block.

Full suite green: **2506 pass / 0 fail**, `tsc --noEmit` clean, Prettier clean.

## Related

Incremental step toward #644 (operator observability rollup — *"better surfacing of active runs, queue state, and reason codes; lightweight diagnostics"*). Small, self-contained, no dependency on the data-layer work.

## Summary by CodeRabbit

* **New Features**
  * Logs links now open with a pre-filled search, taking you directly to matching results.
  * IPC “last error” entries now deep-link to the relevant logs view with the group already filtered.

* **Bug Fixes**
  * Improved handling of group names in log links, including names with spaces or slashes.
  * Links now preserve the search query when navigating to logs, making error investigation faster.


Peyton-Spencer (commented):
No blocking findings. The static and live-poll href paths preserve the query string via normal navigation, encode the folder key before HTML escaping, and the logs init seeds searchTerm then applies filters immediately. Targeted tests pass locally: bun test src/web/ipc-inspector.test.ts src/web/page-scripts.test.ts.

Peyton-Spencer (commented):
Holding merge for one follow-up: the CodeRabbit note looks valid from the PR description/diff surface. The SSR IPC row handles the last-error cell from both message/task lanes, but the client poll renderer appears to only use `messageLane.lastError`, so task-lane errors can disappear on refresh.

Next step: update the live poll `leCell` path to choose `messageLane.lastError ?? taskLane.lastError` (or equivalent matching SSR behavior), preserving the new `/logs?q=` deep-link formatting, then add/adjust the focused page script test.

PR #935, opened 2026-06-27
test(backends): cover shared VM cleanup paths

Peyton-Spencer:
## Summary

- Add deterministic `SharedVmManager` tests for optional OpenCode data mounts.
- Add monkey-patched orphan cleanup tests that avoid invoking Docker.
- Add fallback coverage for shell cleanup failures and stop failures.

## Test Count

- Before: 2505 tests, 0 failures on fresh `origin/main`.
- After: 2509 tests, 0 failures.

## Coverage

- Overall line coverage: 92.27% -> 92.54%.
- `src/backends/shared-vm.ts` line coverage: 66.67% -> 91.04%.
- `src/backends/shared-vm.ts` function coverage: 72.73% -> 86.67%.

## Files Covered

- `src/backends/shared-vm.ts`

## Remaining Gaps

- `src/backends/local-backend.ts` remains low coverage because much of it shells out to container/runtime operations.
- Channel adapters (`discord.ts`, `telegram.ts`, `whatsapp.ts`) still have large integration-heavy regions uncovered.
- `src/index.ts` still has broad orchestrator paths that would benefit from additional simulation harness coverage.

## Verification

- `bun test src/backends/shared-vm.test.ts`
- `bun test`
- `bun test --coverage`

## Summary by CodeRabbit

* **Tests**
  * Expanded cleanup coverage for shared VM behavior.
  * Added checks for orphan VM handling, failure tolerance, and data mount setup when available.
  * Improved test cleanup to remove leftover app data between runs.

### DIFF
--- a/src/backends/shared-vm.test.ts
+++ b/src/backends/shared-vm.test.ts
@@ -176,31 +176,29 @@ describe('SharedVmManager', () => {
   it('stops only running orphaned shared VMs', async () => {
     const manager = new SharedVmManager();
     (manager as any).containerName = 'omniclaw-shared-claude-current';
-    const shellSpy = spyOn(Bun, '$').mockImplementation(
-      (() => ({
-        quiet: () => ({
-          text: () =>
-            JSON.stringify([
-              {
-                status: 'running',
-                configuration: { id: 'omniclaw-shared-claude-old' },
-              },
-              {
-                status: 'stopped',
-                configuration: { id: 'omniclaw-shared-claude-stopped' },
-              },
-              {
-                status: 'running',
-                configuration: { id: 'omniclaw-shared-claude-current' },
-              },
-              {
-                status: 'running',
-                configuration: { id: 'unrelated-container' },
-              },
-            ]),
-        }),
-      })) as unknown as typeof Bun.$,
-    );
+    const shellSpy = spyOn(Bun, '$').mockImplementation((() => ({
+      quiet: () => ({
+        text: () =>
+          JSON.stringify([
+            {
+              status: 'running',
+              configuration: { id: 'omniclaw-shared-claude-old' },
+            },
+            {
+              status: 'stopped',
+              configuration: { id: 'omniclaw-shared-claude-stopped' },
+            },
+            {
+              status: 'running',
+              configuration: { id: 'omniclaw-shared-claude-current' },
+            },
+            {
+              status: 'running',
+              configuration: { id: 'unrelated-container' },
+            },
+          ]),
+      }),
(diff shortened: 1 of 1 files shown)


src/backends/shared-vm.test.ts
```diff
@@ -52,6 +52,10 @@ describe('SharedVmManager', () => {
       process.env.OMNICLAW_EXTRA_DIR = originalExtraDir;
     }
     fs.rmSync(extraDir, { recursive: true, force: true });
+    fs.rmSync(path.join(DATA_DIR, 'opencode-data'), {
```
Peyton-Spencer: This cleanup removes the entire `DATA_DIR/opencode-data` tree. `DATA_DIR` resolves to the checkout's normal `data/` directory, and `data/` is ignored, so a developer or local runner can have real OpenCode runtime state under this path. Running this test would delete all of it, not just the fixture created by the new mount test. Please make the fixture test-owned, e.g. create/remove a unique child under `opencode-data` or snapshot whether the top-level directory existed and only remove the exact files this test created.

Peyton-Spencer (commented):
Thanks for adding coverage here. I found one fixture-isolation issue.

Verification: `bun test src/backends/shared-vm.test.ts` passes locally.

src/backends/shared-vm.test.ts
Peyton-Spencer: Fixed in 78475fc by removing the suite-wide `DATA_DIR/opencode-data` deletion. The OpenCode mount test now records whether the directory existed before the test and only removes the empty directory when the test created it. Verified with `bun test src/backends/shared-vm.test.ts`.

Peyton-Spencer:
Cross-ref: #945 adds overlapping coverage to the same `src/backends/shared-vm.test.ts` regions and will conflict on merge. This PR is the broader superset — recommend landing it first, then rebasing #945 down to just its two unique cases (orphan list-failure + unparseable-cache-state). Details in #945's thread.

Peyton-Spencer:
PM: superseded by the 4-way shared-VM test consolidation — full keep/close plan and coverage map on #970. This PR overlaps #970 on `cleanupOrphans` coverage and will conflict on merge. Recommend closing after its unique tests (if any) are cherry-picked into #970. Root cause: #952.

Peyton-Spencer:
PM cleanup: closing — superseded by the shared-VM test consolidation that merged as #970, and this PR is CONFLICTING against main (cannot merge as-is). Coverage already landed. Reopen if any unique case was missed. (Resolves the #952 shared-vm.test.ts collision.)

PR #938, opened 2026-06-28
fix(web): preserve dashboard agents working-count on live stats patch

Peyton-Spencer:
## Problem

The dashboard **agents** stat card renders a `(N working)` annotation on the static page (shipped in #884), telling the operator how much of the fleet is actively processing a message or running a task. But the live SSE stats patch (`patchStats` in `server.ts`) re-rendered the card with a bare count:

```ts
`<div class="value" id="stat-agents">${Object.keys(state.getAgents()).length}</div>`
```

So on the **first** `agent_status` / `task_update` event, the `(N working)` annotation silently disappeared — the static render and the live update disagreed, and the #884 feature regressed the moment anything happened on the dashboard.

## Fix

- Extract two shared helpers into `task-stats.ts`:
  - `countWorkingAgents(queueDetails)` — agents in flight on either lane (message lane active & not idle, or task lane active).
  - `formatAgentsValue(total, working)` — the `N` / `N (M working)` formatting.
- Use them from **both** the static dashboard render (`dashboard.ts`) and the live patch (`server.ts`) so the two can never diverge again.

No behavior change to the static render (logic is identical, now deduplicated). The live path now matches it.

## Tests

- `task-stats.test.ts` (new) — direct unit coverage for `countWorkingAgents` (idle-waiting lane excluded, both-lanes-counted-once, multi-group sum) and `formatAgentsValue`.
- `server.test.ts` — two SSE live-patch tests: working annotation **preserved** on `agent_status`, and **omitted** when no lane is active.

```
2516 pass / 0 fail   (bun test)
tsc --noEmit          clean
prettier --check      clean
```

## Summary by CodeRabbit

* **New Features**
  * Dashboard and live stats now show agent counts with clearer working-agent status, such as `total (N working)` when applicable.

* **Bug Fixes**
  * Fixed inconsistent agent statistics between the dashboard and real-time updates.
  * Improved handling for cases where no queue details are available, so the agents count displays cleanly without extra wording.

* **Tests**
  * Added coverage for live stats updates and agent count formatting across active, inactive, and zero-agent scenarios.

### DIFF
--- a/src/web/dashboard.ts
+++ b/src/web/dashboard.ts
@@ -5,7 +5,11 @@ import { renderShell, escapeHtml, escapeJsonForHtml } from './shared.js';
 import { allPageScripts } from './page-scripts.js';
 import { buildAgentChannelData } from './agent-channels.js';
 import type { RemotePeerAgents } from '../discovery/types.js';
-import { formatActiveTaskStatsFromState } from './task-stats.js';
+import {
+  formatActiveTaskStatsFromState,
+  countWorkingAgents,
+  formatAgentsValue,
+} from './task-stats.js';
 
 function imageRev(value: string): string {
   return createHash('sha256').update(value).digest('hex').slice(0, 12);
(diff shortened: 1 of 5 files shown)


Peyton-Spencer (commented):
Reviewed the helper extraction and the SSE/static dashboard behavior. No blocking findings.

The shared `countWorkingAgents`/`formatAgentsValue` path fixes the live patch regression without expanding scope, and the added unit + SSE coverage exercises the important active/idle lane cases.

Non-blocking: the live SSE stat still uses local `state.getAgents()` while static render can include remote peers. That is pre-existing/architectural because SSE clients do not currently carry a resolved remote-peer snapshot, so I would handle it as a separate follow-up rather than block this fix.

Local verification:
- `bun test src/web/task-stats.test.ts src/web/server.test.ts src/web/dashboard.test.ts`
- `bun run typecheck`
- `bun run format:check`

GitHub checks are green except CodeRabbit still pending at review time.

PR #940, opened 2026-06-28
fix(web): sanitize SolidStart proxy headers

### LINKED ISSUE #936: security: SolidStart proxy forwards credentials and hop-by-hop headers
## Summary

When `WEB_UI_SOLID=true`, `src/web/solid-handler.ts` proxies authenticated, non-API page requests to the internal SolidStart server by cloning all inbound request headers:

```ts
const headers = new Headers(req.headers);
headers.set('host', `127.0.0.1:${internalPort}`);
headers.set('accept-encoding', 'identity');
```

That forwards hop-by-hop/proxy-sensitive headers and dashboard credentials (`Authorization`, `Cookie`) to the internal loopback page server even though the main Bun server has already enforced authentication and CSRF policy.

## Severity

P3 / Low-Medium security hardening.

This requires access to the Web UI endpoint, so it is not an unauthenticated remote takeover. The risk is still worth fixing because the SolidStart server is a separate HTTP trust boundary and should not receive client connection metadata or reusable dashboard credentials by default.

## Impact

- Hop-by-hop headers such as `Connection`, `Keep-Alive`, `TE`, `Trailer`, `Transfer-Encoding`, and `Upgrade` can be forwarded across a proxy boundary where they have different semantics and should be consumed by the proxy.
- `Authorization` and `Cookie` values are exposed to the internal SolidStart page server even though auth is handled in `src/web/server.ts` before delegation.
- If a future SolidStart route logs headers, performs SSR fetches with inherited headers, or adds middleware that treats forwarded headers as authoritative, dashboard credentials or connection-control data may leak or affect internal routing behavior.

## Reproduction

1. Enable SolidStart mode with `WEB_UI_SOLID=true`.
2. Send an authenticated request to any non-main-process page route with extra headers, for example `/settings`.
3. Observe `initSolidHandler()` forwarding a cloned `Headers(req.headers)` object to `fetch(targetUrl, ...)` after only overriding `Host` and `Accept-Encoding`.
4. In a test or instrumented internal SolidStart handler, the loopback request receives the original `Authorization`/`Cookie` headers and any caller-supplied hop-by-hop headers that the runtime allows through.

## Suggested Fix

Build a fresh outbound header allowlist for the internal SolidStart request instead of cloning all inbound headers. Preserve only headers needed by the page app, for example `Accept`, `Accept-Language`, `Content-Type`, and safe Datastar/SolidStart-specific request headers if any are required.

Also explicitly drop hop-by-hop and credential headers:

- `authorization`
- `cookie`
- `connection`
- `keep-alive`
- `proxy-authenticate`
- `proxy-authorization`
- `te`
- `trailer`
- `transfer-encoding`
- `upgrade`

Add a regression test in `src/web/solid-handler.test.ts` proving credentials and hop-by-hop headers are not forwarded while ordinary safe headers still are.

## Areas Reviewed This Run

- `src/web/solid-handler.ts`: SolidStart loopback proxy header forwarding, body forwarding, and internal target construction.
- `src/web/server.ts`: auth ordering, CSRF guard placement, SolidStart delegation, SSE routes, and CORS ordering.
- `src/web/page-scripts.ts` and `src/web/shared.ts`: remaining `innerHTML` paths and Markdown sanitizer usage.
- `src/task-preprocessor.ts`: workflow path validation, env allowlist, subprocess invocation, output caps.
- `src/backends/local-backend.ts`: subprocess argument arrays, IPC path validation, backend read/write guards.
- `src/media.ts` and `src/web/image-cache.ts`: attachment and remote image SSRF validation, DNS pinning, byte caps, and content-type restrictions.

Peyton-Spencer:
## Summary

- Build a fresh allowlisted header set for the internal SolidStart proxy instead of cloning inbound request headers.
- Preserve only page-rendering headers needed by the loopback app: `Accept`, `Accept-Language`, and `Content-Type`.
- Add regression coverage that `Authorization`, `Cookie`, hop-by-hop headers, and arbitrary caller headers are not forwarded.

Fixes #936.

## Testing

- `bun test src/web/solid-handler.test.ts`
- `bunx prettier --check src/web/solid-handler.ts src/web/solid-handler.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/web/solid-handler.test.ts
+++ b/src/web/solid-handler.test.ts
@@ -126,7 +126,15 @@ describe('solid-handler', () => {
       new Request('http://public.test/dashboard?tab=agents', {
         method: 'POST',
         headers: {
+          accept: 'text/html',
+          'accept-language': 'en-US',
           'accept-encoding': 'gzip',
+          authorization: 'Basic secret',
+          connection: 'keep-alive',
+          cookie: 'omniclaw_session=secret',
+          'content-type': 'text/plain',
+          te: 'trailers',
+          upgrade: 'websocket',
           'x-test': 'present',
         },
         body: 'hello world',
@@ -151,8 +159,16 @@ describe('solid-handler', () => {
 
     const headers = init.headers as Headers;
     expect(headers.get('host')).toBe(`127.0.0.1:${internalPort}`);
+    expect(headers.get('accept')).toBe('text/html');
     expect(headers.get('accept-encoding')).toBe('identity');
-    expect(headers.get('x-test')).toBe('present');
+    expect(headers.get('accept-language')).toBe('en-US');
+    expect(headers.get('content-type')).toBe('text/plain');
+    expect(headers.get('authorization')).toBeNull();
+    expect(headers.get('connection')).toBeNull();
+    expect(headers.get('cookie')).toBeNull();
+    expect(headers.get('te')).toBeNull();
+    expect(headers.get('upgrade')).toBeNull();
+    expect(headers.get('x-test')).toBeNull();
 
     const getResponse = await handler!(
       new Request('http://public.test/settings', { method: 'GET' }),
--- a/src/web/solid-handler.ts
+++ b/src/web/solid-handler.ts
@@ -21,6 +21,12 @@ type SolidHandler = (req: Request) => Promise<Response>;
 let solidHandler: SolidHandler | null = null;
 let solidPort: number | null = null;
 
+const SOLID_PROXY_REQUEST_HEADER_ALLOWLIST = new Set([
+  'accept',
+  'accept-language',
+  'content-type',
+]);
+
 /**
  * Initialize the SolidStart handler by importing the compiled output.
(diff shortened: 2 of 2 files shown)


Peyton-Spencer (commented):
Reviewed the SolidStart proxy header sanitization patch against the main web-server routing/auth flow and the SolidStart app code.

LGTM. Building a fresh allowlisted header set closes the sensitive-header forwarding issue without breaking the current SolidStart SSR assumptions: page rendering does not depend on forwarded `Cookie`/`Authorization`, and API/SSE routes stay in the main process. The regression test covers the important negatives (`Authorization`, `Cookie`, hop-by-hop headers, arbitrary caller headers) while preserving the headers the loopback app still needs.

Local verification:

- `bun test src/web/solid-handler.test.ts` -> 3 pass / 0 fail
- `bunx prettier --check src/web/solid-handler.ts src/web/solid-handler.test.ts` -> pass
- `bun run typecheck` -> pass

PR #941, opened 2026-06-28
fix(security): validate scheduled task group folders

Peyton-Spencer:
## Summary

- Reject traversal and absolute `group_folder` values when creating scheduled tasks via the Web UI API.
- Refuse already-persisted scheduled tasks with unsafe group folders before scheduler dispatch or filesystem writes.
- Add regression coverage for both the Web API validation path and scheduler defense-in-depth path.

## Security Impact

Before this change, a crafted persisted scheduled task could use a `group_folder` such as `../outside`. The scheduler created `GROUPS_DIR/<group_folder>` before confirming the task's group exists, so the value could escape the groups root and create directories outside the intended workspace boundary.

## Testing

- `bun test src/web/routes.test.ts src/task-scheduler-runner.test.ts`
- `bunx prettier --check src/task-scheduler.ts src/web/routes.ts src/web/routes.test.ts src/task-scheduler-runner.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards to reject unsafe task group folder paths containing traversal segments.
  * Task creation now returns a clear 400 error when an invalid folder path is submitted.
  * Scheduled tasks with unsafe group folders are stopped early and չեն’t proceed to execution or backend dispatch.
* **Tests**
  * Added coverage for rejected path traversal during task creation.
  * Added coverage to ensure unsafe scheduled tasks are abandoned before dispatch.

### DIFF
--- a/src/task-scheduler.ts
+++ b/src/task-scheduler.ts
@@ -35,6 +35,7 @@ import {
 } from './db.js';
 import { GroupQueue } from './group-queue.js';
 import { logger } from './logger.js';
+import { rejectTraversalSegments } from './path-security.js';
 import { ResumePositionStore } from './resume-position-store.js';
 import { writeScheduledRunHandoff } from './task-handoffs.js';
 import { runTaskPreprocessor } from './task-preprocessor.js';
@@ -208,6 +209,28 @@ async function runTask(
     runtime.logger.info(
       { taskId: task.id, agentFolder: task.group_folder },
       'Task skipped — agent is disabled',
+    );
+    return;
+  }
+
+  try {
+    rejectTraversalSegments(task.group_folder, 'scheduled task group folder');
+  } catch (err) {
+    const folderError = err instanceof Error ? err.message : String(err);
+    appendPhaseEvent('group_resolved', 'error', false, folderError);
+    runtime.logTaskRun({
+      task_id: task.id,
+      run_at: runAt,
+      duration_ms: 0,
+      status: 'error',
+      result: null,
+      error: folderError,
+      outcome_state: 'abandoned',
+      outcome_reason: folderError,
+    });
+    runtime.logger.warn(
+      { taskId: task.id, groupFolder: task.group_folder, err: folderError },
+      'Refusing scheduled task with unsafe group folder',
     );
     return;
   }
--- a/src/task-scheduler.ts
+++ b/src/task-scheduler.ts
@@ -202,17 +202,6 @@ async function runTask(
     return;
   }
 
-  // Skip runs when the owning agent is disabled via the web UI off-switch.
-  // next_run was already advanced before this call, so when the agent is
-  // re-enabled, future ticks pick the task up at its next scheduled time.
-  if (deps.isAgentEnabled && !deps.isAgentEnabled(task.group_folder)) {
-    runtime.logger.info(
-      { taskId: task.id, agentFolder: task.group_folder },
-      'Task skipped — agent is disabled',
-    );
-    return;
-  }
-
   try {
     rejectTraversalSegments(task.group_folder, 'scheduled task group folder');
   } catch (err) {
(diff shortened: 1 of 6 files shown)


src/task-scheduler.ts
```diff
@@ -212,6 +213,28 @@ async function runTask(
     return;
   }
 
+  try {
+    rejectTraversalSegments(task.group_folder, 'scheduled task group folder');
```
Peyton-Spencer, line 206: This still lets traversal segments through when they normalize back inside the root. `rejectTraversalSegments()` calls `path.normalize()` before splitting, so `agent/../outside` becomes `outside` and passes. That means the Web API will persist a task with `group_folder: "agent/../outside"`, and the scheduler will later run `fs.mkdirSync(path.join(GROUPS_DIR, task.group_folder), ...)`, creating `GROUPS_DIR/outside` before group resolution.

For this security boundary, please reject raw `..` path components instead of normalized-away ones, e.g. split the original string on path separators before/while checking, or add a test case for `agent/../outside` and make the helper fail it.

Peyton-Spencer:
Found one blocking validation gap in the traversal fix.

The current checks reject leading `../outside`, but still accept embedded traversal segments such as `agent/../outside` because `rejectTraversalSegments()` normalizes the path before looking for `..`. That leaves the Web API able to persist a traversal-bearing `group_folder`, and the scheduler can still create a sibling directory before it discovers the group is invalid.

Focused repro:

- `rejectTraversalSegments('../outside', ...)` rejects
- `rejectTraversalSegments('agent/../outside', ...)` passes
- `rejectTraversalSegments('agent/..', ...)` passes

I left the concrete finding inline. Please add coverage for embedded `..` components and update the validation/helper so raw traversal segments are rejected.

Peyton-Spencer:
Addressed Cody's blocking review finding in `96d646a`.

Changes:

- `rejectTraversalSegments()` now checks raw path components before normalization, so `agent/../outside` and `agent/..` are rejected instead of normalizing away the `..` segment.
- Added helper-level regression coverage for embedded/trailing traversal.
- Added Web API coverage for `group_folder: "agent/../outside"` during task creation.
- Updated the scheduler persisted-task defense test to use the embedded traversal bypass form.

Local verification:

- `bun test src/path-security.test.ts src/web/routes.test.ts src/task-scheduler-runner.test.ts`
- `bunx prettier --check src/path-security.ts src/path-security.test.ts src/web/routes.test.ts src/task-scheduler-runner.test.ts`
- `bun run typecheck`

Peyton-Spencer:
Addressed CodeRabbit's scheduler ordering finding in `08acf7b`.

The scheduler now validates `task.group_folder` immediately after confirming the task is still active, before calling `deps.isAgentEnabled()` or any other scheduler dependency with the folder value.

Added regression coverage asserting `isAgentEnabled` is not called for an unsafe persisted `group_folder`.

Local verification:

- `bun test src/path-security.test.ts src/web/routes.test.ts src/task-scheduler-runner.test.ts`
- `bunx prettier --check src/task-scheduler.ts src/task-scheduler-runner.test.ts src/path-security.ts src/path-security.test.ts src/web/routes.test.ts`
- `bun run typecheck`

Peyton-Spencer:
Re-reviewed the updated PR after `96d646a` and `08acf7b`.

The two prior findings are addressed:

- `rejectTraversalSegments()` now checks raw path components before normalization, so `agent/../outside` and `agent/..` are rejected instead of normalized away.
- Scheduler validation now runs immediately after the task active-status check and before `isAgentEnabled` or other scheduler dependencies receive `task.group_folder`.

I did not find remaining actionable issues in this patch.

Local verification:

- direct helper repro: `../outside`, `agent/../outside`, `agent/..`, and `/tmp/x` reject; normal relative folders pass
- `bun test src/path-security.test.ts src/web/routes.test.ts src/task-scheduler-runner.test.ts` -> 104 pass / 0 fail
- `bunx prettier --check src/task-scheduler.ts src/task-scheduler-runner.test.ts src/path-security.ts src/path-security.test.ts src/web/routes.ts src/web/routes.test.ts` -> pass
- `bun run typecheck` -> pass

Issue #157, opened 2026-03-01
Web UI for OmniClaw — full agent management, logs, and control panel


Peyton-Spencer:
## Problem

Discord works as a control interface but has real limitations:
- Streaming intermediate agent responses into channels/threads is noisy UX
- No structured view of agent state, tasks, containers, or IPC traffic
- No way to visualize agent logs in real-time with proper formatting
- Can't easily browse conversation history, task lists, or scheduled jobs
- Resuming/managing sessions requires Discord commands rather than direct interaction

This is becoming a bottleneck as agent complexity grows. A dedicated web UI would give full visibility and control.

## Proposed Solution

Build a web UI that serves as the primary management interface for OmniClaw:

### Core Features
- **Agent Dashboard** — live view of all running agents, their status (idle/executing/error), container info, and resource usage
- **Live Logs** — real-time streaming logs per agent with proper formatting, filtering, and search. Replaces the need to stream intermediate responses into Discord
- **Task Manager** — view, create, edit, pause, resume, and cancel scheduled tasks with full cron/interval editing
- **Conversation Viewer** — browse and search agent conversation history across all channels
- **IPC Inspector** — visualize inter-agent messages, file transfers, and context sharing in real-time
- **Configuration** — edit agent configs, CLAUDE.md files, channel registrations, and environment from the UI

### Architecture Considerations
- Could be a lightweight server running alongside the orchestrator (same process or sidecar)
- WebSocket for real-time log streaming and agent status updates
- REST API for CRUD operations on tasks, configs, agents
- Auth layer (even basic) since this exposes sensitive agent internals

### Enables
- Supersedes need for Discord thread-based agent interaction for management tasks (see #117 for the thread UX direction)
- Could eventually become an alternative input channel itself (send messages to agents from the web UI)
- Better debugging and observability than any chat platform can offer

## Why Now
- Agent count and complexity is growing
- Multi-agent coordination needs better visibility than chat logs
- The software factory vision requires proper tooling around the factory floor

## Open Questions
- Tech stack? (Bun + Hono backend is natural, frontend TBD — SolidJS for consistency with Ditto?)
- Ship as built-in to omniclaw or separate repo?
- MVP scope — what's the minimum useful first version?

Peyton-Spencer:
**Progress update (2026-03-05):**

PR #218 merged — adds conversation viewer (browse chat history in the dashboard). Web UI now has:
- `src/web/server.ts` — Bun.serve() foundation
[...]

PR #943, opened 2026-06-29
feat(web): expose preprocess script field in task create/edit forms

Peyton-Spencer:
## What

Exposes the **preprocess script** field in the web Task Manager's create and edit forms.

The task create (`POST /api/tasks`) and update (`PATCH /api/tasks/:id`) handlers have long accepted a `preprocess_script` field (validated by `normalizePreprocessScriptPath`), but the UI forms never rendered it. Until now the only way to attach a preprocessor to a scheduled task was through the raw API. This closes that gap.

A preprocess script is a relative JS/TS file in the task workflows dir that runs before each execution and returns a `run` / `skip` / `error` decision, optionally rewriting the prompt (see `src/task-preprocessor.ts`).

## Changes

- **Form field** — adds an optional *Preprocess Script* input to both the create (`tmc-preprocess`) and edit (`tme-preprocess`) task modals, with hint text describing what it does and that blank means none.
- **Client wiring** — sends `preprocess_script` (trimmed) in the create and edit payloads. An empty value clears it (the API normalizes blank → `null`). The edit modal hydrates the field from the loaded task.
- **Table indicator** — small `fn` badge in the task table's context column for tasks that have a preprocessor attached, with the path in the tooltip.
- **Styles** — new `.form-hint`, `.form-optional`, and `.badge-preprocess` rules in `shared.ts`.

## Tests

- `tasks.test.ts` — create + edit modal expose the field; table badge appears only when a preprocessor is set; tooltip path is HTML-escaped.
- `page-scripts.test.ts` — create/edit payloads include the trimmed field; edit modal hydrates it.

```
bun test        → 2520 pass, 0 fail
bun run typecheck → clean
prettier --check  → clean
```

## Notes for reviewer

Per the repo PR protocol — please leave review feedback directly in this PR as GitHub file/line comments where applicable, not only in Discord. In particular I'd value a second look at: (1) always sending `preprocess_script` on edit (so blank clears it) vs. only sending when changed, and (2) whether the `fn` table badge is the right affordance or if it adds noise.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

🤖 Part of the OmniClaw Web UI effort (#157)

## Summary by CodeRabbit

* **New Features**
  * Added an optional **Preprocess Script** field to task create/edit forms, with guidance text and validation-friendly input behavior.
  * Tasks can now display a preprocess badge when this value is set, helping users quickly identify tasks with extra preprocessing logic.

* **Bug Fixes**
  * Ensured preprocess script values are saved, loaded, and shown consistently across task creation, editing, and task lists.
  * Improved display safety by properly escaping script text in tooltips.

### DIFF
--- a/src/web/page-scripts.test.ts
+++ b/src/web/page-scripts.test.ts
@@ -101,6 +101,23 @@ describe('tasks page script', () => {
     );
     expect(script).toContain('if(isFinite(nr)&&nr<now)overdue++;');
   });
+
+  it('sends the optional preprocess script on create and edit', () => {
+    const script = allPageScripts().tasks;
+
+    // Create payload includes the trimmed preprocess field
+    expect(script).toContain(
+      'preprocess_script:document.getElementById("tmc-preprocess").value.trim()',
+    );
+    // Edit payload includes the trimmed preprocess field
+    expect(script).toContain(
+      'preprocess_script:document.getElementById("tme-preprocess").value.trim()',
+    );
+    // Edit modal hydrates the field from the loaded task (blank when unset)
+    expect(script).toContain(
+      'document.getElementById("tme-preprocess").value=t.preprocess_script||"";',
+    );
+  });
 });
 
 describe('context page script', () => {
--- a/src/web/page-scripts.ts
+++ b/src/web/page-scripts.ts
@@ -1067,7 +1067,8 @@ document.getElementById("tmc-form").addEventListener("submit",function(e){
     prompt:document.getElementById("tmc-prompt").value,
     schedule_type:document.getElementById("tmc-schedule-type").value,
     schedule_value:schedVal,
-    context_mode:document.getElementById("tmc-context-mode").value};
+    context_mode:document.getElementById("tmc-context-mode").value,
+    preprocess_script:document.getElementById("tmc-preprocess").value.trim()};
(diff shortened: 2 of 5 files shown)


Peyton-Spencer (commented):
Reviewed the UI wiring for create/edit, table rendering, API normalization behavior, and the added tests.

No blocking findings. Always sending `preprocess_script` on edit is the right behavior here because it gives the UI an explicit clear path and matches the API's empty-string-to-null normalization. The `fn` badge is useful as a compact discoverability affordance, and the tooltip is escaped.

Local verification after `bun install --force`:

```bash
bun test src/web/tasks.test.ts src/web/page-scripts.test.ts src/web/routes.test.ts
```

Result: 148 pass / 0 fail.

Peyton-Spencer (commented):
Blocked on two client-side wiring gaps that match the current diff:

- The edit save payload still omits `context_mode`, so changing the edit modal's context mode will not persist.
- The client-side `refreshTasks()` row renderer needs to include the same preprocess badge/tooltip as the SSR `renderTaskTableRows()` path; otherwise the badge can disappear after refresh/create/edit/toggle updates.

CI is green, but I would not merge until those UI paths are aligned and covered.

PR #948, opened 2026-06-30
feat(web): surface most recent lane error on /system queue rollup

Peyton-Spencer:
## What

Surface the **most recent lane error** on the `/system` queue rollup.

The queue card already rolls up retry *counts* — `retrying`, `total retries`, `max retries` — but never the actual failure text. So when an operator sees retries climbing on `/system`, they have to leave the page and drill into `/ipc` to learn *what* is going wrong. This adds the missing "what":

- New `queue.last_error` field on `HealthData`: the single newest lane error across all groups (message **or** task lane, whichever has the most recent timestamp), or `null` when nothing has failed.
- Rendered as a `last error` row on the queue card, reusing the exact `last-error-*` markup the IPC inspector uses (styled globally in `shared.ts`): a lane/folder chip, the truncated message, and the error age, linked through to `/logs`.
- Message is capped at 140 chars — the tile is a glance surface, not a log viewer; the full text stays on `/ipc` and in `/logs`.
- Live-updates over SSE via DOM-safe `textContent` construction (no `innerHTML`).

## Why

Advances the operator-observability rollup (#644): *"recent errors or stalls at a glance"* on `/system` without drilling into `/ipc`. The retry counts tell you *how much* is failing; this tells you *what* the latest failure was.

## Implementation notes

- Pure derivation from the existing `getQueueDetails()` snapshot (`GroupQueueDetail.messageLane.lastError` / `taskLane.lastError`) — no new `WebStateProvider` method, no breaking API change.
- Newest-wins tie-break by `at` timestamp; lane recorded so the chip shows `message · <folder>` or `task · <folder>`.

## Tests

`src/web/system.test.ts` (+5 cases):
- newest error wins across lanes and groups (task lane beats older message lane)
- 140-char truncation with trailing ellipsis
- `null` when no lane carries an error
- rendered row: `sys-queue-last-error` id, `/logs` link, lane·folder chip, and HTML escaping of the message (`boom <script>` -> escaped)
- em-dash render when no lane has failed

Full web suite: `bun test src/web/` -> 886 pass, 0 fail. `bun run typecheck` clean. Prettier clean on changed files.

## Scope

No new route, no schema/migration, no provider-interface change. One new row on the existing `/system` queue card; mobile layout unaffected (same `metric-row` structure as its neighbors).

## Summary by CodeRabbit

* **New Features**
  * The system health page now shows the most recent queue error, including where it happened, the error message, and how recently it occurred.
  * The error message is linked to logs for easier investigation.

* **Bug Fixes**
  * When no queue error exists, the health page now displays a clear dash instead of leaving the field blank.
  * Error text is safely rendered to avoid displaying raw markup.


Peyton-Spencer (commented):
Reviewed the derivation, render path, and SSE update path. No findings.

The newest-wins logic checks both message and task lane errors across every group and keeps the highest timestamp; the UI render escapes server-side HTML and the live update path uses DOM node construction with textContent/title assignment instead of innerHTML. The 140-char cap is reasonable for the /system glance surface and the full error remains available from the linked logs/IPC surfaces.

Verified locally:
- `bun test src/web/system.test.ts src/web/ipc-inspector.test.ts` — 92 pass / 0 fail
- `bun run typecheck` — clean
- `bun test src/web/` — 881 pass / 0 fail

PR #955, opened 2026-07-01
feat(web): annotate /ipc pending stat cards with group breadth

Peyton-Spencer:
## What

Adds a group-breadth annotation to the `pending msgs` and `pending tasks` stat cards on `/ipc`, and wires both cards into the live poll so they stay current.

Before, each card showed only the summed total (intensity):

```
pending msgs   12
```

Now it also shows how many groups actually carry that backlog (breadth):

```
pending msgs   12 (3 groups)
```

## Why

This is the breadth-vs-intensity distinction the observability reference already calls out for the `retrying` card (`N (M)`), applied to the two pending-backlog cards where it was missing. The two numbers answer a question the bare total can't:

- **`12 (1 group)`** — backlog concentrated in a single group → likely one stuck/slow agent.
- **`12 (8 groups)`** — same total spread wide → systemic back-pressure (slot/container pressure).

The format mirrors the existing `retrying` and `groups tracked` cards, so the whole `/ipc` stat row reads consistently. Falls back to a bare `0` when nothing is pending.

## Live-poll fix (bonus)

The `/ipc` poll (`fetch("/api/ipc/queue")` every 5s) updated `stat-groups` and `stat-events` but left `stat-pending-messages` / `stat-pending-tasks` frozen at their render-time values until the operator navigated away and back. Both cards now refresh on each poll with the same breadth format, computed from the queue detail the poll already fetches — no new endpoint.

## Changes

- `src/web/ipc-inspector.ts` — exported `formatPendingWithBreadth(total, groups)`; count groups-with-pending per lane; render both cards through the formatter.
- `src/web/page-scripts.ts` — compute pending totals + breadth in the `/ipc` poll and update both cards (placed before the empty-details early return so they reset to `0`).
- Tests — `formatPendingWithBreadth` unit tests (singular/plural/zero/defensive), updated `/ipc` render assertions, new spread-across-multiple-groups + all-idle cases, and a page-script assertion for the live-poll path.

## Verification

- `bun test` — 2564 pass / 0 fail
- `bun run typecheck` — clean
- `bunx prettier --check` on changed files — clean

Extends #644 (operator observability rollup). No new routes, no API changes, no breaking changes.

## Summary by CodeRabbit

* **New Features**
  * IPC status views now show pending work with both the total count and how many groups are affected, e.g. `3 (2 groups)`.
  * Pending counts now use clearer singular/plural wording and omit group details when there is no pending work.

* **Tests**
  * Added coverage for the new pending-count formatting and updated IPC UI rendering behavior.


Peyton-Spencer:
@OCPeyton could you review this one directly in the PR — file/line comments welcome, especially on the live-poll placement in `page-scripts.ts` (I put the pending-card updates before the empty-details early return so they reset to `0` correctly). CI-equivalent locally: full suite 2564 pass, typecheck + prettier clean.

Peyton-Spencer (commented):
Reviewed the diff and surrounding `/ipc` render/poll code. The formatter is small, the server-rendered aggregate math matches the per-group rows, and the poll-side update keeps the pending stat cards consistent without adding API surface.

Verified locally:

```bash
bun test src/web/ipc-inspector.test.ts src/web/page-scripts.test.ts
```

Result: 80 pass / 0 fail.

One non-blocking note: the existing `ipcScript` still returns early when the page initially rendered without `queue-body`, so an initially empty `/ipc` page will not grow a live queue table later. That looks pre-existing and outside this PR's narrow change; it does not block the pending breadth annotation for normal tracked-group states.

PR #945, opened 2026-06-29
test(backends): cover shared VM cleanup paths

Peyton-Spencer:
## Summary

- Add deterministic monkey-patch coverage for shared VM orphan cleanup.
- Cover cleanup failure logging without invoking a real container runtime.
- Cover stale cached container state when the runtime list response is unparsable.

## Test Count / Coverage

- Before: 2521 tests, 6682 assertions, 92.28% line coverage.
- After: 2524 tests, 6691 assertions, 92.64% line coverage.
- `src/backends/shared-vm.ts`: 66.67% -> 100.00% line coverage.

## Verification

- `bun test src/backends/shared-vm.test.ts`
- `bun test`
- `bun test --coverage`
- `bunx tsc --noEmit`

## Remaining Coverage Gaps

- Channel adapters still have broad low-coverage areas: Discord, Telegram, WhatsApp.
- Runtime-heavy backend code in `src/backends/local-backend.ts` remains difficult to cover without deeper process/container fakes.
- Large route/orchestrator modules (`src/discovery/routes.ts`, `src/index.ts`, `src/web/server.ts`) still have substantial uncovered branches.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for container cleanup and recovery behavior.
  * Verified that only active orphaned containers are stopped, while unrelated or already stopped containers are left alone.
  * Added coverage for graceful handling when container listing fails, including warning logging.
  * Added a scenario confirming the app restarts when cached container state can’t be read as active.

PR #965, opened 2026-07-02
test(backends): cover shared VM runtime listing

Peyton-Spencer:
## Summary

- Add deterministic SharedVmManager tests for runtime list parsing and cached-container reuse.
- Cover orphan cleanup filtering so only running stale shared VMs are stopped.
- Cover malformed runtime list output as non-fatal for cleanup and as stale-cache fallback for ensureRunning.

## Test Counts

- Before: 2568 passing tests across 106 files.
- After: 2572 passing tests across 106 files.
- Focused coverage: `src/backends/shared-vm.ts` now reports 100% line coverage under `bun test src/backends/shared-vm.test.ts --coverage`.

## Verification

- `bun test src/backends/shared-vm.test.ts`
- `bun test src/backends/shared-vm.test.ts --coverage`
- `bun test`
- `bun run typecheck`

## Remaining Coverage Gaps

- `src/backends/local-backend.ts` still has substantial low coverage around container execution and mount/runtime paths.
- Channel adapters (`discord.ts`, `telegram.ts`, `whatsapp.ts`) remain lower coverage because much of their behavior is integration-heavy.
- `src/index.ts` remains low line coverage due to orchestrator startup/runtime branches.

## Summary by CodeRabbit

* **Tests**
  * Expanded shared VM test coverage for runtime list handling, including valid, malformed, and cached container scenarios.
  * Verified cleanup behavior only stops running orphaned containers and ignores unrelated or already-stopped ones.
  * Confirmed malformed runtime output does not break cleanup.

### DIFF
--- a/src/backends/shared-vm.test.ts
+++ b/src/backends/shared-vm.test.ts
@@ -39,6 +39,19 @@ describe('SharedVmManager', () => {
   const originalExtraDir = process.env.OMNICLAW_EXTRA_DIR;
   let extraDir = '';
 
+  function mockContainerList(containers: unknown) {
+    return spyOn(Bun, '$').mockImplementation(
+      (() => ({
+        quiet: mock(async () => ({
+          text: () =>
+            typeof containers === 'string'
+              ? containers
+              : JSON.stringify(containers),
+        })),
+      })) as unknown as typeof Bun.$,
+    );
+  }
+
   beforeEach(() => {
     extraDir = fs.mkdtempSync(path.join(os.tmpdir(), 'omniclaw-extra-'));
     process.env.OMNICLAW_EXTRA_DIR = extraDir;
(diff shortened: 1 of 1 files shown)


src/backends/shared-vm.test.ts
```diff
@@ -39,6 +39,17 @@ describe('SharedVmManager', () => {
   const originalExtraDir = process.env.OMNICLAW_EXTRA_DIR;
   let extraDir = '';
 
+  function mockContainerList(containers: unknown) {
+    return spyOn(Bun, '$').mockImplementation((() => ({
```
Peyton-Spencer, line 43: This helper makes the new runtime-list tests insensitive to the actual command being issued: any `Bun.$` tagged template returns this JSON, so `isAlive`/`cleanupOrphans` could regress from `${LOCAL_RUNTIME} ls --format json` to a different command and the tests would still pass. Since this PR is specifically adding runtime-list coverage, please assert the template strings/values here (or make the helper fail unless it receives the expected `LOCAL_RUNTIME`, `ls`, `--format`, `json` command) so the tests lock down the command as well as the parsing/filtering behavior.

Peyton-Spencer:
Triage note: this is correctly scoped as a test-only shared-VM coverage PR, but it appears to overlap with the other open shared-VM cleanup/runtime test PRs (#935 and #945) against `src/backends/shared-vm.test.ts`. Before merge, please reconcile this as the intended winner/superset or rebase after the competing shared-VM test PR lands to avoid another collision from the deterministic-coverage queue.

Peyton-Spencer:
PM: superseded by the 4-way shared-VM test consolidation — full keep/close plan and coverage map on #970. This PR overlaps #970 on `cleanupOrphans` coverage and will conflict on merge. Recommend closing after its unique tests (if any) are cherry-picked into #970. Root cause: #952.

Peyton-Spencer:
**3-way collision consolidation recommendation.**

PRs #935, #945, #965 all target only `src/backends/shared-vm.test.ts` and are now **all `CONFLICTING/DIRTY`** — they conflict with each other, so the "independently green" property that made parallel coverage PRs safe is gone (root cause: #952, no open-PR guard).

Proposed consolidation:
- **Keep #935** (oldest, 06-27) as the single surviving PR. Rebase it onto `main` to clear the conflict.
- **Close #945 and #965** as duplicates of #935 — port any unique cases from them into #935 first if they diverged.

This turns 3 dirty PRs into 1 green PR and unblocks the coverage target. Until the #952 guard lands, the coverage agent should be pointed away from files that already have an open PR.

Peyton-Spencer:
**PM (2026-07-06): consolidation already landed — recommend closing.**

The shared-VM cleanup consolidation **merged as #970 on 2026-07-04**. This PR (+90–98 lines, `src/backends/shared-vm.test.ts` only) has been `CONFLICTING/DIRTY` ever since because it overlaps the coverage #970 already rewrote — the 07-05 "consolidate among the three of you" note is now stale; there is nothing left to consolidate toward.

Recommended action: **close this PR.** If a git-diff against merged `main` shows any case #970 did *not* cover, reopen a minimal follow-up with just that delta rather than rebasing the full 90-line overlap. Same call applies to the sibling orphan in this set (#935/#945/#965) — all three should clear so `shared-vm.test.ts` stops re-colliding on every coverage run.

Root cause tracked in #952 (dup-coverage guard still unbuilt).

Peyton-Spencer:
PM cleanup: closing — superseded by the shared-VM test consolidation that merged as #970, and this PR is CONFLICTING against main (cannot merge as-is). Coverage already landed. Reopen if any unique case was missed. (Resolves the #952 shared-vm.test.ts collision.)

PR #968, opened 2026-07-03
feat(web): attribute /system longest-running age to its owning group

Peyton-Spencer:
## What

The `/system → queue` card surfaces the age of the oldest in-flight **task** (`longest running`) and **message run** (`longest message run`), but it discarded *which group* owned each. An operator seeing `longest running: 4.2m` then had to open `/ipc` and manually scan every lane to find the culprit — exactly the cross-page hop that #644 ("consolidated diagnostics in existing pages, not a new dashboard") is trying to eliminate.

This PR attributes those two ages to their owning group folder key.

## Changes

- **Data layer** (`buildHealthData`): track the owning `folderKey` alongside `longest_running_task_ms` / `longest_running_message_ms`; expose as new `longest_running_task_group` / `longest_running_message_group` on `HealthData.queue` (`string | null`).
- **Render**: queue card now shows `2.1m · omniaura-discord`. `metricRow` escapes the whole value, so the folder key is XSS-safe.
- **Live updates** (`page-scripts.ts`): both rows now stream group-attributed via the SSE poll loop. Note the `longest message run` row previously had **no** live-update handler at all — it was static until a full page reload; this wires it in to mirror the task row.
- **0ms edge**: a freshly-started task (`runningMs: 0`) is now attributed to its owner, so the row never renders an age with no group to jump to.
- **Docs**: `docs/observability.md` queue-field table updated.

## Tests

- `src/web/system.test.ts`: assert `longest_running_task_group` / `longest_running_message_group` values, the group-suffixed rendered HTML, null-group em-dash cases, and the 0ms attribution edge.
- `src/web/page-scripts.test.ts`: assert both live-update handlers (including the newly-added message-run handler).
- Full suite green: **2570 pass / 0 fail**; `bun run typecheck` clean; Prettier clean.

Refs #644

## Summary by CodeRabbit

* **Bug Fixes**
  * The `/system` queue view now shows the owning group next to the “longest running” task and message durations.
  * When no matching activity is present, these metrics correctly display an em dash (`—`).
  * Freshly started tasks are now attributed to their group and shown as `0ms · <group>` when applicable.

* **Documentation**
  * Updated the `/system` observability docs to clarify how queue metric values are displayed.


src/web/system.ts
```diff
@@ -321,6 +343,7 @@ export function buildHealthData(
       msgRunningMs > longestRunningMessageMs
```
Peyton-Spencer, line 343: Freshly-started message runs have the same 0ms attribution hole that this PR fixed for tasks. `processing_groups` is incremented from `messageLane.active`, and `GroupQueue.getDetailedStats()` can report `runningMs: Date.now() - startedAt`, which may be exactly `0` for a just-started run. In that case the render/SSE gates show `0ms`, but this `>` check leaves `longest_running_message_group` null, so the operator still gets an unattributed age. Consider mirroring the task logic here (`msgRunningMs >= longestRunningMessageMs || longestRunningMessageGroup === null`) and adding the same 0ms test coverage for message lanes.

Peyton-Spencer (commented):
Reviewed the queue attribution changes. I left one inline finding on the message-lane 0ms edge: task attribution handles freshly-started 0ms runs, but message attribution still uses a strict `>` check, so `/system` can render `0ms` without the owning group for a just-started message run.

Verification run locally:

```bash
bun test src/web/system.test.ts src/web/page-scripts.test.ts
bun run typecheck
```

Peyton-Spencer:
Leaving this unmerged for now: the message-lane 0ms attribution blocker is still valid, so the PR needs the message path to mirror the task-lane inclusive/null-fallback ownership logic before merge.

PR #972, opened 2026-07-04
feat(web): add markdown format to conversation export

Peyton-Spencer:
## What

Adds a `markdown` format to the conversation export endpoint (`GET /api/messages/{jid}/export?format=markdown`), alongside the existing `json` and `text` formats.

## Why

Agent and user messages are already authored in markdown. Exporting a conversation as markdown yields a clean, portable transcript that renders directly in GitHub issues/PRs, docs, and note apps — more useful than the flat `text` dump for sharing or archiving agent sessions. This rounds out the Phase 3 "conversation export" line item (#505), which shipped JSON + plain text.

## Changes

- Accept `format=markdown` (invalid formats still return `400` with an updated message listing all three)
- Render:
  - `# Conversation: <name>` heading
  - Bulleted metadata: JID (code-formatted), export timestamp, message count
  - Per-message `### <sender> · <ISO timestamp>` sections with content embedded verbatim, separated by `---` horizontal rules
- Serve as `Content-Type: text/markdown; charset=utf-8` with a `.md` `Content-Disposition` filename
- Regression test asserting content-type, filename, metadata, and message rendering

## Scope note

Purely additive to an existing, well-tested handler — no changes to the `json`/`text` paths. The endpoint is backend/API-only (no frontend caller yet), so there's no UI surface to update.

## Testing

- `bun test src/web/` → **885 pass, 0 fail**
- `bun run typecheck` → clean

## Summary by CodeRabbit

* **New Features**
  * Conversation exports now support **Markdown** format.
  * Markdown exports include a readable chat transcript with message sections, timestamps, and separators.
  * Exported files now download with a `.md` filename and the correct Markdown content type.

* **Bug Fixes**
  * Export requests now return a clearer error message when an unsupported format is selected.
  * Added test coverage for Markdown export behavior.

### DIFF
--- a/src/web/routes.test.ts
+++ b/src/web/routes.test.ts
@@ -1431,6 +1431,35 @@ describe('conversation export', () => {
     expect(text).toContain('Hi! How can I help?');
   });
 
+  it('returns markdown export with proper formatting', async () => {
+    const state = makeState({
+      getMessages: () => sampleMessages,
+      getChats: () => sampleChats,
+    });
+    const res = await handle(
+      new Request(
+        'http://localhost/api/messages/dc%3A123/export?format=markdown',
+      ),
+      state,
+    );
+    expect(res.status).toBe(200);
+    expect(res.headers.get('content-type')).toContain('text/markdown');
+    expect(res.headers.get('content-disposition')).toContain(
+      'Test_Chat-export.md',
+    );
+
+    const md = await res.text();
+    expect(md).toContain('# Conversation: Test Chat');
+    expect(md).toContain('- **JID:** `dc:123`');
+    expect(md).toContain('- **Messages:** 2');
+    expect(md).toContain('### Alice · 2026-03-01T10:00:00.000Z');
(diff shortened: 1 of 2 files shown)


Peyton-Spencer (commented):
Reviewed the `format=markdown` export path and surrounding export behavior.

No blocking findings. The implementation is additive, keeps the existing JSON/text paths intact, uses the existing filename sanitization path, and the markdown output shape is covered by a focused regression test.

Verified locally:
- `bun test src/web/routes.test.ts`
- `bun run typecheck`

CI is green for the required jobs I checked; CodeRabbit was still pending at review time.

Peyton-Spencer (commented):
Reviewed for merge readiness in the bounded merger pass. Not merging yet because the new markdown export path calls `new Date(msg.timestamp).toISOString()` directly, so a malformed or missing stored timestamp can throw and turn an export request into a 500. The existing text export path is non-throwing here, so please mirror that defensive handling and add a focused regression test before merge.

PR #978, opened 2026-07-05
feat(web): add /api/logs/recent structured log history endpoint

Peyton-Spencer:
## What

Adds `GET /api/logs/recent` — returns recent **structured** log records as JSON, backed by a bounded in-memory ring buffer.

## Why

The raw structured log stream (`/api/logs/stream`) is SSE-only — a client only sees records that arrive *after* it connects. There was no way to fetch log **history** as JSON.

The HTML dashboard already backfills newly connected SSE clients with recent *rendered* log lines (`recentLogs`), but that's HTML, not structured data. JSON consumers — the SolidStart frontend, external tooling, or a fresh raw-log viewer — had no backfill at all. This closes that gap for the Web UI (#157) log tooling.

## How

- **`src/web/log-buffer.ts`** — new `LogRingBuffer` class: bounded FIFO of serialized (JSON-safe) records with `push()`, `recent({ level, limit })`, `size`, `clear()`. Oldest-first; drops oldest past capacity. Pure and fully unit-tested.
- **`src/web/server.ts`**:
  - A single always-on `logger.subscribe` populates the buffer (skips `trace`), reusing the existing `serializeLogRecord`. Capacity `MAX_RECENT_LOG_RECORDS = 1000` (deeper than the 500-line HTML replay cap — structured records are cheap).
  - New route `GET /api/logs/recent` -> `{ logs, total }`. Supports `?level=` (validated; invalid values ignored) and `?limit=` (parsed, positive, capped). Non-GET -> 405.
  - Subscription is unsubscribed in `stop()`.
  - Added to the peer-trusted route allowlist alongside `/api/logs/stream`.

## Tests

- `src/web/log-buffer.test.ts` — 9 unit tests (ordering, capacity eviction, level filter, limit, filter-before-limit, copy-on-read, capacity clamping, `clear`).
- `src/web/server.test.ts` — 5 endpoint tests (JSON records, level filter, invalid-level fallback, limit, 405).

`bun run typecheck` clean; full `bun test src/web/` passes. Prettier-formatted.

Co-Authored-By: Claude Opus 4.6

## Summary by CodeRabbit

* **New Features**
  * Added a new recent-logs API that returns structured log history in JSON, including total count and support for filtering by log level.
  * Added a limit option for recent log results so clients can request only the newest matching entries.
  * Improved log retention so more recent structured logs are available for retrieval without affecting the existing live log view.

* **Bug Fixes**
  * Invalid log-level and limit values are now handled safely, returning sensible results instead of failing.

### DIFF
--- a/src/web/log-buffer.test.ts
+++ b/src/web/log-buffer.test.ts
@@ -1,0 +1,86 @@
+import { describe, it, expect } from 'bun:test';
+import { LogRingBuffer } from './log-buffer.js';
+
+describe('LogRingBuffer', () => {
+  it('stores records oldest-first', () => {
+    const buf = new LogRingBuffer(10);
+    buf.push({ level: 'info', msg: 'a' });
+    buf.push({ level: 'info', msg: 'b' });
+    expect(buf.recent().map((r) => r.msg)).toEqual(['a', 'b']);
+    expect(buf.size).toBe(2);
+  });
+
+  it('drops the oldest records past capacity', () => {
+    const buf = new LogRingBuffer(3);
+    for (const msg of ['a', 'b', 'c', 'd', 'e']) {
+      buf.push({ level: 'info', msg });
+    }
+    expect(buf.size).toBe(3);
+    expect(buf.recent().map((r) => r.msg)).toEqual(['c', 'd', 'e']);
+  });
+
+  it('filters by level', () => {
+    const buf = new LogRingBuffer(10);
+    buf.push({ level: 'info', msg: 'i1' });
+    buf.push({ level: 'error', msg: 'e1' });
+    buf.push({ level: 'info', msg: 'i2' });
+    expect(buf.recent({ level: 'error' }).map((r) => r.msg)).toEqual(['e1']);
+    expect(buf.recent({ level: 'info' }).map((r) => r.msg)).toEqual([
+      'i1',
+      'i2',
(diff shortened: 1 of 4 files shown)


Peyton-Spencer (commented):
Reviewed the implementation and tests.

No blocking findings. The dedicated always-on `logger.subscribe` for the recent structured buffer looks appropriate here: it keeps `/api/logs/recent` independent from connected SSE clients, avoids coupling JSON history to the rendered Datastar replay path, and is torn down in `stop()`. The extra subscriber cost is bounded to one in-process function call per emitted log record plus a capped array update.

`MAX_RECENT_LOG_RECORDS = 1000` also looks sane for this use case. The serialized records are narrow, the response limit is capped at the same value, and this gives JSON consumers enough backfill without turning the web process into long-term log storage.

Verified locally:
- `bun test src/web/log-buffer.test.ts src/web/server.test.ts`
- `bun run typecheck`
- `bun test src/web/`

Peyton-Spencer (commented):
Holding merge for one small API-contract fix: `/api/logs/recent?level=warn` returns a filtered `logs` array but `total` is still the unfiltered ring-buffer size. Since this is a new JSON endpoint, please make the metadata match the filtered result (or rename it to `bufferSize`) and add a test assertion for the chosen contract.

PR #979, opened 2026-07-05
fix(tasks): validate preprocess_script exists at creation (#973)

### LINKED ISSUE #973: Preprocess-gated tasks silently stuck 'blocked' for multi-channel agents (workflow path mismatch, no existence check)
## Summary

A scheduled task with a `preprocess_script` can be **permanently `blocked`** with a silent `Preprocessor failed: Module not found "…/groups/<group_folder>/task-workflows/<script>.ts"` error, even though the agent created the workflow file exactly where the `schedule_task` tool docs told it to.

Root cause: for **multi-channel / 4-layer-context agents**, the folder the agent writes to (`/workspace/group`) is **not** the folder the preprocessor resolves the workflow under (`groups/<task.group_folder>/task-workflows/`). Nothing validates the file exists at creation time, so the misconfiguration is only discovered at runtime — and then it retries forever with no notification.

## Evidence (live repro)

- Task `task-1779144774417-kxanmf` ("backend MCP watcher"), `preprocess: watch-backend-mcp.ts`, `cron: 0 7 * * *`.
- Snapshot state: `last_outcome_state: "blocked"`, `last_outcome_reason: Preprocessor failed: error: Module not found "/Users/…/omniclaw/groups/clayton-discord/task-workflows/watch-backend-mcp.ts"`.
- The workflow file **did exist** in the creating agent's workspace at `/workspace/group/task-workflows/watch-backend-mcp.ts` for ~2 months, and runs cleanly (`bun run …` -> valid `{"action":"skip",…}`). It just wasn't at the host path the preprocessor resolves.

## Root cause detail

1. **Creation** — `ipc.ts` `create_task` stores `group_folder = targetFolder` (the agent's group folder, e.g. `clayton-discord`) and calls `normalizePreprocessScriptPath()`, which validates **only** the string (traversal + extension). It does **not** check that the file exists at the resolved location.
2. **Resolution** — `task-preprocessor.ts` `defaultWorkflowsDir()` resolves the workflow under `path.resolve(GROUPS_DIR, task.group_folder, TASK_WORKFLOWS_DIR)` = `groups/clayton-discord/task-workflows/`.
3. **Mount mismatch** — in the 4-layer setup (`local-backend.ts`), a channel-context container mounts:
   - `/workspace/group`  <- `groups/<channelFolder>` (the **channel** folder)
   - `/workspace/agent`  <- `groups/<agentContextFolder>` (the **agent identity** folder, e.g. `clayton-discord`)

   For this agent, `task.group_folder` (`clayton-discord`) == the **agent** folder == `/workspace/agent`, **not** `/workspace/group`. But the `schedule_task` MCP tool docs instruct: *"write a TypeScript workflow file under `/workspace/group/task-workflows/`"*. Following the docs puts the file in `groups/<channelFolder>/task-workflows/`, which the preprocessor never reads.

Result: the docs' guidance is wrong for any agent whose `group_folder` differs from the folder mounted at `/workspace/group` (i.e. every multi-channel agent), and the failure is silent + permanent.

## Impact

- Preprocess-gated tasks silently never run for multi-channel agents.
- No alerting: `blocked` tasks retry every schedule tick forever; the only signal is buried in a snapshot field.
- The existing security guard (`assertPathWithin`) can't help — the path is valid, it just doesn't exist.

## Proposed fixes (pick 1–2)

1. **Validate existence at creation** (`ipc.ts` `create_task`/`edit_task`): resolve the workflow path the same way `runTaskPreprocessor` does and `fs.existsSync()` it. If missing, reject with an actionable error that **names the resolved directory** and emits it back to the creating agent — turning a silent 2-month failure into an immediate "put the file in `groups/clayton-discord/task-workflows/`".
2. **Fix the `schedule_task` tool docs** to instruct writing under the folder that maps to `task.group_folder` (the agent identity folder, `/workspace/agent/task-workflows/`, for multi-channel agents), or expose the resolved directory in the tool description dynamically.
3. **Alert / auto-pause on repeated preprocessor-resolution failure** (e.g. after N consecutive `blocked` outcomes with `Module not found`), so permanently-broken tasks surface instead of retrying silently.

## Workaround applied

Copied the workflow to `/workspace/agent/task-workflows/watch-backend-mcp.ts` (= `groups/clayton-discord/task-workflows/`), which should unblock the live task at its next run (2026-07-04 11:00 UTC). This is a per-task band-aid; the creation-time validation (fix #1) is the durable fix.

I'm happy to follow up with a PR implementing fix #1 + a doc correction if that direction looks right.

Peyton-Spencer:
## What

Closes #973. A scheduled task with a `preprocess_script` could be **permanently `blocked`** with a silent `Preprocessor failed: Module not found "…/groups/<group_folder>/task-workflows/<script>.ts"` — even though the agent created the workflow file exactly where the `schedule_task` tool docs said to.

Root cause: for **multi-channel / 4-layer-context agents**, the folder the agent writes to (`/workspace/group/task-workflows/`, the *channel* folder) is **not** the folder the scheduler resolves the workflow under (`groups/<task.group_folder>/task-workflows/` = the *agent-identity* folder mounted at `/workspace/agent/`). Nothing validated the file existed at creation time, so the misconfiguration only surfaced at the next scheduled run and then retried every tick with no notification.

Live repro: task `task-1779144774417-kxanmf` ("backend MCP watcher", `cron: 0 7 * * *`) has been failing daily; the workflow ran cleanly by hand but wasn't at the host path the scheduler resolves.

## Changes

- **Existence check** — new `checkPreprocessScriptExists` / `resolvePreprocessWorkflowPath` in `task-preprocessor.ts`, reusing the *exact* resolution `runTaskPreprocessor` uses (so the check and the runtime never disagree).
- **Reject early with an actionable error** — `schedule_task` / `edit_task` in `ipc.ts` now reject when the workflow file is missing and emit a `task_error` IPC event that **names the resolved directory**, turning a silent 2-month failure into an immediate "put the file in `…/task-workflows/`" message. The previously-silent invalid-path rejection is now surfaced the same way.
- **Doc correction** — the `schedule_task` MCP tool description (`container/agent-runner/src/ipc-mcp-stdio.ts`) now points at the folder that actually maps to the task's group folder: `/workspace/agent/task-workflows/` when that mount exists (4-layer agents), else `/workspace/group/task-workflows/`.

## Tests

- Unit: existence check (present / missing / no-script / traversal) + resolver path — `task-preprocessor.test.ts`.
- IPC handlers: reject-on-missing (asserts no task created + `task_error` naming the dir) and create-on-present — `ipc-task-handlers.test.ts`.
- Updated the existing deterministic-preprocessor test to create the workflow file first (reflects the new contract).
- `bun run typecheck` clean, agent-runner `tsc --noEmit` clean, full `bun test` green (2580 pass), prettier clean.

## Follow-up (not in this PR)

The issue also suggests option 3 — alert / auto-pause after N consecutive `blocked` `Module not found` outcomes so already-broken tasks surface rather than retrying silently. Happy to do that as a separate PR if wanted.

Note: this fixes the *class* of bug and makes the failure actionable going forward. The one live broken task still needs its workflow file placed at the resolved host path (`groups/clayton-discord/task-workflows/`) — the `/workspace/agent` copy in the container doesn't appear to propagate to that host path, which is worth confirming separately.

### DIFF
--- a/src/task-preprocessor.ts
+++ b/src/task-preprocessor.ts
@@ -1,4 +1,5 @@
 import { spawnSync } from 'child_process';
+import { existsSync } from 'fs';
 import path from 'path';
 
 import {
@@ -112,7 +113,9 @@ function resolveWorkflowPath(scriptPath: string, workflowsDir: string): string {
   return resolved;
 }
 
-function defaultWorkflowsDir(task: ScheduledTask): string {
+function defaultWorkflowsDir(
+  task: Pick<ScheduledTask, 'group_folder'>,
+): string {
   if (path.isAbsolute(TASK_WORKFLOWS_DIR)) {
     return TASK_WORKFLOWS_DIR;
   }
@@ -120,6 +123,69 @@ function defaultWorkflowsDir(task: ScheduledTask): string {
   const groupDir = path.resolve(GROUPS_DIR, task.group_folder);
   assertPathWithin(groupDir, GROUPS_DIR, 'task workflow group folder');
   return path.resolve(groupDir, TASK_WORKFLOWS_DIR);
+}
+
+/**
+ * Resolve the absolute workflow directory and file path a preprocess script
+ * will run from, using the same rules as {@link runTaskPreprocessor}. Exposed so
+ * task creation/edit can validate a script *before* it is persisted, instead of
+ * discovering a mismatch only at the next scheduled run. See issue #973.
+ */
+export function resolvePreprocessWorkflowPath(
+  groupFolder: string,
+  preprocessScript: string,
+  options: { workflowsDir?: string } = {},
+): { workflowsDir: string; workflowPath: string } {
+  const workflowsDir = path.resolve(
+    options.workflowsDir ?? defaultWorkflowsDir({ group_folder: groupFolder }),
+  );
+  const workflowPath = resolveWorkflowPath(preprocessScript, workflowsDir);
+  return { workflowsDir, workflowPath };
+}
+
+/**
+ * Validate that a task's preprocess script exists at the path the scheduler
+ * will resolve it from. Returns `ok: true` when there is no script (nothing to
+ * check) or the file is present; otherwise returns an actionable error that
+ * names the resolved directory so the caller can place the file correctly.
+ *
+ * This closes the silent-`blocked` failure mode where a multi-channel agent
+ * writes the workflow under `/workspace/group/task-workflows/` while the
+ * scheduler resolves it under the agent-identity folder
+ * (`/workspace/agent/task-workflows/`). See issue #973.
+ */
+export function checkPreprocessScriptExists(
+  groupFolder: string,
+  preprocessScript: string | null | undefined,
+  options: { workflowsDir?: string } = {},
+): { ok: true } | { ok: false; error: string } {
+  if (!preprocessScript) return { ok: true };
+  let resolved: { workflowsDir: string; workflowPath: string };
+  try {
+    resolved = resolvePreprocessWorkflowPath(
+      groupFolder,
+      preprocessScript,
+      options,
+    );
+  } catch (err) {
+    return {
+      ok: false,
+      error: err instanceof Error ? err.message : String(err),
+    };
+  }
+  if (!existsSync(resolved.workflowPath)) {
+    return {
+      ok: false,
+      error:
+        `preprocess_script "${preprocessScript}" was not found at ` +
+        `${resolved.workflowPath}. Create the workflow file in the task's ` +
+        `workflow directory (${resolved.workflowsDir}) before scheduling. ` +
+        `For multi-channel / 4-layer-context agents this directory is the ` +
+        `agent-identity folder mounted at /workspace/agent/task-workflows/, ` +
+        `not /workspace/group/task-workflows/.`,
+    };
+  }
+  return { ok: true };
 }
 
 function parseDecision(stdout: string): TaskPreprocessDecision {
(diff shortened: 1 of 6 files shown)


src/task-preprocessor.ts
```diff
@@ -122,6 +125,69 @@ function defaultWorkflowsDir(task: ScheduledTask): string {
...
+      preprocessScript,
+      options,
+    );
+  } catch (err) {
+    return {
+      ok: false,
+      error: err instanceof Error ? err.message : String(err),
+    };
+  }
+  if (!existsSync(resolved.workflowPath)) {
```
Peyton-Spencer, line 176: This should probably verify that the resolved entry is a regular file, not just that it exists. As written, a directory named `workflow.ts` passes creation/edit validation and then still fails later when `bun run` tries to execute it, preserving a version of the delayed runtime failure this PR is trying to eliminate. `statSync(resolved.workflowPath).isFile()` (or `lstatSync`, depending on whether symlinked workflows should be allowed) would close that gap and is easy to cover with one test.

Peyton-Spencer (commented):
Reviewed PR #979.

One non-blocking validation gap left inline: `checkPreprocessScriptExists` currently uses `existsSync`, so a directory named like `workflow.ts` passes validation but still fails at runtime. I would prefer `statSync(...).isFile()` / `lstatSync(...).isFile()` plus a small test before merge, but I do not think it invalidates the approach.

On the two requested design questions:

- `/workspace/agent` detection in `ipc-mcp-stdio.ts`: acceptable as a pragmatic signal because it matches the actual mount topology the agent can observe. An env var would be cleaner long-term, but the current fallback is low-risk and keeps this fix local.
- Rejecting `edit_task` when the new `preprocess_script` is missing: correct behavior. Warn-and-allow would keep the same delayed blocked-task failure mode. The implementation only rejects when `preprocess_script` is present in the edit payload, so unrelated edits can still go through.

Verified locally:
- `bun test src/task-preprocessor.test.ts src/ipc-task-handlers.test.ts src/ipc-processing.test.ts`
- `bun run typecheck`

Note: `bunx tsc --noEmit -p container/agent-runner/tsconfig.json` fails in this local checkout due missing container-only SDK packages (`@cursor/sdk`, `@anthropic-ai/claude-agent-sdk`, `@modelcontextprotocol/sdk`, `@opencode-ai/sdk`). The PR's GitHub `agent-runner-typecheck` check is green.

PR #980, opened 2026-07-05
fix(web): escape quotes in client HTML helper

### LINKED ISSUE #976: security: encode quotes in Web UI client escaping for attribute contexts
## Summary

The Web UI's browser-side escaping helper, `window.__esc`, escapes text by assigning to `textContent` and reading `innerHTML`. That protects element text from `<`/`>` injection, but it does not encode double quotes. Several page scripts concatenate `window.__esc(...)` into quoted HTML attributes and then assign the assembled string to `innerHTML`, so attacker-controlled values containing `"` can break out of the attribute and inject new attributes or markup.

## Severity

P2 / Medium.

The issue requires access to data that the Web UI renders, but multiple rendered fields are user- or agent-controlled and persisted, including scheduled task prompts, run result/error strings, message search chat JIDs, IPC last-error messages, and discovery peer/network identifiers. In deployments where the Web UI is reachable by a browser session, this can become stored or reflected XSS in the operator dashboard.

## Affected Code

- `src/web/shared.ts`: `window.__esc` currently returns `div.innerHTML`, which does not encode quotes for attribute contexts.
- `src/web/page-scripts.ts`: uses `window.__esc(...)` inside quoted attributes before `innerHTML` assignment, including:
  - task row `title` attributes for `task.prompt`, `task.chat_jid`, `task.next_run`, `task.last_run`
  - task run `title`, `data-run-at`, and `data-task-id` attributes
  - message search `data-jid`
  - IPC queue `title` and class-name fragments
  - network/discovery `data-network-id`, `data-instance-id`, and `title`
- `src/web/shared.ts`: command/task palette and run-history snippets also use `window.__esc(...)` in attribute contexts.

## Reproduction Sketch

One minimal string that demonstrates the escaping mismatch:

```js
const value = '" autofocus onfocus=alert(1) x="';
// Equivalent behavior to the current helper for quote handling.
const escaped = value.replace(/[&<>]/g, c => c === '&' ? '&amp;' : c === '<' ? '&lt;' : '&gt;');
console.log('<td title="' + escaped + '">x</td>');
```

Output:

```html
<td title="" autofocus onfocus=alert(1) x="">x</td>
```

In OmniClaw, an authenticated user or any path that can create/render a scheduled task can set a prompt containing the quote-breakout payload. When the Tasks page refreshes, `refreshTasks()` builds:

```js
'<td class="td-prompt" title="'+window.__esc(task.prompt)+'">'
```

and assigns the full string to `tbody.innerHTML`, allowing the browser to parse injected attributes/markup.

## Recommended Fix

- Replace `window.__esc` with an escaping helper that encodes `&`, `<`, `>`, `"`, and `'` when the result may be used in HTML attribute contexts.
- Prefer DOM construction APIs (`textContent`, `setAttribute`, `dataset`) for dynamic values instead of string concatenation into `innerHTML` on new code paths.
- Add regression tests that assert quote characters are encoded by the helper and that representative task/search/network render paths cannot inject attributes.

## Duplicate Checks

Checked open/all issues and PRs for `client escape quote attribute XSS __esc innerHTML`, `XSS Web UI innerHTML attribute quote`, and `sanitizeHtml __esc title data-task-id`; no duplicate was found.

Peyton-Spencer:
## Summary

- Replace the browser-side `window.__esc` helper with deterministic escaping for `&`, `<`, `>`, double quotes, and single quotes.
- Align server-side `escapeHtml` with the same quote-safe attribute behavior.
- Add regression coverage that the emitted client helper encodes quotes and no longer relies on `textContent`/`innerHTML` quote behavior.

Closes #976.

## Security Notes

The previous helper used `textContent` plus `innerHTML`, which escapes text nodes but leaves quote characters unencoded. Several Web UI scripts reuse `window.__esc(...)` inside quoted attributes before assigning generated strings to `innerHTML`, so quote payloads could break out of attributes.

## Verification

- `bun test src/web/shared.test.ts src/web/page-scripts.test.ts`
- `bunx tsc --noEmit`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTML escaping to include both double and single quotes, helping prevent unsafe rendering in more cases.
  * Updated the client-side escaping helper to use a more reliable string-based approach for sanitizing special characters.

* **Tests**
  * Expanded coverage to verify the additional escaping behavior and the generated client-side helper output.

### DIFF
--- a/src/web/shared.test.ts
+++ b/src/web/shared.test.ts
@@ -11,9 +11,19 @@ import {
 
 describe('escapeHtml', () => {
   it('escapes the HTML-sensitive characters used by the web UI', () => {
-    expect(escapeHtml('&<>"')).toBe('&amp;&lt;&gt;&quot;');
-    expect(escapeHtml("it's fine")).toBe("it's fine");
+    expect(escapeHtml('&<>"\'')).toBe('&amp;&lt;&gt;&quot;&#39;');
     expect(escapeHtml('')).toBe('');
+  });
+});
+
+describe('client escape helper', () => {
+  it('encodes quotes for generated HTML attribute contexts', () => {
+    const html = renderShell('/', 'Dashboard', '<div>content</div>', {});
+
+    expect(html).toContain('window.__esc=function(s){');
+    expect(html).toContain('c===\'"\'?"&quot;":"&#39;"');
+    expect(html).toContain('/[&<>"\']/g');
+    expect(html).not.toContain('document.createElement("div");d.textContent');
   });
 });
 
--- a/src/web/shared.ts
+++ b/src/web/shared.ts
@@ -23,7 +23,8 @@ export function escapeHtml(str: string): string {
     .replace(/&/g, '&amp;')
     .replace(/</g, '&lt;')
     .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;');
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
 }
 
 export function escapeJsonForHtml(json: string): string {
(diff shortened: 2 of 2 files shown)


Peyton-Spencer:
Reviewed current head `5b333bf` for the client/server HTML escape changes in PR #980.

No blocking findings.

The change covers the reported attribute-context issue: the browser-side `window.__esc` now performs deterministic single-pass escaping for `&`, `<`, `>`, `"`, and `'`, so values concatenated into double-quoted attributes such as task titles, data IDs, search result JIDs, and network/discovery IDs can no longer break out via quote payloads. Server-side `escapeHtml` is now aligned with the same quote-safe behavior. I also checked the nearby call sites and did not find an unquoted-attribute use that would require a different escaping contract.

Rendering risk looks low. Text-node uses of `escapeHtml` will still render apostrophes normally after entity decoding, and the broader web rendering tests pass.

Local verification:
- `bun test src/web/shared.test.ts src/web/page-scripts.test.ts`: 89 pass, 0 fail
- `bunx tsc --noEmit`: pass
- `bun test src/web`: 885 pass, 0 fail
- `bun test`: 2574 pass, 0 fail

Artifact/diff check:
- `git diff --check origin/main...HEAD`: clean
- Working tree clean after install/tests
- Branch changes only `src/web/shared.ts` and `src/web/shared.test.ts`

PR #989, opened 2026-07-08
feat(web): add task run history export endpoint

Peyton-Spencer:
## What

Adds a **task run history export** endpoint so operators can pull a scheduled task's execution history for offline audit and analysis:

```
GET /api/tasks/{id}/runs/export?format=json|csv&limit=N
```

Part of the Web UI (#157) task-manager surface — complements the existing `/api/tasks/{id}/runs` viewer with a downloadable, analysis-friendly export.

## Details

- **JSON** (default) mirrors the conversation-export envelope: `{ task_id, exported_at, run_count, runs[] }`, served as an attachment (`task-{id}-runs.json`).
- **CSV** is spreadsheet-friendly for success-rate / latency analysis — columns: `run_at, status, duration_ms, outcome_state, outcome_reason, result, error`. Uses RFC-4180 escaping (`csvCell`) so commas, quotes, and newlines in a run's `result`/`error` don't corrupt rows.
- `limit` defaults to 1000, capped at the existing `MAX_EXPORT_LIMIT` (5000).
- Error handling: `404` unknown task, `400` invalid format, `405` non-GET — consistent with sibling task routes.

## Testing

- 6 new tests in `src/web/routes.test.ts` (JSON envelope, CSV header + RFC-4180 escaping, invalid format, non-GET, limit cap, unknown task).
- `bun test src/web/routes.test.ts` → 72 pass, 0 fail.
- `bun run typecheck` → clean.
- Full `src/web/` suite: only pre-existing unrelated failure (`renderPeerRows` relative-time flake in `network.test.ts`, fails on `main` without this change).

## Scope

API-only, mirroring the pattern of #978 (`/api/logs/recent`). A download button in the tasks UI can follow as a small separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a new export option for task run history, available in JSON or CSV format.
  * Export responses now include download-friendly headers and a clear file attachment name.

* **Bug Fixes**
  * Improved handling for invalid task IDs, unsupported export formats, and non-GET requests.
  * Added safer CSV escaping for quotes, commas, and multiline values.
  * Enforced a maximum limit on exported runs to prevent overly large responses.

PR #990, opened 2026-07-08
test(web): fix time-drifted peer last-seen tests failing CI on main

Peyton-Spencer:
## What

Fixes the `renderPeerRows` / `renderPeerOnlineCell` last-seen tests in `src/web/network.test.ts` that now **fail on `main`** — turning CI red on every open PR (including #989).

## Root cause

`formatPeerLastSeen(lastSeen, nowMs?)` humanizes a timestamp, falling back to an absolute locale date once the gap exceeds ~30 days:

```ts
if (diff < 30 * 86_400_000) return `${Math.floor(diff / 86_400_000)}d ago`;
return new Date(t).toLocaleDateString();
```

`renderPeerRows` / `renderPeerOnlineCell` call it **without** injecting `nowMs`, so they humanize against the real `Date.now()`. But two tests built their fixtures from a **fixed fake date**:

```ts
const now = Date.parse('2026-06-07T12:00:00.000Z');
const lastSeen = new Date(now - 5 * 60_000).toISOString();
```

Once real wall-clock time passed `2026-06-07` by 30 days (i.e. from ~2026-07-07 onward), the render path crossed the locale-date fallback and emitted an absolute date instead of `"5m ago"`. The regex `/(now|\d+m ago|\d+h ago|\d+d ago)/` no longer matched → **permanent failure, not a flake**.

The `formatPeerLastSeen` unit tests are unaffected — they correctly inject a fixed `NOW` as `nowMs`.

## Fix

Anchor both render-path fixtures to `Date.now()` so the relative-chip branch renders deterministically regardless of when CI runs, and assert the exact relative string:

- `renders an offline peer with a last-seen relative chip…` → fixture is `Date.now() - 5m`, asserts `>5m ago<`
- `emits both the relative chip and the absolute title…` → fixture is `Date.now() - 2h`, asserts `>2h ago<` (previously this one passed only via the locale-date fallback, silently no longer exercising the relative path it is named for)

## Verification

- `bun test src/web/network.test.ts` → **32 pass / 0 fail**
- `bunx tsc --noEmit` → clean

Test-only change; no production behavior touched.

### DIFF
--- a/src/web/network.test.ts
+++ b/src/web/network.test.ts
@@ -256,8 +256,12 @@ describe('renderPeerRows', () => {
   });
 
   it('renders an offline peer with a last-seen relative chip and absolute title', () => {
-    const now = Date.parse('2026-06-07T12:00:00.000Z');
-    const lastSeen = new Date(now - 5 * 60_000).toISOString();
+    // `renderPeerRows` humanizes lastSeen against the real `Date.now()`, so the
+    // fixture must be anchored to the wall clock too. A fixed fake `now` drifts
+    // past the 30-day locale-date fallback in `formatPeerLastSeen` and renders
+    // an absolute date instead of "Nm ago" — a permanent CI failure (not a
+    // flake) once real time passes the fixture by a month.
+    const lastSeen = new Date(Date.now() - 5 * 60_000).toISOString();
     const html = renderPeerRows([
       makePeer({
         instanceId: 'offline-1',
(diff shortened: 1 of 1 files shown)


Peyton-Spencer (commented):
Verified — LGTM. This is the keystone fix for the current CI-red-on-every-open-PR situation.

**Confirmed root cause / scope:**
- The failure reproduces identically on two independent branches (#996 and #998), which means it's failing on `main` itself — a *permanent* failure, not a flake. Both fail at `network.test.ts:275` with the render path emitting an absolute date (`6/7/2026`) instead of a `Nm ago` chip, because the fixtures were anchored to a fixed `2026-06-07` fake `now` that has since drifted past the 30-day locale-date fallback in `formatPeerLastSeen`.

**The fix is correct and additive-quality:**
- Anchoring both fixtures to `Date.now()` keeps the relative-chip branch deterministic regardless of when CI runs.
- Tightening the assertions from the loose `/(now|\d+m|\d+h|\d+d) ago/` regex to exact `>5m ago<` / `>2h ago<` is a genuine strengthening: the `renderPeerOnlineCell` test previously passed only via the locale-date fallback, so it had stopped exercising the relative-chip path it's named for. Now it does.
- Test-only; no production behavior touched. All checks green here.

**Recommend prioritizing this merge** — it's the single change that turns CI green for the ~18 clean PRs currently blocked behind it.

PR #995, opened 2026-07-08
test(simtest): cover discovery environment state

Peyton-Spencer:
## Summary

- Adds deterministic assertions for the simtest discovery environment's default network page state.
- Covers initially-offline trusted peer insertion so offline peers remain visible without discovered addresses.
- Covers direct missing-peer client lookup returning null.
- Stabilizes the network last-seen test by deriving the fixture timestamp from current time instead of a stale fixed 2026 date.

## Test Count

- Baseline on origin/main before changes: `2573 pass, 1 fail` across 106 files. Failure was `src/web/network.test.ts` last-seen date drift.
- After changes: `2577 pass, 0 fail` across 106 files.

## Verification

- `bun test src/simtest/discovery-sim.test.ts` -> 14 pass, 0 fail
- `bun test src/web/network.test.ts` -> 32 pass, 0 fail
- `bun test` -> 2577 pass, 0 fail

## Remaining Coverage Gaps

- Simtest admin API still has many route branches that could use direct tests for invalid JSON/body validation and scenario mutation paths.
- Web UI route coverage is broad but still heavily HTML-string based; browser-level interaction coverage remains separate from this PR.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for peer discovery behavior, including default network state, trusted peer details, offline trusted peers, and missing peer lookups.
  * Updated an existing offline peer test to use the current runtime clock, keeping the timestamp check reliable across time drift.

### DIFF
--- a/src/simtest/discovery-sim.test.ts
+++ b/src/simtest/discovery-sim.test.ts
@@ -458,4 +458,77 @@ describe('createSimDiscoveryEnvironment', () => {
       'Remote peer not found: missing-peer',
     );
   });
+
+  it('builds a deterministic default network page state for the seeded peer', () => {
+    const env = createSimDiscoveryEnvironment(new FakeState());
+
+    const pageState = env.getNetworkPageState();
+
+    expect(pageState).toMatchObject({
+      instanceId: 'sim-local-instance',
+      instanceName: 'Sim Local OmniClaw',
+      discoveryAvailable: true,
+      discoveryEnabled: true,
+      runtime: {
+        enabled: true,
+        active: true,
+        currentNetwork: {
+          id: 'sim-lan',
+          label: 'Simulated LAN',
+        },
+        trustedNetworks: [],
+      },
+      pendingRequests: [],
+    });
+    expect(pageState.peers).toHaveLength(1);
+    expect(pageState.peers[0]).toMatchObject({
+      instanceId: 'peer-remote-1',
+      name: 'Remote OmniClaw',
+      host: 'remote-sim.local',
+      port: 3100,
+      addresses: ['192.168.1.80'],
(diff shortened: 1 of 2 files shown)


src/simtest/discovery-sim.test.ts
```diff
@@ -458,4 +458,77 @@ describe('createSimDiscoveryEnvironment', () => {
...
+        online: false,
+        status: 'trusted',
+      }),
+    );
+  });
+
+  it('returns null when direct peer-client lookup targets an unknown peer', () => {
+    const env = createSimDiscoveryEnvironment(new FakeState());
+
+    expect(env.context.createPeerClient({ instanceId: 'missing' } as any)).toBe(
```
Peyton-Spencer, line 530: This new assertion breaks the PR's typecheck: `env.context.createPeerClient` is optional on `DiscoveryRouteContext`, and its signature requires both the peer and the route context. CI reports `TS2722` and `TS2554` here. Please either narrow/assert the helper before calling it and pass `env.context` as the second argument, or exercise the route path that already performs the optional guard.

Peyton-Spencer (commented):
I found one blocking issue: the new test currently fails `tsc --noEmit`. Focused Bun tests pass locally, but the PR cannot merge while the typecheck job is red.

Peyton-Spencer:
#995 is no longer merge-ready after #990 landed.

GitHub now reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, likely because both PRs touch `src/web/network.test.ts`. Please rebase/update this branch onto current `main`, resolve the overlap, and let CI rerun. The simtest coverage itself still looks low-risk once the branch is clean again.

PR #996, opened 2026-07-08
fix(web): validate context file realpaths

Peyton-Spencer:
## Summary

- resolve discovered local `CLAUDE.md` paths to their real filesystem targets before reading them
- reject context files whose realpath escapes `GROUPS_DIR`, preventing symlink metadata/hash leaks through the web/discovery context inventory
- add regression coverage for an outside-target symlink while preserving normal context-file listing

## Security Impact

`listLocalContextFiles()` previously validated the path string under `GROUPS_DIR` before calling `statSync` and `readFileSync`. A symlink named `CLAUDE.md` inside a group folder could point outside `GROUPS_DIR`; the scanner would then follow it and expose the target file's size, mtime, and SHA-256 hash through context-file inventory APIs.

This change validates both the discovered path and the resolved realpath before stat/read.

## Tests

- `bun test src/web/context-files.test.ts`
- `bun run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local context file discovery to follow symlinks safely before reading file details.
  * Context files that point outside the allowed directory are now ignored, preventing unexpected files from appearing.
  * Existing directory scanning behavior remains intact for valid `CLAUDE.md` files.

### DIFF
--- a/src/web/context-files.test.ts
+++ b/src/web/context-files.test.ts
@@ -8,12 +8,14 @@ import { listLocalContextFiles } from './context-files.js';
 
 const realExistsSync = fs.existsSync;
 const realReaddirSync = fs.readdirSync;
+const realRealpathSync = fs.realpathSync;
 const realStatSync = fs.statSync;
 const realReadFileSync = fs.readFileSync;
 
 const mockedFs = fs as unknown as {
   existsSync: typeof fs.existsSync;
   readdirSync: typeof fs.readdirSync;
+  realpathSync: typeof fs.realpathSync;
   statSync: typeof fs.statSync;
   readFileSync: typeof fs.readFileSync;
 };
@@ -31,6 +33,7 @@ function file(name: string): FakeDirent {
 afterEach(() => {
   mockedFs.existsSync = realExistsSync;
   mockedFs.readdirSync = realReaddirSync;
+  mockedFs.realpathSync = realRealpathSync;
   mockedFs.statSync = realStatSync;
   mockedFs.readFileSync = realReadFileSync;
 });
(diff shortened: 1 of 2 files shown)


Peyton-Spencer:
CI note: the `test` job failure appears unrelated to this PR.

I reproduced the failing test locally with:

```bash
bun test src/web/network.test.ts
```

Failure:

```text
renderPeerRows > renders an offline peer with a last-seen relative chip and absolute title
```

The test uses a fixed `2026-06-07T11:55:00.000Z` timestamp. As of today, that timestamp renders as an absolute date (`6/7/2026`) instead of matching the test's expected `now|Xm ago|Xh ago|Xd ago` relative-chip pattern. The focused security regression test and typecheck passed:

```bash
bun test src/web/context-files.test.ts
bun run typecheck
```

Peyton-Spencer (commented):
Reviewed the security fix. I did not find any blocking issues.

What I checked:
- The discovered `CLAUDE.md` path is still constrained under `GROUPS_DIR` before filesystem access.
- The resolved `realpath` is also constrained with `assertPathWithin`, so sibling-prefix escapes and symlink targets outside `groups/` are rejected before `statSync` or `readFileSync` run.
- The regression test covers an in-tree `CLAUDE.md` symlink resolving to `../.env` and would fail if the outside target were stat/read.

Verification run locally on the PR branch:
- `bun test src/web/context-files.test.ts`
- `bun run typecheck`

Peyton-Spencer:
PM: the red `test` job here is **not caused by this PR's changes**. The one failing test is `renderPeerRows > renders an offline peer with a last-seen relative chip and absolute title` in `src/web/network.test.ts` — a time-drift date assertion (received `6/7/2026`). That exact flake was fixed on `main` by #990 (merged 2026-07-08), which this branch predates.

Fix: rebase/merge `main` into `ocpeyton/security-context-files-realpath` and CI should go green. The realpath-validation change itself (2 files, +59/-2) is clean and unrelated to the failure.

PR #998, opened 2026-07-09
feat(web): centralize lane reason-code registry with operator tooltips

Peyton-Spencer:
## Summary

Centralizes the message/task **lane reason-code taxonomy** into a single source of truth and surfaces human-readable descriptions as **operator tooltips** on the reason chips. Incremental sub-item of the operator-observability tracker (#644 — "reason codes wired through the data layer").

## Problem

The ordered reason lists / allow-lists were hand-copied across **four** files (`agents-page.ts`, `ipc-inspector.ts`, `system.ts`, `page-scripts.ts`):

- The `Set<MessageLaneReason>` literals are **not completeness-checked** by TypeScript — a `Set` missing a member is still a valid `Set<MessageLaneReason>`. Adding a new reason code could silently drop it from a page's count map or allow-list (misclassified as `unknown`).
- Chips render the raw kebab-case code (`no-work`, `back-pressure`, `cooling-down`) with **no explanation** for operators.

## Changes

- **New canonical registry in `group-queue.ts`** (co-located with the reason types):
  - `MESSAGE_LANE_REASONS` / `TASK_LANE_REASONS` — ordered lists
  - `MESSAGE_LANE_REASON_META` / `TASK_LANE_REASON_META` — `{ label, description }` maps, keyed by the union so the compiler **rejects any partial map**
  - `describeLaneReason()` — resolves metadata with a safe fallback for unknown codes
- **Refactor** `agents-page`, `ipc-inspector`, `system` to derive their lists/sets from the registry instead of redefining literals (net −38 lines of duplication in consumers).
- **Tooltips**: description surfaced as `title=` on server-rendered chips (`/agents`, `/ipc`); `window.__laneReasonDesc` injected into the shell so the live-updating client renderers (IPC queue table, agent-detail badge) show the same tooltip. Server and client stay consistent.

## Tests

- Registry exhaustiveness (every derivation output is in the canonical list, every reason is reachable), no-duplicates, non-empty label/description coverage, `describeLaneReason` fallback behavior.
- Shell tooltip-map injection assertion in `shared.test.ts`.
- `bun run typecheck` clean; full suite **2579 pass**. (The one unrelated failure — `network.test.ts` offline-peer last-seen — is a pre-existing time-drift flake already fixed by #990; it fails identically on clean `main`.)

## Notes

- No new endpoints, no API/schema changes, no new top-level route — extends existing surfaces per #644's guidance.
- No reason codes added or removed; taxonomy is unchanged, only consolidated + labeled.

## Summary by CodeRabbit

* **New Features**
  * Added hover tooltips to lane reason badges across the web UI, showing clearer explanations for queue and agent status reasons.
  * Expanded client-side rendering so reason descriptions are available consistently in the browser.

* **Bug Fixes**
  * Standardized lane reason handling to keep displayed reason lists and descriptions in sync.
  * Improved fallback behavior for unknown reasons so they still render safely.

* **Tests**
  * Added coverage to verify reason lists, descriptions, and tooltip output stay correct.


src/web/ipc-inspector.ts
```diff
@@ -273,7 +266,11 @@ function renderLaneReason<T extends string>(
   allowedCodes: ReadonlySet<T>,
 ): string {
   const code = allowedCodes.has(reason) ? reason : 'unknown';
-  return `<span class="lane-reason reason-${code}">${escapeHtml(code)}</span>`;
+  const meta = describeLaneReason(code as MessageLaneReason);
```
Peyton-Spencer, line 269: This helper is also used for task-lane chips, but `describeLaneReason()` currently always reads `MESSAGE_LANE_REASON_META` for overlapping codes. That makes `/ipc` task tooltips misleading: task `running` renders `Actively processing a message.`, task `back-pressure` says `Messages are pending...`, and the live client path has the same issue because `window.__laneReasonDesc` is built only from message metadata. Could this take the lane kind or metadata map as an argument so task chips use `TASK_LANE_REASON_META`?

Peyton-Spencer (commented):
I found one tooltip correctness issue. Targeted tests and typecheck pass locally.

